PR #46, opened 2026-01-07
Fix concurrency issue in RAdapter.cs

pieteckhart:
I noticed that the dictionaries used for caching inside the RAdapter were not thread safe.
Changing them to ConcurrentDictionary solves a problem I am having when running lots of unit tests in parallel.

I hope the performance penalty for using ConcurrentDictionary instead of the regular dictionary is acceptable...

### DIFF
--- a/src/PeachPDF/Html/Adapters/RAdapter.cs
+++ b/src/PeachPDF/Html/Adapters/RAdapter.cs
@@ -17,6 +17,7 @@ using PeachPDF.Html.Core;
 using PeachPDF.Html.Core.Handlers;
 using PeachPDF.Html.Core.Utils;
 using PeachPDF.Network;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


pieteckhart:
I'm now using a different approach using SemaphoreSlim beacause I think using the ConcurrentDictionary makes things to slow.

```C#
private readonly SemaphoreSlim generatorSemaphore = new SemaphoreSlim(1, 1);

...

await generatorSemaphore.WaitAsync();

try
{
    document = await generator.GeneratePdf(html, pdfConfig);
}
finally
{
    generatorSemaphore.Release(); // Always release the semaphore to avoid deadlocks
}
```
I'm also considering using an ObjectPool to allow to run in parallel without having to create a generator for each request.
Then I would not need the semaphore.

I'm leaving this PR open as an indicator that there is a thread-safety issue but feel free to close this if this is not desired.

jhaygood86:
I will be closing this in favor of #91 , which fixes all known thread safety issues. We are also explicitly documenting that PdfGenerator is not intended to be used simultaneously across multiple threads.

PR #48, opened 2026-03-06
Upgrades MimeKit to 4.15.1 from 4.14.0 to avoid a vulnerability in previous versions.

Leon-Lj:
All tests in `PeachPDF.Tests` passes after upgrade.


jhaygood86:
Ths is included in the next update indirectly

PR #55, opened 2026-06-18
Add WOFF/WOFF2 font conversion and tests

jhaygood86:
Add support for converting WOFF and WOFF2 webfont containers to OpenType before embedding. Introduces WoffConverter, Woff2Converter and a FontFormatConverter wrapper to detect/convert formats (uses Brotli and zlib paths as needed). Update PdfSharp adapter and PdfCIDFont handling to ensure correct embedding of TrueType and CFF/OTF fonts (FontFile2 vs FontFile3/OpenType behavior). Add unit and integration tests (FontFormatConverterTests, FontEmbeddingIntegrationTests) verifying detection, conversion, decompression, and PDF embedding invariants.


jhaygood86:
All Copilot review comments addressed in commit 08fb600:

**`PdfSharpAdapter.cs`**
- Both `memoryStream` and `convertedStream` are now `using var` — buffers released promptly after `AddFont` returns (FontResolver copies the bytes synchronously, so this is safe).

**`WoffConverter.cs`**
- Added 44-byte minimum header check → `InvalidDataException`
- Reject `compLength > origLength` per WOFF spec → `InvalidDataException`
- Validate table data fits within file bounds
- Throw instead of silently returning a zero-padded buffer when zlib decompresses fewer bytes than expected

**`Woff2Converter.cs`**
- Added 48-byte minimum header check → `InvalidDataException`
- Validate `compressedStart + totalCompressedSize <= woff2.Length` before creating the `MemoryStream`
- Reject reserved `transformVersion` values: for glyf/loca tables (1 and 2 are reserved), for all other tables (1, 2, 3 are reserved) → `InvalidDataException`
- Added `checked((int)TransformLength)` and bounds check against decompressed buffer before slicing each table

**Tests**
- Fixed typos: `EmbeddsAs` → `EmbedsAs` in both CFF and TTF test names
- All `C:\Windows\Fonts` lookups guarded with `OperatingSystem.IsWindows()` and resolved via `%SystemRoot%\Fonts` so tests skip cleanly on Linux/macOS CI
- `CffFont_EmbeddedStream_HasNo_Length1`: added `Assert.Contains("/FontFile3")` guard before scanning object blocks; changed `LastIndexOf("obj", ...)` to `LastIndexOf(" obj", ...)` to prevent false match inside `endobj`
- Added `Woff2Convert_SystemFont_ProducesValidOpenType` — exercises the full Brotli + inverse-transform path when a `.woff2` file is present in the system fonts directory (skips gracefully if none found)

jhaygood86:
Second-round Copilot review comments addressed in commit 3f568e2:

**`WoffConverter.cs`** (lines 102, 108)
- Changed `new byte[origLen]` → `new byte[checked((int)origLen)]` in both the compressed and verbatim branches. `origLen` is `uint` and implicitly widens to `long` for array allocation, but using `checked((int)...)` makes the cast explicit and throws `OverflowException` for malformed tables with implausibly large `origLength` values before allocating memory.

**`Woff2Converter.cs`** (lines 540, 553)
- `Read255UInt16`: added `pos >= src.Length` guard before the initial read, `pos + 1 >= src.Length` guard for the two-byte `wordCode` case, and `pos >= src.Length` guards for the two `oneMoreByteCode` cases. Truncated input now throws `InvalidDataException` instead of `IndexOutOfRangeException`.
- `ReadBase128`: added `pos >= data.Length` guard inside the loop body before `data[pos++]`. Same rationale — corrupt/truncated WOFF2 input now surfaces a controlled error.

**Test helpers** (`FontFormatConverterTests.cs` lines 281, 346, 372; `FontEmbeddingIntegrationTests.cs` line 248)
- All `new byte[uint_value]` array allocations in the `WrapTtfAsWoff` / `WrapTtfAsWoffWithCompressedFirstTable` helpers now use `checked((int)...)` for consistency with the production code fix.

All 1502 tests pass after these changes.

PR #83, opened 2026-07-10
Fall back to explicitly registered fonts on platforms without system font discovery

### LINKED ISSUE #82: PdfGenerator cannot be constructed on iOS/Android: FontResolver static ctor throws NotImplementedException on platforms without system font discovery
## Environment

- PeachPDF 0.9.1 (behavior also present on current `main`)
- .NET MAUI iOS app (`net10.0-ios`), reproduced on an iPad Air simulator, iOS 26.2

## What happens

Constructing `PdfGenerator` throws before any HTML is involved:

```
System.TypeInitializationException: The type initializer for 'PeachPDF.PdfSharpCore.Utils.FontResolver' threw an exception.
 ---> System.NotImplementedException: FontResolver not implemented for this platform (PeachPDF.PdfSharpCore.Utils.FontResolver.cs).
   at PeachPDF.PdfSharpCore.Utils.FontResolver..cctor()
   at PeachPDF.Adapters.PdfSharpAdapter..ctor()
   at PeachPDF.PdfGenerator..ctor()
```

`FontResolver`'s static constructor implements system-font discovery for Windows, Linux, and macOS and throws on everything else. Because `PdfSharpAdapter` instantiates `FontResolver` unconditionally, `PdfGenerator` is unusable on those platforms — even when the caller intends to supply every font explicitly via `AddFontFromStream`, which otherwise works without any system fonts at all.

## Suggested fix

On unrecognized platforms, initialize `SupportedFonts` to an empty array instead of throwing, so resolution is driven entirely by explicitly registered fonts. Platform-specific discovery for iOS/Android could still be added later; this just removes the hard failure. I've opened a small PR with this change.

## Verification

With that change (plus `AddFontFamilyMapping("Segoe UI", "<registered family>")`, since the HTML engine's default family isn't present on mobile), PeachPDF works very well on .NET MAUI iOS: a 3-page US-Letter document with a 28-row table, data-URI images, and `@page` margin-box footers renders in ~1.2 s cold / ~0.85 s warm on an iPad simulator, paginates the table with repeated header rows, resolves `counter(page)/counter(pages)`, and prints via `UIPrintInteractionController`.

Great library — thanks for building it.

maxwell-fraser:
## Summary

`FontResolver`'s static constructor throws `NotImplementedException` on any platform other than Windows, Linux, or macOS, which makes `PdfGenerator` unconstructible on iOS/Android (.NET MAUI) — even when the caller supplies every font explicitly via `AddFontFromStream`. This PR initializes `SupportedFonts` to an empty set on such platforms instead, so explicitly registered fonts drive resolution.

Fixes #82

## Details

- Only the final fall-through branch of the static constructor changes; Windows/Linux/macOS discovery is untouched.
- With zero system fonts and nothing registered, the behavior is the existing clear failure from `ResolveTypeface` (`FileNotFoundException: "No Fonts installed on this device!"`) — the same thing an empty Linux font environment produces today.
- Platform-specific discovery for iOS/Android could still be added later; this change just removes the hard constructor failure so the registered-fonts workflow is usable.

## Verification

- Verified on .NET MAUI iOS (`net10.0-ios`, iPad Air simulator, iOS 26.2): with fonts registered through `AddFontFromStream` and the default families remapped via `AddFontFamilyMapping`, a 3-page US-Letter document (28-row table, data-URI images, `@page` margin-box footers with page counters) renders in ~1.2 s cold / ~0.85 s warm and prints via `UIPrintInteractionController`.
- `dotnet test` on macOS: 2326 passed / 490 failed — byte-for-byte identical counts on unpatched `main` (those failures are font-environment dependent on macOS), i.e. no test regressions from this change.

### DIFF
--- a/src/PeachPDF.Tests/PdfSharpCore/FontResolverTests.cs
+++ b/src/PeachPDF.Tests/PdfSharpCore/FontResolverTests.cs
@@ -113,5 +113,37 @@ namespace PeachPDF.Tests.PdfSharpCoreTests
             Assert.NotNull(bytes);
             Assert.True(bytes.Length > 0);
         }
+
+        [Fact]
+        public void DiscoverSupportedFonts_UnsupportedPlatform_ReturnsEmpty()
+        {
+            var fonts = FontResolver.DiscoverSupportedFonts(isOSX: false, isLinux: false, isWindows: false);
+
+            Assert.Empty(fonts);
+        }
+
+        [Fact]
+        public void DiscoverSupportedFonts_Linux_DelegatesToLinuxSystemFontResolver()
+        {
+            var fonts = FontResolver.DiscoverSupportedFonts(isOSX: false, isLinux: true, isWindows: false);
+
+            Assert.Equal(LinuxSystemFontResolver.Resolve(), fonts);
+        }
+
+        [Fact]
+        public void DiscoverSupportedFonts_Windows_ReturnsSystemFontFiles()
+        {
+            var fonts = FontResolver.DiscoverSupportedFonts(isOSX: false, isLinux: false, isWindows: true);
+
+            Assert.NotNull(fonts);
+        }
+
+        [Fact]
+        public void DiscoverSupportedFonts_OSX_UsesLibraryFontsDirectory()
+        {
+            // /Library/Fonts/ doesn't exist on non-macOS test runners, but the branch
+            // must still be exercised so it's covered on the platforms where it does.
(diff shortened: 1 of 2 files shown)


jhaygood86 (approved):
This works for now.

jhaygood86:
It looks like the test suite is highlighting that there's no test coverage on this case. If you want to take a stab at fixing that, feel free. If not, I can address it. Either way, this will be in 0.9.2 that is planned for release on Monday

[diff-cover-report.html](https://github.com/user-attachments/files/29895860/diff-cover-report.html)

PR #209, opened 2026-07-22
Let author display override UA table display

jhaygood86:
## Summary

An author (or inline) `display` declaration on a table-tagged element — `table`, `tr`, `td`, `th`, `thead`, `tbody`, `tfoot`, `caption`, `col`, `colgroup` — was silently dropped by a gate (`DomParser.IsStyleOnElementAllowed`) that forced these elements back to their UA default `display: table-*`, keeping only `display: none` as an exception.

This violates the CSS cascade ([CSS Cascade 5 §6.3](https://www.w3.org/TR/css-cascade-5/#cascade-origin)) and [CSS Display 3](https://www.w3.org/TR/css-display-3/): an author `display` declaration overrides the UA `display: table-*`. There is no spec basis for forcing table tags back to their default.

## Change

Remove the gate so the override wins by the normal cascade. This is safe because everything downstream is already driven by the **computed `Display` value**, not the tag name:
- the anonymous-table fixup passes (`CorrectAnonymousTables*`) key off `box.Display`;
- layout-engine selection (`CssBox.PerformLayoutImp`) picks flex/table/block by `Display`.

So a `td` set to `display: block`/`flex` is no longer treated as a table cell — which is what lets a pure-CSS charting framework (e.g. Charts.css) reset a semantic `<table>` into a flex layout.

## Why

This is the first of a stacked series enabling a Charts.css showcase. Without it, a `<table class="charts-css">` renders as a plain table instead of a chart.

## Tests

- New `TableDisplayOverrideTests`: inline and stylesheet `display` overrides on the table and each internal element (including a full Charts.css-style reset), plus a guard that the UA default still applies when there is no override.
- Full net8.0 suite green (4496 passed, 0 failed) — no table/Acid2/anonymous-table regressions.

---
_Generated by [Claude Code](https://claude.ai/code/session_01S5W74ou6ZHgzScpiHjikGx)_

### DIFF
--- a/src/PeachPDF.Tests/Integration/AnonymousTableBoxIntegrationTests.cs
+++ b/src/PeachPDF.Tests/Integration/AnonymousTableBoxIntegrationTests.cs
@@ -123,16 +123,17 @@ namespace PeachPDF.Tests.Integration
         }
 
         [Fact]
-        public async Task MisparentedTableCell_WrappedInAnonymousRow_AtItsOriginalDocumentPosition()
+        public async Task MisparentedTableCell_WrappedInAnonymousRowThenTable_AtItsOriginalDocumentPosition()
         {
-            // Rule 3.1 (CorrectAnonymousTablesGenerateMissingParents): a `table-cell` box whose parent
-            // isn't a `table-row` must be wrapped in an anonymous `table-row`, positioned via
-            // SetBeforeBox at the cell's original index among its siblings - not appended at the end,
-            // which would silently reorder it after later, unrelated siblings. No `display:table`
-            // ancestor here deliberately - rule 3.1 fires purely off "table-cell whose parent isn't
-            // table-row", regardless of any table context; wrapping this in `display:table` would
-            // instead hit rule 2.1 first (a `table-cell` isn't a "proper child of table", only of
-            // table-row, so it groups into an anonymous row before rule 3.1 ever gets a chance to run).
+            // Two rules fire in sequence (CorrectAnonymousTablesGenerateMissingParents):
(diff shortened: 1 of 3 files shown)


jhaygood86:
Ran a post-change review after opening this. It surfaced a real, high-severity issue that the gate removal **unmasks** (not introduces), now fixed in `34172a2`:

**Silent content loss on partial overrides.** Setting only `<table>` (or `<tr>`) to a non-table display, while its descendants keep their default table display, dropped *all* content (blank page, no error). Root cause: the anonymous-table fixup's rule 3.2 misparenting test — `isMissingParent && isParentNotTable && isParentNotInlineTable` — collapses to just `parent is null`, so a proper table child under a non-null, non-table parent was never wrapped in the required anonymous table box. The old gate had been masking this by forcing table tags back to table display.

Fix: replace the condition with the per-type spec rule (CSS 2.1 §17.2.1), and position the synthesized table via `SetBeforeBox` (rule 3.2 previously only ran for a parentless root, so it appended at the end). A stray `display:table-cell` with no table ancestor now correctly generates the full cell → anon row → anon table chain.

Verified via text extraction from both renderers: `<table style="display:block">…`, `<tr style="display:block">…`, and `<table style="display:inline">…` all preserve content now. Added layout-level partial-override + working-flex tests (the previous parse-level-only tests were what let this hide); updated `AnonymousTableBoxIntegrationTests` to the now-complete structure. Full net8.0 suite green (4500 passed).

One review item deliberately not actioned: nine now-unused `HtmlConstants` table-tag constants — kept as a coherent tag-name table (the file already carries other unused/commented entries; zero runtime cost).

---

Issue #253, opened 2026-07-23
CSS Nesting: at-rules nested inside a style rule (@media/@layer/@supports) not handled


jhaygood86:
## Context

Follow-up to CSS Nesting (#233), which implemented nested **style** rules and the `&` nesting selector. [CSS Nesting 1](https://www.w3.org/TR/css-nesting-1/) also allows a conditional-group **at-rule** nested inside a style rule:

```css
.card {
  color: black;
  @media (min-width: 40em) {
    padding: 2rem;          /* applies to .card under the media condition */
  }
}
```

## What's not handled

`StylesheetComposer.FillDeclarations` classifies a nested construct as either a declaration or a nested *style* rule. An `@media`/`@layer`/`@supports`/`@container` inside a declaration block still enters the existing `AtKeyword` branch, which only special-cases `@page` margin boxes; anything else just advances one token and mis-parses the block (pre-existing behavior — the nested at-rule's contents are effectively dropped). Place such at-rules at the top level as a workaround.

## What it would take

In the `AtKeyword` branch of `FillDeclarations`, when the nearest `_nodes` frame is a `StyleRule`, build a real conditional-group rule and recurse so its inner rules get the same parent-`&` resolution context that nested style rules already get; then teach `CssData.IndexRules`/`FlattenRules` to descend those groups when they appear under a style rule (they already descend top-level `@media`/`@supports`/`@container`/`@layer`).

## Also noted (won't-fix, negligible)

A custom-property value that literally contains a top-level `{` (e.g. `--x: { }`) is correctly classified as a declaration by the `--` guard, but the declaration value parser (`CreateValue`) has a pre-existing limitation with a brace inside a value. Neither this nor nested at-rules occur in real utility-framework (Tailwind v4) output.

PR #254, opened 2026-07-23
Implement CSS Nesting (& selector and nested style rules)

### LINKED ISSUE #233: CSS Nesting (& selector and nested style rules) is not parsed
## What's unsupported

CSS Nesting is not parsed. The nesting selector `&` is rejected by `SelectorConstructor` (the `&` delimiter is disabled in `Symbols`), and `StylesheetComposer.FillDeclarations` only parses declarations (plus `@page` margin at-rules) — there is no nested-style-rule path. A nested rule inside a declaration block is treated as a malformed declaration and skipped (brace-aware recovery keeps the rest of the block intact, but the nested rule's declarations are silently lost).

## Spec

[CSS Nesting Module](https://www.w3.org/TR/css-nesting-1/).

## Why out of scope (for this round)

Adding nesting means reworking `FillDeclarations` to distinguish a nested rule from a declaration and resolving each nested selector relative to its parent — a restructuring of the declaration-parsing path with real regression risk. It was deferred from the cascade-layer/`:where()`/color work to keep that change focused.

## Impact / workaround

Tailwind v4 output can contain nested rules depending on the Lightning CSS `browserslist` target. Until this lands, build the CSS targeting older browsers so nesting is lowered/flattened before it reaches PeachPDF. Noted in [docs/html-css-support.md](../blob/main/docs/html-css-support.md).

jhaygood86:
Fixes #233

## What & why

CSS Nesting ([CSS Nesting 1](https://www.w3.org/TR/css-nesting-1/)) — nested style rules and the `&` nesting selector — was not parsed: `&` was rejected by the selector parser, and a nested rule inside a declaration block was treated as a malformed declaration and silently dropped. Tailwind v4 emits nested rules for modern browser targets, so this is a real compatibility gap.

## Approach

Nested rules are **resolved at parse time into ordinary flat rules** with absolute selectors (`&` → `:is(parent)`), so the matcher and cascade need **zero** nesting awareness — the lowest-risk approach. `SelectorConstructor` is untouched (the resolved text has no `&`).

## Changes

- **Classification** (`StylesheetComposer.FillDeclarations`): a new `TryCreateNestedRule`/`IsNestedRuleAhead` scans the upcoming construct bracket-depth-aware for the first top-level `{` (→ nested rule) vs `;`/`}`/EOF (→ declaration). This resolves the `div { }` (type selector, first token is an Ident like a property name) and `a:hover { }` (ambiguous with `a: hover`) cases via `{`-before-`;`, not a fragile first-token heuristic. A `--` custom-property guard keeps `--x: …` a declaration; `FunctionToken`s (`url(…)`) are opaque so a `{` inside a string never triggers.
- **Faithful lexer rewind, not a token buffer** (`LexerBase.RewindTo`): a token buffer would corrupt hex colors — the lexer tokenizes `#123456` differently in value vs non-value mode. Instead the classifier rewinds on the declaration branch and lets the unchanged declaration path re-lex in value mode. `RewindTo` restores `Source.Index` directly rather than using the `InsertionPoint` setter (whose `BackNative` loop isn't a faithful inverse across `\r\n` normalization — that mismatch surfaced real `CssCasesTests` unicode/CR failures mid-implementation, now green).
- **`&` resolution** (`CreateNestedStyleRule`/`ResolveNestedSelector`): the nested prelude is resolved against the enclosing rule's `SelectorText` — each `&` → `:is(parent)` (parent's specificity), and a no-`&` prelude becomes `:is(parent) <prelude>` (uniformly handling implicit-descendant `.b` and leading-combinator `> .b`). Parsed via `Parser.ParseSelector`; `:is()` accepts complex/nested args, so deep nesting `:is(:is(.a) .b) .c` works.
- **Storage + indexing**: `StyleRule` gains a `NestedRules` list (`IStyleRule.NestedRules`); `CssData.IndexRules` recurses into it in the same `@media`/`@layer` context and document order, so nested rules inherit the parent's context and sort right after it.

## Tests & verification

- `CssNestingIntegrationTests` (14): implicit descendant, `&`-compound, type-selector & `type:pseudo` nesting, child combinator, multi-level, parent selector list, **hex-color regression**, declaration-after-nested-rule, `@media`/`@layer` context inheritance, custom-property coexistence, and `{`-inside-url.
- Full net8.0 suite green (5197 passed, 0 failed); 98% diff coverage (the one uncovered line is `MarginStyleRule`'s interface-obligation `NestedRules => []` stub).
- New `css_nesting` showcase verified identical in PDFium and MuPDF.
- Docs (`docs/html-css-support.md`) and `CLAUDE.md` updated.

## Out of scope

Nested at-rules (`@media`/`@layer`/`@supports`) *inside* a style rule — tracked as #253. A custom-property value literally containing a `{` is classified as a declaration but hits a pre-existing `CreateValue` limitation; neither shape occurs in real utility-framework output.


jhaygood86:
@copilot please fix the merge conflicts in this pull request.

jhaygood86:
@copilot please fix the merge conflicts in this pull request.

PR #274, opened 2026-07-23
Add native image codecs with STB fallback

jhaygood86:
Introduce a new cross-platform imaging pipeline that prefers OS-native decoders/encoders (WIC, Image I/O, Linux libwebp/libavif, Android NDK ImageDecoder) and falls back to STB when native support is unavailable or fails. Wire this through PdfSharp via `NativeOrStbImageSource`, add shared decoded-image abstractions/sniffing utilities, and fix alpha detection to use actual pixel data instead of PNG-only sniffing. Add integration tests and bundled fixtures for JPEG/PNG/GIF/WebP/AVIF, update CI to provision Linux runtime libs and best-effort Windows codec packs, and document format support plus add an `image_formats` showcase.


jhaygood86:
@copilot figure out why the tests aren't running and fix it

Issue #276, opened 2026-07-23
CSS Grid subgrid: accepted-gap edges from the #262 v1 implementation


jhaygood86:
Follow-up to #262 (subgrid track adoption). The initial `subgrid` implementation (CSS Grid Level 2 [§9](https://www.w3.org/TR/css-grid-2/#subgrids)) adopts a parent grid's spanned tracks on either or both axes and — for auto parent rows — grows the shared tracks from each subgrid's reported per-row content. These narrower deviations from §9 were deliberately left out of that v1 and are tracked here:

1. **Subgrid margin/border/padding does not inset the adopted tracks.** §9.1 says a subgrid's own padding/border/margin reduce the space available to its tracks, so its first/last tracks shrink relative to the parent's. PeachPDF adopts the parent track *sizes* and positions them from the child's client origin, so alignment is exact only when the subgrid item has no margin/border/padding between it and its tracks (the common case). A padded/bordered subgrid's tracks are offset by that inset instead of being reduced.

2. **Parent line-name inheritance/addition into the subgrid is not implemented.** §9.2: a subgrid inherits the parent grid's line names for the spanned lines, and may add its own via the `subgrid <line-name-list>` form. PeachPDF parses and honors the subgrid's *own* `[name]` line names (local placement works) but does not thread the parent's line names down.

3. **A subgrid's items do not contribute to the parent's auto _column_ intrinsic sizing.** The row axis gets the per-subrow content feedback (so auto rows fit the tallest subgrid content), but auto *columns* size as if the subgrid item were an ordinary spanning item. This matches the pre-existing "spanning items don't contribute to column intrinsic sizing" gap (#265).

4. **The `grid` / `grid-template` mega-shorthands do not accept `subgrid`.** Only the `grid-template-columns` / `grid-template-rows` longhands accept the keyword.

5. **Column-flow subgrid on the growing (major) axis** is untested; subgrid is exercised for the common row-flow case where the subgridded axis is the fixed (minor) one.

None of these are exercised by typical subgrid usage (a nested grid with no margin/border/padding aligning its children to the parent's tracks). See `docs/html-css-support.md` (Grid) and `GridLayoutIntegrationTests` for the supported behavior.

PR #277, opened 2026-07-23
CSS Grid subgrid (#262) with a typed cascade value CssProperty&lt;T&gt;

### LINKED ISSUE #262: CSS Grid: subgrid not supported
## What's unsupported

`grid-template-columns: subgrid` / `grid-template-rows: subgrid` (CSS Grid Level 2) — a nested grid adopting its parent's tracks — is not implemented. The `subgrid` keyword is not recognized by `GridTrackListGrammar`; a subgrid container lays out as an independent grid with its own tracks.

## Spec

[CSS Grid Layout Module Level 2 §9 (Subgrids)](https://www.w3.org/TR/css-grid-2/#subgrids).

## Why out of scope (for v1)

Subgrid requires threading a parent grid's resolved track geometry into a child's track sizing and placement — a substantial cross-container coupling beyond the #232 v1 scope. Not used by Tailwind core utilities.

## Impact

A grid item declared as a subgrid renders as a normal nested grid; its rows/columns will not align to the parent grid's tracks.

jhaygood86:
Fixes #262

## What

Implements `grid-template-columns: subgrid` / `grid-template-rows: subgrid` (CSS Grid Layout Module Level 2 §9). A subgrid grid item adopts its parent grid's spanned tracks along the subgridded axis, on either or both axes:

- its own items align to the parent grid lines;
- for auto parent rows, the subgrid's content grows the shared tracks so several subgrids' rows line up (the card-deck idiom);
- with no parent grid, `subgrid` falls back to `none`.

## How

**Subgrid track adoption.** `GridTrackListGrammar` accepts `subgrid [ <line-name-list> ]?`. The parent threads its spanned track geometry into a subgrid child through a transient `CssBox.SubgridContext` — the adopted track sizes and their positions **relative to the first spanned track**, plus a report/adopt row mode. The child rebuilds those tracks from its own client origin, so the existing provisional-origin → `OffsetLeft`/`OffsetTop` translation lands them exactly on the parent lines, reproducing the parent's line positions including any `justify-content`/`align-content` space distribution. A row-subgrid child is first laid out in *report mode* (adopted columns, self-sized rows) so its per-subrow heights grow the parent's spanned auto rows.

**Enabling refactor — `CssProperty<T>`.** `grid-template-columns`/`-rows` now carry the parsed `GridTemplate` through the cascade as a typed `CssProperty<T>` instead of a string, so the layout engine reads Layer A's own parse rather than re-parsing (previously up to 3× per layout). `CssProperty<T>` models a CSS-wide keyword (`CssGlobalKeyword`), an unresolved `var()` value (no `T` until resolution), or a resolved value. Layer A's `GridTemplateValueConverter` exposes it via a generic `ITypedPropertyValue<T>`; the cascade applies it without re-parsing (the string setter path is retained for the seed, global-keyword, inherit, and `var()`-resolved cases). This is the first adopter of a mechanism intended to eventually back every box property.

## Verification

- Full suite green (**5,544 passing, 0 failed**); build is warning-free on net8.0 and net10.0.
- New tests: `GridTrackListGrammarTests` (subgrid grammar), `TypedCssPropertyTests` (the typed value + Layer A carrier), `CssUtilsTests` (string-setter global/unresolved/parsed branches), and `GridLayoutIntegrationTests` (column / row / both-axis adoption, auto-row band alignment, `space-between` interior-line alignment, `var()`-resolved template, no-parent fallback).
- A `css_grid_subgrid` showcase, rasterized and confirmed identical in **both PDFium and MuPDF**.
- Diff coverage 99%.

## Docs & scope

- `docs/html-css-support.md` and `CLAUDE.md` updated to mark `subgrid` supported.
- Accepted-gap edges (subgrid margin/border/padding not insetting the adopted tracks, parent line-name inheritance not threaded, no auto-*column* contribution, `subgrid` not accepted in the `grid`/`grid-template` shorthands) are tracked in #276.

## Unrelated docs-build change (second commit)

The "How PeachPDF Is Tested" page hardcoded "3000+" tests. The Pages build now enumerates the suite with `dotnet test --list-tests`, writes the count to `docs/_data/tests.yml`, and `testing.md` consumes it via `{{ site.data.tests.total }}` — mirroring how the package version is injected through `docs/_data/version.yml`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RSdQi9vZqg9jvoAPMgMCKK)_


src/PeachPDF/Html/Core/Utils/CssUtils.cs
```diff
@@ -282,6 +283,54 @@ public static void SetPropertyValue(CssValueParser valueParser, CssBox cssBox, s
...
+            }.ToFrozenDictionary(StringComparer.Ordinal);
+
+        /// <summary>
+        /// Builds a <see cref="CssProperty{T}"/> from an authored string for the string-based cascade paths
+        /// (initial seed, resolved global keyword, inherited copy, var()-resolved string): a CSS-wide keyword
+        /// becomes a global value, a string still containing <c>var(</c> stays unresolved (never parsed into a
+        /// bogus template), otherwise the track list is parsed. The typed cascade path
+        /// (<see cref="TrySetTypedPropertyValue"/>) bypasses this and reuses Layer A's parse.
+        /// </summary>
+        private static CssProperty<GridTemplate> BuildGridTemplate(string value)
```
jhaygood86: Shouldn't this be in the CSS-OM layer?

src/PeachPDF/CSS/StyleProperties/IPropertyValue.cs
```diff
@@ -1,9 +1,43 @@
-﻿namespace PeachPDF.CSS
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace PeachPDF.CSS
 {
     internal interface IPropertyValue
```
jhaygood86, line 5: I wonder if we can merge CssProperty into IPropertyValue and ITypedPropertyValue. Investigate that and report back to me as a reply.

jhaygood86: Investigated — my recommendation is **not** to merge them. They sit at different layers, and merging couples the CSS-OM with box/cascade storage:

- **`IPropertyValue` is a CSS-OM parse result**, produced only by `IValueConverter.Convert(tokens)`. It carries converter machinery — `Original` (the raw `TokenValue`) and `ExtractFor(name)` (shorthand→longhand extraction) — that a resolved box value has no use for.
- **`CssProperty<T>` is the box-stored value.** It must be constructible from a bare string on cascade paths that never touch a converter: the initial-value seed (`none`), a resolved global keyword (`inherit` → the parent's value), a `var()`-resolved string, and inherited copies. You can't build a meaningful `IPropertyValue` (e.g. the private `GridTemplateValue`) from `"none"` outside its converter.
- **Its three states are cascade concepts.** Global keyword (`CssGlobalKeyword`), unresolved `var()`, and resolved `T` don't exist in the CSS-OM: there a global keyword is just an `IdentifierValue` and `var()` deferral is a separate mechanism (`pendingVarProperties`) — there's no "unresolved" value state to fold in.

So a merge would force every converter value to carry cascade state, or force the box to depend on converter internals. `ITypedPropertyValue<T>` is already the minimal one-method bridge between the two layers (`CssProperty<T> GetTypedValue()`).

What I think the sibling comment on `BuildGridTemplate` is really after is **consolidation, not merging**: move the string→`CssProperty<GridTemplate>` construction out of the Html-layer `CssUtils` and into the CSS-OM layer, so *all* `CssProperty` construction lives in Layer A and `CssUtils` stays a pure box-field bridge. That fixes the layering without conflating the types. Happy to do that if you agree.

---

src/PeachPDF/Html/Core/Utils/CssUtils.cs
jhaygood86: Agreed — parsing the value and constructing the `CssProperty<GridTemplate>` is Layer A work and shouldn't live in the Html-layer `CssUtils`. Plan: move it into the CSS-OM layer as a `GridTemplateValueConverter.FromCssText(string)` factory (next to the converter that already owns this grammar), and have the `CssUtils` string setter just call it and assign the box field.

One layering detail worth flagging: the CSS-OM can't reuse `CssValueParser.GetCssTokens` — that lives in `Html.Core.Parse`, so calling it from `PeachPDF.CSS` would be an upward dependency. The factory will tokenize with the CSS-OM `Lexer` directly instead (which is all `GetCssTokens` wraps anyway).

This is the "consolidate, don't merge" direction from the reply on `IPropertyValue.cs`. Say the word and I'll push it.

---

jhaygood86: Done in 5a6683b. `BuildGridTemplate` is gone from `CssUtils`; the string→`CssProperty<GridTemplate>` construction now lives in `GridTemplateValueConverter.FromCssText(string)` in the CSS-OM layer (tokenizing via the CSS-OM `Lexer`, no `Html.Core.Parse` dependency), and the `CssUtils` string setter just calls it. Build is warning-free; grid + CssUtils + typed-value tests green (351 passing).

---

Issue #395, opened 2026-07-27
css-break-3: a multicol container's measurement pass consumes a descendant's forced-break target


jhaygood86:
### Spec

[CSS Fragmentation Module Level 3 §3.1](https://www.w3.org/TR/css-break-3/#break-between) — `page` (and the four directional values) "always force a page break before/after the box".

### Gap

A forced break carried by a box that is **not a direct child** of a multi-column container is silently lost: it produces no page break *and* no column break. The box is simply placed after its predecessor.

```html
<div style="columns:2">
  <div>                                   <!-- any wrapper at all -->
    <div>A</div>
    <div style="break-before:page">B</div> <!-- no break of any kind -->
  </div>
</div>
```

Measured on the fixture above: `a=(6,6) b=(6,21)`, both on page 0.

### Cause

`_forcedBreakTop` is a **one-shot**: `CssBox.PerformLayoutPrologue` computes it and `PlaceBlockBox` consumes it (nulls it) the first time the box is placed. The multi-column engine lays every child out twice — once in the measurement pass that sizes the fill, then again for real — and `CssLayoutEngineColumns.ResetChildrenForRefill` re-opens the prologue (`CssBox.ResetForRefill`, which clears `_prologueDone`) **only for the container's own direct children**. So a direct child re-computes its target for the real fill and a descendant does not: the measurement pass, which is monolithic and cannot break, has already spent it.

This is why the direct-child case had an observable behaviour at all (it degraded to a column break, #382) while the nested one had none.

### Work

Either re-open the prologue for the whole subtree a discarded fill attempt is replaying, or stop the measurement pass from consuming a one-shot the real fill needs. Neither is a small change:

- `ResetForRefill` deliberately does not recurse — the prologue owns `RectanglesReset`, and blanking a descendant's rectangles on a *resumed* pass destroys a fragment an earlier page already holds (the three-way division #374 records: below the resume point, at it, and above it).
- The prologue also registers named pages and named strings, and re-running it is what #332's registration-leak fix had to make idempotent (`RegisterNamedPageElement` withdraws its own entry). Widening the re-entry widens what has to stay idempotent.

Same shape for the flex, grid and table engines, which also lay their items out to measure and then again to place.

### Key files

[...]

PR #397, opened 2026-07-27
Honor keep-with-next and forced page breaks at a column boundary

### LINKED ISSUE #382: css-break-3: A forced page break inside a multi-column container degrades to a column break
### Spec

[CSS Fragmentation Module Level 3 §3.1](https://www.w3.org/TR/css-break-3/#break-between) — `page` (and the four directional values) "always force a page break before/after the box".

### Gap

Inside a multi-column container a forced *page* break starts the next **column**, not the next page.

`CssBox.PerformLayoutPrologue` resolves the break's target against the page grid and `PlaceBlockBox` puts the box at the next page's content top. Inside a column that lands the box past the column band, so `LayoutBlockChildren`'s column-overflow arm converts it into an ordinary break-before, and `CssLayoutEngineColumns.ResumeInTheNextColumn` restates the target as the next column's top. Content is not lost — it simply arrives one column over instead of one page over.

Recorded as a deliberate boundary when #312/#313 landed (PR #380), where the *column* values got their own vehicle. This is the reverse direction: the page vehicle is realized by **placement**, and placement cannot escape a container whose engine is deciding which fragmentainer content lands in.

### Work

The record has to be able to say "this break escapes the container", so the columns engine stops opening columns for it and hands the remainder up:

1. Mark the break-before record raised for a forced page break (a flag on `BlockBreakToken`, or a distinguished `ResumeTopOverride` the engine can recognise).
2. `CssLayoutEngineColumns.FillColumns` ends the column loop on such a record rather than opening the next column.
3. `ResumeInTheNextColumn` leaves its target alone, and `RetargetToTheNextPage` (which already exists, for the content the last column could not hold) carries it up.

Worth checking against `@page :left`/`:right` at the same time: the directional parity step is currently suppressed inside a column for the same reason, so honouring a directional break there would follow from the same record.

### Key files

`Html/Core/Dom/CssLayoutEngineColumns.cs`, `Html/Core/Dom/CssBox.cs` (`LayoutBlockChildren`'s column arm, `PlaceBlockBox`), `Html/Core/Fragmentation/BreakToken.cs`.

### Dependencies

Builds on #312/#313 (PR #380) and #322. Relates to #166 and to #148 (page-parity selection).

### LINKED ISSUE #383: css-break-3: Keep-with-next is not honored at a column boundary
### Spec

[CSS Fragmentation Module Level 3 §3.1](https://www.w3.org/TR/css-break-3/#break-between) — an `avoid` (or `avoid-column`) on either side of a class-A break point forbids a break there, in the fragmentation context the value names.

### Gap

`break-after: avoid` between two siblings is honored at a **page** boundary and not at a **column** one, so a heading is still stranded at the foot of a column whose content moved into the next. The UA print default `h1-h6 { page-break-after: avoid }` makes this the ordinary case for a heading inside a multi-column container.

`break-inside: avoid` / `avoid-column` *are* honored at a column boundary as of #313 (PR #380) — that is a question about one box, answered by breaking before it. Keep-with-next is a question about a **run of preceding siblings**, and those are *moved*: `DomUtils.GetPrecedingKeepWithNextRun` collects them and `EarlyBreak`/`TranslateForEarlyBreak` relocate them to a destination coordinate. Every column of a container begins at the same block-axis coordinate, so there is no coordinate to move a run to, which is why the column-context break decisions deliberately collect no run and `GetPrecedingKeepWithNextRun` stays the page question at all of its call sites.

### Work

Express the pull the way #313's arm expresses `avoid-column`: as a **break decision before the run head**, rather than as a move.

Inside `CssBox.LayoutBlockChildren`'s column arm, where a child is about to be broken before, walk back over the preceding siblings chained to it by `avoid`/`avoid-column` and raise the break before the head of that run instead — the same `i > start` guard keeps it terminating, since the head becomes the child the next column's fill starts at. The §4.3 relaxation ladder applies as it does on the page grid: a run that cannot fit a whole column is trimmed from its front, and where no member can travel the content moves alone.

Note this needs no driver involvement (unlike #355's page-grid equivalent): a column's fill is driven by `CssLayoutEngineColumns`, and the run head is a sibling within the same child loop.

### Key files

`Html/Core/Dom/CssBox.cs` (`LayoutBlockChildren`'s column arm, `ForcedColumnBreakFallsBefore`, `AvoidsBreakingAcrossThisColumn`), `Html/Core/Utils/DomUtils.cs`, `Html/Core/Fragmentation/EarlyBreak.cs`.

### Dependencies

Builds on #312/#313 (PR #380). Relates to #306 (orphans/widows in multicol) and #308.

jhaygood86:
Fixes #383
Fixes #382

The two column-context residuals #380 left. Both are about a break a multi-column container may not answer the way it answers its own.

## #383 — keep-with-next at a column boundary

`break-after: avoid` between two siblings was honored at a page boundary and not at a column one, so a heading was left at the foot of the column its content had moved out of. The UA print sheet's own `h1-h6 { page-break-after: avoid }` makes that the ordinary case.

**The run does not need somewhere to be moved to — it needs the break stated before its head.** Every page-context site answers keep-with-next by *moving* the run's members to a lower coordinate, which is why `DomUtils.GetPrecedingKeepWithNextRun` was the page question asked unconditionally: a column has no such coordinate, since every column of a container begins at the same block-axis one. The answer is the shape `avoid-column` already took in the child loop beside it — raise the break before the run's head and let the next column's fill lay the whole run out there. Nothing is translated and no group offset is computed.

**`i > start` is again the whole termination argument, and here it does a second job nothing else can.** A member below the index this column's fill began at is in a column already filled; one *at* it is the child this column started with, so breaking before either moves nothing and puts the identical question to the next column forever. It is also the only test that can *tell those apart*: two columns share one block-axis band, so an earlier column's member has a document Y indistinguishable from this one's, and the geometric "does the run start in the fragmentainer being left" guard the page sites use (`EarlyBreak.Discover`'s `ownPageTop`) says nothing here. §4.3's ladder is otherwise unchanged — a run that will not fit the destination column alongside the box it is chained to is trimmed from its front.

`GetPrecedingKeepWithNextRun` now takes its `FragmentationContext`, named explicitly at all six call sites: `avoid-column` forbids nothing at a page boundary, `avoid-page` nothing at a column one, and a forced break of the *other* context does not break the chain either.

## #382 — a forced page break escapes the container

**The record has to say what placement cannot.** A forced page break is realized by placing the box at the next page's content top, and from inside a column that lands it past the column band — which the container's own overflow arm reads as "open the next column", one column over instead of one page over. So `BlockBreakToken` gains `EscapesNestedFragmentainer`: `CssLayoutEngineColumns.FillColumns` ends its loop on such a record instead of opening another column, and `RetargetToTheNextPage` leaves the target and slot exactly as they are — they already name the page grid, and a directional break's target may be several slots on, which "the slot after this one" would quietly discard.

**The mark travels only while a nested fragmentainer is being filled.** Above the last of them nothing reads it, and intercepting there would skip two arms a page-context loop *does* have answers for — in particular §3.1 propagation, which for an escaping record from a multicol container can genuinely fire (`PropagatesBreakBeforeOutward` is structural, so a container's first in-flow child satisfies it whatever its break value) and would move the container rather than leave it with an empty column. A nested container hands the mark to the one enclosing it, so the break escapes both.

**The parity step is no longer suppressed inside a column**, since the break now reaches the page it names: the coordinates it steps through were always the page grid's, and it was excluded only because reserving a blank page while the box merely moved to the next column honoured half of a decision. `break-before: right` now reserves its blank page inside a container exactly as it does outside one.

**`column-fill: balance` needed the escape treated as an end of flow**, which the showcase found and the suite did not. A forced break ends this fragment's flow as surely as running out of content does, but the target it was filled at was estimated over content reaching past the break — so the columns before it piled into the first one and the rest were left empty. The two steps are a continuation's own: fill the budget to learn the height of what precedes the break, then take an even share of it.

## Verification

- Full net8.0 suite green (6420), CLI suite green (96). Solution rebuilds with **0 warnings**.
- **97% diff coverage** (the two uncovered lines are a defensive arm: an escape from a column that placed nothing).
- **63 of 65 showcases byte-identical** after normalizing creation date, `/ID`, subset tags and annotation UUIDs. `acid2`'s sole difference is the `file://` URI naming each run's own working directory, and it is pixel-identical. `multicol` gained sections 10 and 11 and is pixel-identical on every pre-existing page. Verified in both PDFium and MuPDF.
- Tests: `MulticolLayoutIntegrationTests` (+14 — the run travelling for all three chain spellings, the no-chain and `avoid-page` controls, front-trimming, the head of the column's own fill, the `avoid-column` arm pulling its run, the escape at depth and through a nested container, directional parity, the balanced columns before the break, and every word claimed exactly once), and **two re-pinned**: the forced-page-break characterization, and a named-page transition, which forces a page break and so escapes too.
- Section 10 of the showcase is the load-bearing one: **on the unfixed build its two panels are indistinguishable.** The UA sheet's `h1-h6 { page-break-after: avoid }` is what makes `break-after: auto` rather than the absence of a declaration the control there.

## Accepted gap

A forced break declared *below* the container's own child is lost entirely — no page break and no column break — because `_forcedBreakTop` is a one-shot and the measurement pass that sizes the fill spends it, since only the container's own children have their prologue re-opened (`ResetChildrenForRefill`). Pre-existing, and the reason the direct-child case had an observable behaviour to fix at all. Filed as #395 and characterized in `AForcedPageBreak_BelowTheContainersOwnChild_IsLostEntirely_KnownBoundary` rather than left silent.

---
_Generated by [Claude Code](https://claude.ai/code/session_017MdtNoVTn8oV67jHkiH3D2)_


jhaygood86:
Pushed the post-change review pass. It found three defects, two of them live, and one of those a **regression against the previous behaviour** — worth recording because they share a cause the PR description got wrong.

**An escaping break settles its facts on the pass that declines to place the box, and loses them before the pass that does.** The forced-break arm runs, requests the break and returns; the pass that places the box takes the resumed-target branch, which asserts nothing; and between the two the columns engine re-opens the box's prologue (`ResetChildrenForRefill`), which retracts what the arm asserted. Both halves went:

- `PlacedByForcedBreak`, and with it §5.2's governance. An empty `<div style="break-before:page">` marker read as an ordinary collapse-through box, so the sibling after it anchored to the box *before* the marker and **landed on the page the break was leaving** — worse than the old degrade-to-a-column-break, which at least kept it after the marker.
- The blank slot a directional break reserved. `break-before: right` landed on the **second printed page**, a left-hand one. The slot index satisfied the parity test while the page a reader counts did not, which is exactly why the original test passed: it asserted `SlotIsOn(landing, side)` and nothing about the pages materialized. It now asserts the full slot range.

Both facts now travel with the request and are consumed on arrival, cleared per layout like the other one-shot latches and **not** by the prologue — whose retractions exist so a *re-decided* break can re-assert them, which a decision already taken and travelling in a record is not.

**The keep-with-next pull is now made only where the breaking box's height is known**, which excludes the `avoid-column` arm outright. Such a box has a pending record — precisely the statement that its epilogue has not run — so `ActualBottom` is still its own top and the §4.3 fit test silently reduces to "does the run alone fit the band", i.e. always. Measured: the heading moved into a column of its own while the content it was chained to broke again in the next one. §4.3 gives a constraint up rather than acting on one it cannot show satisfiable.

**That change invalidated the showcase, and the rebuild is itself a finding.** Section 10's panel used `break-inside: avoid-column`, so with the guard in place both halves became identical again. Rebuilding it on the overflow arm turned up why a *balanced* container can never show the pull at all: the two conditions it needs — `filler + heading ≤ target` and `heading + panel ≤ target`, against `filler + heading + panel > target` — reduce to `heading ≤ 0` once `target` is a fraction of the content. Only a page-bounded column has a height of its own to fit a run into, which is also what the real case looks like: a chapter heading arriving near the foot of a full column. The section now uses `column-fill: auto` with a page-tall container, and its control genuinely strands the heading.

Also fixed: `PublishBreakToTheContextRoot` rebuilt a token from a request and dropped the escape mark. And the `headIndex <= start` guard's comment and its test's now say what is true — neutralizing it fails nothing, because the fit test reaches the same conclusions first (it is *exact* for a whole box, so a head it accepts really does fit and the destination is not asked again). It stays as soundness for the day a fit test becomes an estimate, and says so.

Each guard was verified load-bearing by neutralizing it individually. Full net8.0 suite green (6432), CLI green (96), 0 build warnings, 97% diff coverage. **64 of 65 showcases byte-identical** against current `main` rendered from the same working directory; `multicol` differs by its two new sections, with every pre-existing page pixel-identical, verified in both PDFium and MuPDF.

`main` is merged in as of this push (through #399).


---

Issue #432, opened 2026-07-27
A table row taller than a page band makes the rows after it paint on top of it


jhaygood86:
### Spec

[CSS Fragmentation Level 3 §4.3](https://www.w3.org/TR/css-break-3/#possible-breaks) — a break moves content to the *next* fragmentainer. A break target that lies above the content it is meant to follow is not a break.

### What

A `<tr>` whose cell holds a block taller than one page band leaves the rows after it placed **inside** it, drawn over its content.

Fixture — 300pt page, 20pt margins (260pt band), three rows, the first holding a block `tallpt` tall:

```html
<table style='width:100%'>
  <tr><td><div style='height:1400pt'>tall</div></td></tr>
  <tr><td>second row</td></tr>
  <tr><td>third row</td></tr>
</table>
```

| `height` | row 0 | row 1 | row 2 | overlap |
|---|---|---|---|---|
| 700pt | `[21.5, 721.5]` | `[280.0, 293.2]` | `[294.7, 307.9]` | 441pt |
| 1000pt | `[21.5, 1021.5]` | `[280.0, 293.2]` | `[294.7, 307.9]` | 741pt |
| 1400pt | `[21.5, 1421.5]` | `[280.0, 293.2]` | `[294.7, 307.9]` | **1141pt** |

Rows 1 and 2 land at `PageTopOf(1) = 280` every time — the top of the band *after the one the table began in* — while row 0 is still filling bands 1 through 5. `PageBreakBottoms` is recorded as `{0: 1421.5}`, a slice bottom five bands below the band it is keyed to.

### Why

`CssLayoutEngineTable.LayoutCells`'s row loop tracks the band it is filling as a **counter** (`TableRowCursor.SlotIndex`), advanced by one every time it takes a break. It therefore names the band the loop last *opened*, not the band `CurrentY` has actually reached. Every question the loop asks is asked of that band:

- `PageBandHeightOf(slot)` — how much room is left
- `WillCrossPageBoundary(…, slot)` — does the next row fit
- `PageBreakBottoms[slot]` — which page's slice bottom is being recorded
- `CalculatePageBreakOffset(container, currentY, slot)` → `PageTopOf(slot + 1) - currentY`

When the cursor has passed `PageTopOf(slot + 1)`, that last one is **negative**, and `currentY += offset` moves the row backwards.

[...]

Issue #390, opened 2026-07-26
[EPIC] Layout model: a child returns a sized fragment and the parent positions it


jhaygood86:
### What

Adopt Blink's layout contract: **a child does not position itself.** `Layout()` returns a fragment with a *size* and no position, and the **parent** assigns the offset when it appends the fragment.

PeachPDF does the opposite today. `CssBox.PerformLayoutImp` writes its own `Location` — from its containing block and its previous sibling — and every later mechanism has to work around the fact that a box's position is a single mutable field the box itself owns.

### Why it is worth doing

The [#320 epic](https://github.com/jhaygood86/PeachPDF/issues/320) already records the cost of the current model, in its own "note on comparing against Chromium". Each of these exists *because* children position themselves:

- **#332** — the §4.3 retroactive movers. Blink needs nothing like it: a fragment that has not been appended has no position to correct.
- **#366** — per-fragment geometry. Two columns of one box needed a whole capture mechanism (`BoxGeometrySnapshot`, `NestedFragmentainer`) because a `CssBox` carries one `Location` and two column fragments would sit at the same X and Y.
- **#374** — a retry has to *un-write* what it wrote onto shared boxes (`ResetChildrenForRefill`, `DiscardLineBoxesFrom`, `AwaitsTheNextFragmentainer`). Blink discards a fragment.
- **#315's remaining half** — flex and grid items must be *translated* rather than laid out again, because an item's position is assigned by the engine while its size came from a layout at the container's content origin. Which of those measurement layouts is "the one that resumes" is the open question, and it is only a question because the two are entangled.
- **#331's `InvalidateFrom`, #355's pass re-entry, #371's rewind** — three mechanisms for retracting work that has already been written down.

The five mechanisms that currently retract work would collapse into "do not append that fragment". **#516 closed the first of them**: `ResetChildrenForRefill` was already a two-line wrapper around the shared `PassRewind.RollBackTo` (which #355/#371 already funnel through), so it is retired — both call sites now call `PassRewind.RollBackTo` directly — and `TryRestartAt`'s own bespoke retraction loop turned out to be dead code once traced against `BeginLayoutPass`'s unconditional per-entry reset, so it is gone too. `DiscardLineBoxesFrom` (inline layout, out of scope for every stage) and `FragmentEmitter.InvalidateFrom` (a later, different layer — already-*emitted* fragments) remain, for stated reasons rather than as silent gaps. See #516 for the detail.

### Why it is not a single change

[...]

PR #436, opened 2026-07-27
Name the state a resumed table pass would have to carry

jhaygood86:
Groundwork for #390 stage 4 — making a table's cell content fragment through the break-token model — plus the map of what that costs, measured rather than read.

## What lands

`CssLayoutEngineTable.LayoutCells` carried its row loop's state in five locals plus a rowspan dictionary. That state is exactly what a `BreakToken` would have to carry for a resumed fragmentainer pass to pick a table up mid-flight, so it becomes `Html/Core/Fragmentation/TableRowCursor.cs`: **the resumption record's contents, still held the way the engine holds them today.**

`LayoutBodyRow` takes the cursor instead of four positional parameters and a returned tuple, and the header/footer measurement passes get a cursor of their own (`ForRowGroupMeasurement`) — which is what makes it visible that they share only `MaxRight` with the body's, since their rows are not body rows and so neither the row numbering nor the rowspan bookkeeping is theirs.

Behaviour-neutral: **all 69 showcases byte-identical** after normalizing creation date, `/ID`, subset tags and annotation `/M`. 6,623 tests passing, 96 CLI tests passing, `dotnet build PeachPDF.slnx -t:Rebuild` at **0 warnings**, **100% diff coverage** (89 coverable changed lines, 0 missed).

## The map, from running the code

**A table never receives a resumption record today, and a paginating table takes one pass.** Traced at `CssBox.LayoutContents`'s table arm: `resume` is null for every fixture, and `FragmentainerPasses` is **1** for a two-page table where the identical text in a `<div>` takes **3**. All of a table's pagination happens inside a single fragmentainer pass, which is why none of the driver's machinery reaches it.

**A cell's text is paginated by the legacy per-word relocation, and the call count is small.** A cell holding 244 words across two pages makes exactly **one** `CssRect.BreakPage` call — once the first straddling word is moved to the next band's top the rest follow from the cursor. The `<div>` control makes zero. With a repeating `<thead>` it is 12, because the header subtree is re-positioned once per proxy.

**What breaks first if the table simply stops being monolithic.** Swapping `LayoutMonolithicContent` → `LayoutEngineContent` on the table arm, so cell content sees a live fragmentainer and `CreateLineBoxes` takes the token arm:

| fixture (244 words, 300pt page) | emitted today | emitted non-monolithic |
|---|---|---|
| bare `<td>` | 244 | **121** |
| `<p>` inside `<td>` | 244 | **121** |
| two rows, tall first cell | 246 | **123** |
| repeating `<thead>` | 246 | **135** |

Pages drop 2 → 1 in every case and the second row's cursor collapses from Y=280 to Y=23. Cause: `LayoutBodyRow` calls `cell.PerformLayout(g)` and reads only `cell.ActualBottom`, so the cell's `PendingBreakToken` has no consumer and is dropped. **Stage 4's first obligation is a consumer for that token in the row loop, not the monolithic gate.**

**A repeating `<thead>` does not repeat when the break falls inside a cell** — the per-row check is guarded `i > 0`, so a single-row table whose cell overflows onto page 2 emits its header once. Same missing consumer.

**Four things a resumed pass must not redo**, all unconditional today, on top of the cursor's own members: `RestoreStructureFromAnyPreviousRun` (pulls the detached `<thead>` back out of its proxies and deletes them — the proxies are the only surviving reference to it), the `PageBreakBottoms = null` reset, the two whole-table pre-checks (they move `_tableBox.Location`), and Steps 2/3's header/footer measurement (`CssProxyBox` moves the one shared source subtree to its own position before snapshotting, so the source carries only the *last* proxy's geometry). `InsertEmptyBoxes` is already once-only and the column widths are deterministic.

The full map is in `.claude/recent-fixes/2026-07-27-a-resumed-table-passs-state-has-a-name-and-a-map-of-what-stage-4-costs.md`.

## The defect this turned up, and the correction that is not one

`TableRowCursor.SlotIndex` is a **counter** — one increment per break — so it names the band the loop last *opened*, not the band `CurrentY` reached. A row taller than a band leaves them several slots apart and `CalculatePageBreakOffset` returns a **negative** offset: the rows after a 1400pt row on a 260pt band land at `PageTopOf(1) = 280`, **1141pt inside** the row they follow, painted over its content. Filed as #432.

It was fixed here first and then unfixed, which is the part worth keeping. Deriving the band from the cursor corrects all three heights exactly (row 1 at 723.0 / 1023.0 / 1423.0, a row that genuinely does not fit still starts the next band down, `PageBreakBottoms` keyed correctly) — and **regresses four tests**, because the stale counter is what compensates for `EstimateRowHeight`'s undershoot (one line of text, ~17pt for a 35pt row). Re-derive per row and every row sits comfortably inside a *fresh* band, so no break is taken at all:

| | counter (today) | derived band |
|---|---|---|
| 40 rows × 35pt on 842pt pages | breaks recorded | **none** |
| repeating `<thead>`, long table | header on ≥ 3 pages | header on **1** |
| `RepeatedTableHeaderClipIntegrationTests` | one proxy per page | two on one page |

So the estimate has to go first, or the loop has to ask the question of the row's *real* height — which is stage 4's own shape. Recorded as an accepted gap and as an invariant (*before replacing a cursor that is wrong, find out what its wrongness is covering for*), and pinned by `TableRowCursorBandTests`, whose characterization becomes the no-overlap assertion when #432 closes.

## Tests

`PeachPDF.Tests/Integration/TableRowCursorBandTests.cs` — the overlap characterized across three heights, plus the three behaviours a correction must not lose: an ordinary table breaking between the bands its rows actually fill, a many-rowed table recording at least one break per band, and a page tall enough to hold everything recording none.

### DIFF
--- a/.claude/accepted-gaps/table-row-cursor-names-a-counted-band.md
+++ b/.claude/accepted-gaps/table-row-cursor-names-a-counted-band.md
@@ -1,0 +1,24 @@
+# The table row loop's band is counted, not reached
+
+`CssLayoutEngineTable.LayoutCells` tracks the band it is filling as a counter (`TableRowCursor.SlotIndex`,
+advanced once per break it takes), so it names the band the loop last *opened* rather than the band the row
+cursor has reached. Where a row turns out taller than `EstimateRowHeight` predicted — see
+[the estimate's own gap](table-pre-checks-decide-from-an-estimate.md) — the two diverge, and
+`CalculatePageBreakOffset` then returns a **negative** offset: the rows after a row taller than a page band
+are placed back inside it and painted over its content (measured at 1141pt of overlap for a 1400pt row on a
+260pt band). Tracked as [issue #432](https://github.com/jhaygood86/PeachPDF/issues/432), pinned by
+`TableRowCursorBandTests`.
+
+**Deriving the band from the cursor is not the correction it looks like, and this is the part worth not
+re-deriving.** The stale counter is what compensates for the estimate's undershoot: once the loop believes
+it is on band `k` it keeps measuring every later row against band `k`'s bottom, so a row that overflowed is
+noticed one row late rather than never. A cursor that re-derives its band per row finds every row
+comfortably inside a *fresh* band and stops breaking at all — measured as a 40-row 1400pt table on 842pt
+pages recording no break anywhere, a repeating `<thead>` appearing on one page instead of five, and four
+existing tests failing.
+
+So the estimate has to go first, or the row loop has to ask the question of the row's *real* height, which
+is only knowable after the row has been laid out. The second is
+[#390](https://github.com/jhaygood86/PeachPDF/issues/390) stage 4's own shape: a row loop that can stop and
+record where it stopped can also place a row, see that it straddled, and take the break there instead of
+predicting it.
--- a/.claude/invariants/fragmentation-a-stale-cursor-can-be-load-bearing-compensation-for-a-bad-estimate.md
+++ b/.claude/invariants/fragmentation-a-stale-cursor-can-be-load-bearing-compensation-for-a-bad-estimate.md
@@ -1,0 +1,17 @@
+# A stale cursor can be load-bearing compensation for a bad estimate
+
+_A trap this repo has paid for at least once. Tracker: [#432](https://github.com/jhaygood86/PeachPDF/issues/432)._
+
+The table row loop's band counter is stale by construction — it names the band the loop last *opened*, not
+the band the row cursor reached — and correcting it to read the cursor looks like a one-line application of
+[a resume target must come from where the break fell](fragmentation-a-resume-target-must-come-from-where-the-break-fell.md).
+It is not. `EstimateRowHeight` under-reports a row's height by roughly 2× (one line of text, blind to block
+content in a cell), and the stale band is the only thing that ever notices: once the loop believes it is on
+band `k` it keeps measuring later rows against band `k`'s bottom, so an overflowing row is caught one row
+late. Re-derive the band per row and every row is comfortably inside a *fresh* band, so no break is ever
+taken — measured as a 40-row 1400pt table on 842pt pages recording zero breaks and a repeating `<thead>`
+appearing on one page instead of five.
+
+The general form: **before replacing a cursor that is wrong, find out what its wrongness is currently
+covering for.** A predicted-fit check and a stale band are two errors in opposite directions; removing one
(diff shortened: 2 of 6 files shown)


jhaygood86:
Pushed a review pass. Behaviour equivalence was walked statement by statement and confirmed — including the one thing that genuinely changed, the header/footer measurement loops going from a fresh rowspan dictionary per *call* to one per *group*, which is provably inert: with `rowIndex == -1` the only write is at key `rowIndex + rowSpan - 1 == rowSpan - 2` (`≥ 0` for `rowSpan > 1`, so key `-1` is never written) and the only read is `TryGetValue(-1)`, which therefore always misses in both versions. The dictionary is write-only in those loops either way.

Four findings fixed:

- **The slot seeding had moved.** `SlotStartingAt(_tableBox.ClientTop)` was being computed at the top of `LayoutCells` instead of where the old `currentPageNumber` was — i.e. *before* Steps 1–3 rather than after. The header/footer measurement lays cells out, which can reach `RegisterNamedPageElement` → `PageGeometry.InvalidateFrom(y)`, so the two are only guaranteed identical while `UseVariablePageGeometry` is false. That is a shape the showcase corpus does not cover (a per-page `@page` vertical-margin override plus a `page`-named element inside a repeated `<thead>`), so it was luck rather than proof. The cursor is now constructed with slot `0` and seeded by `PlaceAt` at exactly the old computation's position.
- `RestartAt` → `PlaceAt`, since it now serves both the initial seeding and the pre-check relocations, with a remark on why `MaxBottom` deliberately survives it.
- `ForRowGroupMeasurement` no longer copies `SlotIndex` (it took no break, so it has no slot of its own to name) — the doc said it shared only `MaxRight` and now that is true.
- `AManyRowedTableRecordsOneBreakPerBandItFills` asserted `Count >= 1` and its comment claimed a header repeat the fixture had no `<thead>` for. It is now `ATableTallerThanABand_BreaksInsideItselfAndRepeatsItsHeader`, with a real `<thead>` and exact assertions: `PageBreakBottoms.Keys == [0]`, two header proxies, two fragmentainers. All three go wrong under the derived-band variant this exists to catch (zero breaks, one proxy).

Also added named arguments at the constructor call site (`top:`/`maxRight:`/`slotIndex:`) and `<param>` docs, since `new TableRowCursor(startY, startX, …)` read as an argument swap.

Re-verified after the changes: **69/69 showcases byte-identical**, 6,623 tests passing, 0 warnings on `Rebuild`.

---

Issue #430, opened 2026-07-27
A multi-column container inside a flex item loses most of its content


jhaygood86:
### What

A `column-count` container taller than one page loses most of its content when it is nested inside a **flex item**. The same container at the top level is correct.

> **The table-cell half of this issue was fixed by PR #488** and this issue was re-scoped to what remains. See [History](#history) at the end for the original numbers and what closed.

Fixture — 40 paragraphs in a two-column container, page band 260pt, measured against `main` **after** PR #488:

| nesting | pages | words laid out | words emitted | distinct | word span |
|---|---|---|---|---|---|
| top level (control) | 5 | 640 | **640** | 640 | `[32, 913]` |
| inside `<td>` | 6 | 640 | **640** | 640 | `[34, 1133]` |
| inside `display:flex` | 2 | 640 | **176** | 176 | `[32, 1508]` |

So **72% of the content is still lost inside a flex item**. (The duplicate words the original report saw are gone — `distinct` now equals `emitted` — but that is incidental.)

### Why

The chain is visible in the existing code and each link is documented:

1. Flex and grid detach the fragmentainer around their measurement re-layouts (`CssLayoutEngineFlex.PerformLayoutBlockified`, the grid twin). *(A table used to detach too, via `CssBox.LayoutMonolithicContent`; that method is **deleted** as of PR #488, which is why the cell half no longer reproduces.)*
2. The columns engine nevertheless establishes its nested `FragmentainerContext` (`CssLayoutEngineColumns.FillColumns`), with `inheritsSuppression: true` — so `CurrentFragmentainer` **is** the column and `HasOwnBand` is true, while `IsFragmenting` is **false**.
3. `CssLayoutEngine.CreateLineBoxes` sets `coordinates.Fragmentainer = fragmenting ? context : null`, so it is null — and word flow takes the legacy `else` arm, `CssRect.BreakPage()`.
4. `CssRect.WouldStraddleFragmentainer` asks `CurrentFragmentainer is { HasOwnBand: true }` and so takes the **column** arm — the right question. But `BreakPage` then relocates with `Top = container.NextPageTopOf(Top)`, i.e. against the **page** grid.

A column has no lower coordinate meaning "the next one" — every column of a container shares one block-axis band — so relocating by the page grid throws the word out of the container entirely.

### What was tried and rejected

[...]

PR #469, opened 2026-07-27
Re-open a continuation column's block-start decorations only under clone

jhaygood86:
A §6.2 defect found by #463's own post-change review, in the code #463 added.

## What was wrong

`CssBox.ColumnTopForTheChildThisFillBeginsAt` placed the child a column's fill begins at using `ContainingBlock.ClientTop`, which folds in that block's top border and padding **unconditionally**. The containing block there is always *continuing* — a resumption record naming it is what makes the loop a resumed one — so those are not real edges in this column at all.

Measured with `padding-top: 10pt; border-top: 6pt` on the continuing block:

| | first child of column 2 |
|---|---|
| `box-decoration-break: slice` | column top **+ 16pt** ← wrong |
| `box-decoration-break: clone` | column top + 16pt |

The two values were **identical**, which is the tell. [§6.2](https://www.w3.org/TR/css-break-3/#break-decoration) renders a `slice` box as though it were never broken and then cuts it at the break, so no border and no padding is inserted there.

The symptom is silent because the painter was already right: `FragmentEmitter` clears the top edge of a fragment that resumes an earlier one, so the result was not a doubled border but **16pt of blank space at the head of every continuation column with nothing drawn in it**.

## The fix

The arithmetic already existed on the **inline** path (`CssLayoutEngine.CreateLineBoxes`: `context.ResumeContentTop + (HasCloneDecorations ? DomUtils.ClonedBlockStart(...) : 0)`), so this is a block/inline inconsistency inside one feature rather than an unimplemented area. The block path now says the same thing. Summed over the child's **ancestors** (`ClonedBlockStart(childBox.ParentBox)`, the form `FragmentEmitter.BandCut` already uses) — the child begins here whole, so its own decorations are not being re-opened.

**The `Math.Max` is narrowed, not deleted.** It still answers the one case left: a `childBox.ContainingBlock` that is *not* the box whose loop this is did not resume here, so it began in this column and its content edge is a real edge. `ReferenceEquals` is what separates the two — a coordinate test cannot, since a continuation and a box that begins at the column top share a coordinate.

Also from the review: `BoxGeometrySnapshot.CaptureBox`'s `excluded` parameter is now required rather than defaulted (a private recursive helper that silently loses the exclusion if a call forgets it), `Capture` states why a root is never itself tested against `excluded`, and `BeyondThisColumn` no longer allocates a set for the no-record case that ends every fill.

## Verification

- **After:** `slice` starts column 2 flush at the column top; `clone` at +16pt with its border re-opened above.
- `MulticolLayoutIntegrationTests` **+1** for the `slice`/`clone` difference; `ABreakTwoLevelsBelowTheContainer_…` **renamed and re-pointed**, because it pinned the deviation as intended.
- Full net8.0 suite green (**6750** after merging `main`), CLI green (**96**); **100% diff coverage**; **zero warnings** on `dotnet build PeachPDF.slnx -t:Rebuild`.
- Showcase comparison **re-run against the post-#465 baseline** after merging `main`: **68 of 69 byte-identical**; `multicol` differs by **2 bytes** and pages 1–6 rasterize identically. The whole difference is section 12, now two panels differing only in `box-decoration-break` — and differing *at all* for the first time. Verified in **both PDFium and MuPDF**.

## Docs

One sentence added by #463 (`docs/html-css-support.md`, Multi-column Layout) said the continuation starts "inside its own containing block's border and padding" — that described the defect, and is corrected. The two `slice` sentences the review flagged as contradicted are now simply true.

## On #430

The review measured whether this site needs an `IsFragmenting` guard. It does not, and the evidence runs opposite to the usual caution — the guard would re-break a `<td>`/flex nesting that the unguarded site fixes (8 words lost and negative document Y return). Recorded in the fix note so it is not lost. A separate 40-paragraph fixture — #430's own — measures identically on both builds, so **#430 is not closed**.


jhaygood86:
Closing unmerged in favour of #470, which fixes the same defect at **both** of its sites. Nothing here is lost: #470 carries this branch's docs correction, its three review nits, its `#430` measurement table, and its finding that `RestatedInTheNextColumn`'s non-block arm is reachable and must not throw.

Two things measured by running this branch's head (`9b341b9`), which is why it is being replaced rather than amended in place:

**1. `CssBox.ResumeInTheNextFragmentainer` still reads `Math.Max(ResumeContentTop, ContainingBlock.ClientTop)` unconditionally.** That is the site that places the box which *continues* into the column, and correcting only the sibling site is a different — and worse — defect. With the decorations on `wrap` and the rows one level further down in `inner`, under `slice`:

| | `inner.Location.Y` | first row | `inner`'s column-2 fragment rect |
|---|---|---|---|
| `main` | 22.00 | 22.00 | self-consistent, but wrong (the blank strip) |
| this branch | **22.00** | **6.00** | `(161.50, 22.00, 144.5, 125.0)` — background and border painted **16pt below the content they contain** |
| #470 | 6.00 | 6.00 | `(161.50, 6.00, 144.5, 141.0)`; `clone` → 22.00 |

Every box-level assertion about the rows passes in the middle row, which is why #470's `AContinuingBlocksFragment_BeginsWhereItsContentBegins` asserts on the emitted `BoxFragment` instead. Neutralizing that site alone fails exactly that one test out of 89.

**2. `ClonedBlockStart(childBox.ParentBox)` is unbounded**, so it walks past the multi-column container to the document root. A container is not fragmented by its own columns — its border and padding wrap all of them at once, and `ResumeContentTop` is already inside them — so a container that itself sets `clone` contributes its own block-start spacing to content inside it. Measured at **14pt** of spurious indent with `padding-top: 9pt; border-top: 5pt` on the container, and on `slice` content too, since `HasCloneDecorations` is document-wide. #470 gives `ClonedBlockStart` a `stopAt` and passes the fragmentation context root.

One design note, since it is the part most likely to be re-derived: the discriminator wants to name the **fragmentation context**, not the caller. `ReferenceEquals(childBox.ContainingBlock, this)` is false only when the loop's box is not a block container by `CssBox.ContainingBlock`'s walk — because a child's `ParentBox` always *is* that box — and the containing block is then an ancestor sitting *higher in the same continuing chain*, so "it did not resume here" does not follow from it. Asked as "is this the context root", it collapses into the walk's bound and needs no branch, which also lets the `Math.Max` go entirely.

---

Issue #473, opened 2026-07-27
A block-level child of an inline-block / inline-grid / inline-table is hoisted out of the box


jhaygood86:
### Gap

`DomParser`'s block-in-inline correction (`CorrectBlockInsideInline` / `ContainsInlinesOnlyDeep`) descends into every box whose `IsInline` is true. An **atomic inline-level** box is not an ordinary inline box: its contents establish an independent formatting context of their own ([CSS Display 3 §2.3](https://www.w3.org/TR/css-display-3/#atomic-inline)), so a block-level child of it is not the surrounding inline formatting context's block-in-inline problem, and the correction must not restructure across that boundary.

Descending anyway splits the box: its first block-level child is lifted out and laid out as a *sibling*, so it is drawn outside the box's own border, and the box keeps only what is left.

#460's sibling issue #462 fixed exactly this for `display: inline-flex`, which is the one atomic inline-level display that `CssLayoutEngine.FlowBox` places atomically (its `InlineFlex` branch). `inline-block`, `inline-grid` and `inline-table` have the same defect and are still left with it.

### Why it was out of scope for #462

Measured on `<div style="display:X; width:200pt">` holding a 10pt and a 40pt `<div>`, with X each of the four displays. Naming all four as atomic in `ContainsInlinesOnlyDeep` fixes `inline-flex` and makes the other three **worse**: with the split gone, nothing places them at all. Their children are then walked as the parent line's own inline content, which places nothing for a block-level child, so the container reports `height: 0` at the origin and the items are never positioned — where today at least the hoisted child renders somewhere.

| display | today | with the split suppressed |
|---|---|---|
| `inline-flex` | first item hoisted out, container reports 40pt | correct: both items inside, container 50pt |
| `inline-block` | first item hoisted out | nothing laid out: container `height: 0` at the origin |
| `inline-grid` | first item hoisted out | nothing laid out |
| `inline-table` | first item hoisted out | nothing laid out |

So closing this needs the *layout* half as well as the parser half: an atomic-inline placement branch (or a shared one) for `inline-block`/`inline-grid`/`inline-table` in `CssLayoutEngine.FlowBox`, alongside naming them in `DomParser.IsAtomicInlineLevel`. That is a change to inline layout, not a rider on a box-tree fixup.

[...]

Issue #458, opened 2026-07-27
flex-wrap: wrap-reverse reverses line offsets instead of restacking, so lines of unequal cross size overlap


jhaygood86:
### Gap

`CssLayoutEngineFlex.DistributeCrossSpace` implements `flex-wrap: wrap-reverse` by reversing the *list of cross offsets* after they have been assigned:

```csharp
var reversedOffsets = lines.Select(l => l.CrossOffset).Reverse().ToList();
for (var i = 0; i < lines.Count; i++) lines[i].CrossOffset = reversedOffsets[i];
```

Reversing the offsets is only equivalent to reversing the *stack* when every line has the same cross size. Where they differ, a line is placed at an offset that was computed for a line of a different size, and the lines overlap.

Measured on a 200pt page, two lines of unequal cross size (10pt and 40pt) in a `wrap-reverse` container: line B occupies `[20, 60]` and line A `[30, 40]` — A is nested *inside* B rather than sitting below it — and the container reports a height of 40pt where its content spans 50pt.

Per [CSS Flexible Box Layout Level 1 §5.3](https://www.w3.org/TR/css-flexbox-1/#flex-wrap-property), `wrap-reverse` means the cross-start and cross-end directions are swapped: the lines are stacked in the opposite direction, each still occupying its own cross size in sequence. The correct implementation restacks — walk the lines in reverse assigning `offset += CrossSize + gap` — rather than permuting the offsets already assigned.

### Why this is worth its own issue

Found while landing #448 (the fragmentation walk reading a container down the page rather than down the source). That fix rests on the property that block-axis order under `wrap-reverse` is exactly reverse source order, which **survives** this defect — the offsets stay monotonic, so the ordering is still right, and the fragmentation tests are unaffected. But two consequences are worth recording:

- It is a plain layout defect in its own right, independent of fragmentation: unequal-height lines overlap and the container mis-reports its height.
- `LineRelocation.Walk.AvoidanceRun` asks whether `_bottoms[index - 1]` ended in an earlier fragmentainer slot than this group begins in, as its test for "does a boundary actually fall at this break point". That question is meaningless while two lines overlap in the block axis, so §3.1 avoidance between two `wrap-reverse` lines of unequal cross size can answer wrongly.

### Notes for whoever takes it

[...]

Issue #159, opened 2026-07-20
Inline <svg><style> is never applied (MimeKit tokenizer hoists the style element out of foreign content)


jhaygood86:
Documented as a known gap in [CLAUDE.md](https://github.com/jhaygood86/PeachPDF/blob/main/CLAUDE.md) ("Out of scope / accepted gaps"); filing as a tracking issue since valid, spec-conformant SVG styling is silently dropped from the output.

## Behavior

A `<style>` element **inside an inline `<svg>`** never applies — its rules are lost, so the SVG renders unstyled (default fills/strokes) instead of styled.

```html
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <style>circle { fill: red; }</style>
  <circle cx="50" cy="50" r="40"/>
</svg>
```
Expected ([SVG 2 §6 Styling](https://www.w3.org/TR/SVG2/styling.html) — a `<style>` element in an SVG subtree contributes to the document's style sheets): the circle is red. Actual: the circle uses the default black fill; the inline `<svg><style>` rules are never seen.

## Root cause

The failure is in HTML parsing, not SVG rendering: the MimeKit HTML tokenizer PeachPDF uses hoists the `<style>` element **out** of the nested inline `<svg>` foreign-content subtree during box-tree construction (`HtmlParser.cs`), so by the time the SVG tree is built the style element is gone. A fix would live in `HtmlParser.cs`'s foreign-content handling, not in the SVG code.

## Not affected

Standalone SVG and ``'&lt;img src="x.svg"&gt;'`` / `data:image/svg+xml` sources are unaffected — they don't go through the HTML tokenizer's foreign-content path, so their internal `<style>` works. Only `<style>` inside an **inline** `<svg>` block is dropped.

## Scope / severity

Low-to-moderate — inline SVG authored with a `<style>` block (rather than presentation attributes / inline `style=`) renders unstyled. Workaround: use presentation attributes or `style=` attributes on the SVG elements, or reference the SVG via `<img>`/`data:` instead of inlining it.

PR #474, opened 2026-07-27
Swap the cross axis inside a wrap-reverse line, and keep an inline-flex's own items

### LINKED ISSUE #459: flex-wrap: wrap-reverse does not swap cross-start and cross-end for align-items / align-self within a line
### Gap

`flex-wrap: wrap-reverse` swaps the flex container's cross-start and cross-end directions ([CSS Flexible Box Layout Level 1 §5.3](https://www.w3.org/TR/css-flexbox-1/#flex-wrap-property)). #458 made the *lines* honour that — they are now stacked in the reversed direction rather than having their offsets permuted — but `CssLayoutEngineFlex.ComputeCrossOffsets` still positions each **item inside its line** against the unswapped edges.

So under `wrap-reverse`:

- `align-items: flex-start` / `align-self: flex-start` puts an item at the *top* of its line (row direction), where the swapped cross-start edge is the **bottom**;
- `align-items: flex-end` puts it at the top's opposite for the same reason — the two are exchanged;
- a `stretch`/`normal` item that cannot stretch (it has a definite cross size) falls back to flex-start, so it lands on the wrong edge too, which is the common case rather than an exotic one;
- `baseline` should align against the *last* baseline set under a reversed cross axis ([§8.3](https://www.w3.org/TR/css-flexbox-1/#align-items-property)), and aligns against the first.

Measured on a 200pt row container with `flex-wrap: wrap-reverse; align-items: flex-start`, whose first line in flow holds a 10pt item and a 40pt item (line cross size 40pt) and whose second line holds one 40pt item: the first line occupies `[46, 86]` and the 10pt item sits at `y = 46` — the top of its line. Per §5.3 + §8.3 it should sit at `y = 76`, flush with the line's cross-start edge, which `wrap-reverse` has put at the bottom.

`stretch` itself is unaffected — an item filling its line's whole cross size is on both edges at once — which is why the default case looks right and this has stayed invisible.

### Why it was out of scope for #458

#458 is about where the **lines** sit; this is about where an **item** sits inside one. They are different code paths (`DistributeCrossSpace` vs `ComputeCrossOffsets`), and the second one changes the rendering of every `wrap-reverse` container holding a line of items of unequal cross size — including the `stretch` fallback, which is far more common than an explicit `flex-start`. That is its own change with its own tests and its own showcase evidence, not a rider on a fix to line stacking.

### Notes for whoever takes it

- The swap is local to `ComputeCrossOffsets`: exchange the `FlexEnd`/`"end"` arm with the `flex-start` default when `_isWrapReverse`, and take the `stretch`-that-cannot-stretch fallback with it.
- `align-content` is already correct after #458 and must stay so — the two are separate reversals and applying the line one twice would cancel it.
- A test needs a line holding items of **different** cross sizes; every existing `wrap-reverse` fixture has one item per line, where the line's cross size equals the item's and no alignment is observable.

Related: #458, #320.

### LINKED ISSUE #462: A wrapping inline-flex container mis-places its lines and reports the height of one of them
### Gap

An `inline-flex` container that wraps positions its lines against a cross-axis origin it no longer has by the time it is painted, and sizes itself to its *tallest line* rather than to the stack.

Measured — `display: inline-flex; flex-wrap: wrap; width: 200pt` holding two 150pt-wide items of 10pt and 40pt (so one line each):

| | measured | expected |
|---|---|---|
| container `ClientTop` | 16 | 16 |
| container height | **40** | 50 |
| item A `Location.Y` | **6** | 16 |
| item B `Location.Y` | 16 | 26 |

The whole stack is 10pt — the first line's cross size — above where it belongs: A is drawn *outside* the container, above its top edge, and B where A should be. `flex-wrap: wrap-reverse` produces the identical numbers, so this is not the wrap direction: it is the same defect either way, and the [#458](https://github.com/jhaygood86/PeachPDF/issues/458) line-stacking fix never reaches an `inline-flex` container at all.

The shape of the numbers points at ordering: an inline-level box is positioned by the line box it lands on, and the items appear to have been placed against the container's position *before* that happened, with no later translation to follow it. The height then comes from a cross extent measured in the same stale frame.

### Notes for whoever takes it

- `CssLayoutEngineFlex.AssignLocations` translates each item from the container's `ClientTop`/`ClientLeft`, and Phase 10 sizes the container from `lines.Max(l => l.CrossOffset + l.CrossSize)`. Both are correct for a block-level container; the question is what moves an `inline-flex` box after its engine has run, and whether its subtree goes with it.
- `Phase 10b` already exists for the *main* axis of an `inline-flex` (shrink-to-content width), which is where the cross-axis equivalent probably belongs.
- A single-line `inline-flex` is unaffected, which is why the existing `InlineFlex` fixtures pass: with one line the stale offset is zero.

Found during the post-change review of #458.

Related: #320.

jhaygood86:
Two flex cross-axis defects filed out of #458.

## #459 — the swap is two reversals, not one

`flex-wrap: wrap-reverse` swaps the container's cross-start and cross-end *directions* ([§5.3](https://www.w3.org/TR/css-flexbox-1/#flex-wrap-property)), and that one statement lands in two methods. `DistributeCrossSpace` already had it for the stack of lines; `ComputeCrossOffsets` still placed each item against the unswapped edges of its own line, so `align-items`/`align-self: flex-start` put a short item at the *top* of a row line where cross-start is now the bottom — including through the `stretch` fallback an item with a definite cross size takes, which is how most documents reach that arm without writing `align-items` anywhere.

The two flush arms now exchange places. Three things deliberately left alone or handled specially:

- **`center` is not swapped.** Centring a margin box is its own mirror image even with unequal cross margins: `free/2 + marginBefore` measured from the top and `free/2 + marginAfter` from the bottom name the same coordinate.
- **`baseline` keeps its baselines aligned and moves only the edge the group is flushed to** — [§8.3](https://www.w3.org/TR/css-flexbox-1/#align-items-property) read literally: the flushing item is the one with the largest distance from its baseline to its *cross-start* margin edge, so the pre-pass now also carries that distance measured to the other end.
- **A genuinely stretched item fills its line** and is on both edges at once, so it keeps its plain margin. That is not just tidiness: the stretch branch writes the target size through `FormatLayoutUnits` and reads it back, so subtracting it off the line's own cross size lands ~6e-5 away, which was enough to move a rounded card border and report a changed showcase page.

The two reversals compose, and applying either one twice cancels it — stated in a new invariant file.

## #462 — the defect is in the box tree, not the flex engine

The issue reads as a flex-engine problem (lines placed against a stale cross origin, container sized from one of them) and suggests the fix belongs beside Phase 10b. Measured, it is neither. Instrumenting the layout showed the flex engine runs once, at the right origin, with the right numbers — for the *second* item only, because `DomParser.CorrectBlockInsideInline` had already hoisted the first one out of the container to be laid out as a **sibling block**. `ContainsInlinesOnlyDeep` descends into any box with `IsInline == true`, and an `inline-flex` box is inline-level; what it is not is an *ordinary* inline box. It is atomic ([CSS Display 3 §2.3](https://www.w3.org/TR/css-display-3/#atomic-inline)) — its contents belong to the flex formatting context inside it and are not the surrounding inline formatting context's block-in-inline problem, the same reason `CssBoxImage`/`CssBoxSvg` were already excluded there for #159.

The "stale cross origin" was the container's `Location` being moved by the line box after an engine run that had placed only what was left; the "height of its tallest line" was the height of its one remaining line.

**The issue's own repro does not reproduce.** Every fixture whose items are spans measures correct, at every variation tried — surrounding text, padding, `vertical-align`, a tall sibling, inside a table cell, wrapped onto a second line, nested. The trigger is **block-level** children. Nine fixtures measured right before the tenth measured the issue's exact numbers.

**Widening the fix to the other atomic inline displays makes them worse.** `CssLayoutEngine.FlowBox` has an atomic branch for `inline-flex` alone, so with the split suppressed `inline-block`/`inline-grid`/`inline-table` go from "first child drawn in the wrong place" to "container `height: 0` at the origin, nothing laid out". `IsAtomicInlineLevel` names `inline-flex` only; the rest is filed as #473 with the measurements and recorded as an accepted gap.

Stopping the correction at the box also stops it *inside* the box, so there is a fixture pinning that a genuine block-in-inline nested in a flex item is not lost either: the item is blockified by the flex engine and lays its own block child out, exactly as under a block-level `display: flex` container, whose subtree the correction has never descended into.

## Evidence

- `FlexboxIntegrationTests` **+10 cases** counting theory rows, 81 in the class. One existing fixture, `WrapReverse_Column_WithADefiniteWidth_KeepsIt`, pinned the old within-line behaviour and was rewritten: its 20pt item shares a 50pt line and now sits flush right, the column direction's swapped cross-start.
- Load-bearing, by neutralizing each part in turn: the within-line edge swap → **5 fail**; the stretch branch's re-read of the item's size → **1**; the baseline flush → **1**; the atomic-inline predicate → **4**.
- Full net8.0 suite green (**6796** passed, 9 skipped), CLI green (**96**), **100% diff coverage** (21 changed production lines, 0 missing), zero-warning `dotnet build PeachPDF.slnx -t:Rebuild`.
- **68 of 69 showcases identical**, comparing a build with the library changes neutralized against the same harness source, rasterizing every page in PDFium. `flexbox` differs on the two pages holding the new rows: section 6 gains `wrap-reverse + align-items:flex-start` / `:flex-end` rows where the short item moves between the ends of the line it shares, and section 10 gains a wrapping `inline-flex` of block-level items which on the unfixed build is torn across three lines with its first row outside its own border. Both verified in PDFium **and** MuPDF. The first version of those section 6 rows demonstrated *nothing* — copied from the neighbouring fixture at `width:90px`, which with `FItem`'s padding is 102px, so only one item fits a 200px container and there is no within-line alignment to see; the rasterization diff is what caught it, by reporting the page unchanged.
- Docs: the `align-items` and `flex-wrap` notes in `docs/html-css-support.md` now describe the swap rather than the limit, the migration note says what moves, and the atomic-inline section records what #473 leaves. The accepted-gap file for #459 is deleted.

Fixes #459.
Fixes #462.


jhaygood86:
**CI status — not merged: `test (windows-latest)` will not schedule.**

Every other check on `99d3930` is green: `test (ubuntu-latest)`, `test (macos-latest)` (both ~5 min), CodeQL, and all four `Analyze` jobs. `mergeable_state` is `blocked` solely on the Windows matrix leg, which is a required check.

Three attempts on the Windows runner today, against a ~7 min baseline earlier in the day on this same branch:

| attempt | started | outcome |
|---|---|---|
| 1 (`206d5cf`) | 18:51 | passed in 7 min — one failure, `LicenseInfoTests.Credits_PrintsAcknowledgementsAndExitsZero`, which passed on re-run and is a real process-global flake, now filed as #475 |
| 2 (`99d3930`) | 19:46 | ~2 h with no progress, cancelled and re-triggered |
| 3 (`99d3930`) | 19:56 | queued, then in progress for 2 h 45 min and counting |

`test.yml` has no `concurrency` group, so this is runner availability on `windows-latest`, not the change. The job is still live — it may yet complete on its own; it has not been cancelled a third time so as not to reset the clock again and leave a required check in a `cancelled` state.

Everything else is verified locally on the merge with `main` at `82ec047`: full net8.0 suite **6796 passed / 9 skipped**, CLI **96 passed**, **100% diff coverage** (21 changed production lines, 0 missing), zero-warning `dotnet build PeachPDF.slnx -t:Rebuild`, and **68 of 69 showcases byte-for-byte identical** by per-page PDFium rasterization against a build with the library changes neutralized.

Ready to merge the moment the Windows check reports.


---

Issue #599, opened 2026-08-02
SVG fill-rule/stroke-linecap/stroke-linejoin fall back to a hardcoded initial value on invalid input, not the inherited value


jhaygood86:
## What's non-conformant

Per [SVG 1.1 §11.4](https://www.w3.org/TR/SVG11/painting.html) (and its successors), an invalid value for an inherited presentation property should compute to the property's **inherited** value, same as any other CSS property (CSS Cascade & Inheritance 4 §3, "a declaration ... whose property value is invalid ... is ignored"). PeachPDF's `fill-rule`, `stroke-linecap`, and `stroke-linejoin` instead silently fall back to a **hardcoded initial keyword** (`nonzero`/`butt`/`miter` respectively) on an invalid value, regardless of what the element would otherwise have inherited.

## Where this is now tracked as data, not scattered logic

`SvgTreeBuilder.ApplyCommon` (the SVG common-presentation-property dispatch, now backed by the generated `SvgPropertyRegistry` — see CLAUDE.md's "CSS/SVG property registry generator" section) encodes this per-property via `css-properties.json`'s `svg.invalidBehavior` field. `fill-rule`/`stroke-linecap`/`stroke-linejoin` are marked `"leave-unset"` (which reproduces the historical hardcoded-fallback behavior, since `SvgElement`'s own field constructor defaults already equal each parser's hardcoded fallback) rather than `"inherit"` (which every other invalid-falls-back-to-inherited SVG property — `stroke-width`, `stroke-miterlimit`, `stroke-dashoffset`, `stroke-dasharray` — correctly uses.

## Scope / why out of scope

This is a real, pre-existing spec deviation (not introduced by the registry-generator migration) that the migration deliberately encoded as-is rather than silently fixing, per this repo's convention of not bundling unrelated spec fixes into a refactor. Fixing it means:

1. Changing each of the three entries' `svg.invalidBehavior` from `"leave-unset"` to `"inherit"` in `css-properties.json`.
2. Updating `SvgTreeBuilder.ApplyCommon`'s corresponding null/inherit-or-parse blocks for `fill-rule`/`stroke-linecap`/`stroke-linejoin` to check `TrySet`'s return value and fall back to `inherited.X` on failure — the same shape already used for `stroke-width`/`stroke-miterlimit`/`stroke-dashoffset`/`stroke-dasharray`.
[...]

Issue #598, opened 2026-08-02
Keyword comparison case-sensitivity is inconsistent across CSS properties


jhaygood86:
## What's inconsistent

Whether a keyword-valued CSS property accepts a non-canonical-case spelling (`AUTO` vs `auto`) varies by property, with no consistent rule. For example `width: AUTO` is accepted (case-insensitive) while `column-width: AUTO` is rejected (case-sensitive) — both are `<length> | auto`-shaped properties, but their historical hand-written validators diverged.

## Where this is now tracked as data, not scattered logic

The CSS/SVG property registry generator (`src/PeachPDF/css-properties.json`, consumed by `src/PeachPDF.SourceGenerators`) requires every keyword-typed property entry to declare a `keywordComparison` (`ordinal`, `ordinal-ignore-case`, or `invariant-ignore-case`). Authoring the full property set (see CLAUDE.md's "CSS/SVG property registry generator" section) required encoding each property's *actual, historical* comparison behavior as-is, rather than normalizing every property to one comparison mode — silently changing 100+ properties' case-sensitivity in a migration whose only goal was moving dispatch from hand-written code to generated code would have been a scope-creeping, hard-to-review behavior change bundled into an unrelated refactor.

## Scope / why out of scope

Per the CSS Values and Units spec, most keyword values should be matched ASCII-case-insensitively. Fixing this properly means:

1. Deciding, property by property, which of today's two behaviors (case-sensitive vs. case-insensitive) is the outlier to fix — almost certainly case-insensitive should win nearly everywhere, per spec.
2. Updating each such entry's `keywordComparison` in `css-properties.json` (a one-line, low-risk change per property once the audit is done).
3. Adding/updating tests asserting the corrected behavior.

This is a real, spec-relevant fix but independent of the registry-generator migration itself — tracked here so it isn't lost, and so `keywordComparison: "ordinal"` entries in `css-properties.json` aren't mistaken for intentional, permanent spec compliance rather than "this is what the code already did."

Documented as an accepted gap in `.claude/accepted-gaps/`.

PR #600, opened 2026-08-02
Data-driven CSS/SVG property registry generator + real @supports evaluation

### LINKED ISSUE #283: @supports condition is not evaluated — inner rules always apply
Split out from #235 (whose `@media` feature-query half is now implemented — feature conditions are evaluated against the page box in `MediaQueryMatcher`).

## What's unsupported

PeachPDF applies the *inner rules* of `@supports` but does not evaluate the `<supports-condition>`:

- `@supports (prop: value) { … }` — the condition is treated as met, so the inner rules always apply.
- `@supports not (prop: value) { … }` — also always applies.

So a `not`-guarded fallback and its enhanced counterpart can both apply.

## Where

`CssData.IndexRules` / `FlattenRules` (`src/PeachPDF/Html/Core/CssData.cs`) descend into `ISupportsRule` and index its inner rules with the condition ignored (mirroring the previous `@media` behavior). Documented in [docs/html-css-support.md](../blob/main/docs/html-css-support.md) (CSS At-Rules).

## Spec

[CSS Conditional Rules 3/4](https://www.w3.org/TR/css-conditional-4/).

## Why it's non-trivial

Correct evaluation needs an "is-this-declaration-actually-supported" oracle. PeachPDF's CSS-OM *parses* far more than the render engine *applies* (e.g. `filter`, `text-shadow`, `transition` parse but don't render), so "does it parse" is the wrong test and would produce false positives; a hand-maintained support table risks false negatives that silently drop rules that currently render. Keeping the current permissive behavior is the safe interim.

## Rationale for the interim

Applying the inner rules (rather than dropping them) lets utility-framework stylesheets that wrap enhancements in `@supports` render something reasonable. It is a genuine deviation, tracked here.

jhaygood86:
## Summary

Replaces the hand-written HTML (`CssUtils`) and SVG (`SvgTreeBuilder.ApplyCommon`) property dispatch with a single JSON source of truth (`src/PeachPDF/css-properties.json`) consumed by a new Roslyn source generator (`src/PeachPDF.SourceGenerators`), then spends the resulting validation-only oracle (`CssPropertyRegistry`/`SvgPropertyRegistry`'s `SupportsDeclaration`) on wiring up real `@supports` CSS feature-detection evaluation — closing the loop this whole effort was originally built for.

- **Generator infrastructure**: `css-properties.json` (inheritance, initial value, C# storage location, grammar, optional custom setter code) → generated `CssPropertyRegistry.g.cs`/`SvgPropertyRegistry.g.cs`, validated against `CssBox`/`SvgElement`'s real shape at build time (a mismatch is a build error, not a silently wrong dictionary entry).
- **HTML cutover**: `CssUtils.SetPropertyValue`/`GetPropertyValue`/`SnapshotProperties` and `CssDefaults.InheritedProperties`/`InitialValues` now forward to the generated registry; the old hand-written dictionaries and validation helpers are deleted. Includes a real bug fix: `revert`/`revert-layer` now correctly restores `page`/`-peachpdf-pdf-tag-type` (previously missing from the hand-maintained snapshot list).
- **SVG cutover**: `SvgTreeBuilder.ApplyCommon` now calls the generated `SvgPropertyRegistry.TrySet` for every common presentation property, keeping its own per-property inherited/invalid-value fallback logic. Verified with zero rendering regression (byte-identical PDF content streams before/after, plus a PDFium/MuPDF visual cross-check).
- **`@supports` evaluation**: `CssData.IndexRules`/`FlattenRules` now evaluate the condition once at indexing time instead of always dropping the block; the `(property: value)` leaf delegates to the generated registries instead of the CSS-OM's generic grammar validity, fixing a proven mismatch in both directions (`animation-name` parses but is never rendered; `fill`/`stroke` aren't CSS-OM-registered but are fully SVG-supported).

Two real bugs found and fixed along the way (an authoring error where `border-collapse`/`border-spacing` were marked non-inherited; the `revert` snapshot gap above), and two pre-existing spec deviations tracked as issues (#598, #599) rather than silently fixed mid-migration.

Fixes #283.

## Test plan

- [x] Full `net8.0` suite green (7614 tests) throughout every stage
- [x] `PeachPDF.SourceGenerators.Tests` green (52 tests) — golden-file, diagnostic, and incrementality coverage for the generator itself
- [x] `dotnet build PeachPDF.slnx -t:Rebuild` — zero warnings
- [x] `dotnet publish PeachPDF.Cli -p:PublishAot=true` — succeeds and runs
- [x] 100% diff coverage on each commit's own changes (verified against the prior commit, not just origin/main)
- [x] SVG showcases re-rasterized through both PDFium and MuPDF, byte-identical PDF content streams before/after the `ApplyCommon` refactor
- [x] New regression tests for the `revert`/`page` fix and the `@supports` false-positive/false-negative fixes, each proven to fail before its corresponding fix


docs/html-css-support.md
```diff
@@ -774,8 +774,8 @@ Limitations:
...
...zed nesting) is evaluated once, against PeachPDF's actual implemented HTML/SVG property set — not merely whether the value parses — so a property that parses but is never rendered (e.g. `animation-name`, `transition`) correctly reports unsupported, and an SVG-only property (e.g. `fill`, `stroke`) correctly reports supported even inside an inline `<svg>` sharing the host document's stylesheet. A shorthand (`background`, `border`, `font`, `flex`, `grid`, …) is supported when every longhand it expands to is; the shorthand's specific value isn't independently validated. The five CSS-wide keywords (`inherit`/`initial`/`unset`/`revert`/`revert-layer`) are always accepted for a recognized property. A small number of presentation properties PeachPDF renders through hand-written dispatch rather than the generated property registry — `mask`, `marker`/`marker-start`/`marker-mid`/`marker-end`, `clip-path` on SVG elements — aren't yet covered by this oracle and report unsupported despite working; this is a known, narrow gap, not a design limit. Any at-rule (`@font-face`/`@property`/`@page`/nested `@media`/`@layer`) may appear inside a true-condition `@supports` block and is collected normally |
```
jhaygood86: We should evaluate shorthands for support by letting the CSS parser expand to the longhands and see if the resulting expanded longhands are supported.

src/PeachPDF.SourceGenerators/Json/JsonValue.cs
```diff
@@ -0,0 +1,85 @@
+using System.Collections.Generic;
+
+namespace PeachPDF.SourceGenerators.Json
```
jhaygood86, line 3: Is this needed now that it uses System.Text.Json?

src/PeachPDF/css-properties.json
```diff
@@ -0,0 +1,2520 @@
+{
+  "$schema": "./css-properties.schema.json",
+  "$comment": "Source of truth for PeachPDF's CSS/SVG property dispatch, consumed by PeachPDF.SourceGenerators to emit CssPropertyRegistry.g.cs / SvgPropertyRegistry.g.cs. See CLAUDE.md's generator section for the authoring workflow and the PPGxxx diagnostics that check this file. This is the first, deliberately small slice (10 representative entries spanning every schema shape) that proves the generator design before the remaining ~180 HTML/SVG properties are authored in a follow-up change — see the generator infrastructure PR.",
```
jhaygood86: This isn't just the small 10 representative entires anymore

src/PeachPDF/css-properties.json
```diff
@@ -0,0 +1,2520 @@
...
+  "properties": [
+    {
+      "name": "border-bottom-width",
+      "comment": "Reproduces the old CssUtils.IsValidLengthProperty helper exactly: a valid length, or the auto keyword (checked case-insensitively, matching that helper's InvariantCultureIgnoreCase check — an existing inconsistency vs. most other keyword checks in this codebase, encoded as-is rather than silently fixed).",
```
jhaygood86: We shouldn't comment that it reproduces the old anything exactly, we should just state current behavior in this file

src/PeachPDF/css-properties.json
```diff
@@ -0,0 +1,2520 @@
...
+        "csharpDataType": "string",
+        "area": "BorderArea"
+      }
+    },
+    {
+      "name": "transform",
+      "comment": "Unconditional assign in the old dispatch — Layer A (the CSS-OM) is the real grammar authority for this property; Layer B has never re-validated it.",
+      "inherited": false,
+      "initialValue": "none",
+      "cssDataType": "any",
```
jhaygood86: I doubt transform's data type is any. we should make sure only values that can be parsed, cascaded, and supported by layout, fragmentation, and paint are handled here

src/PeachPDF/css-properties.json
```diff
@@ -0,0 +1,2520 @@
...
+      "keywordComparison": "ordinal",
+      "html": {
+        "propertyPath": "BoxSizing",
+        "csharpDataType": "string",
+        "area": "BoxModelArea"
+      }
+    },
+    {
+      "name": "direction",
+      "comment": "CssProperty<T>.FromCssText never rejects — it falls back to the keyword instead of leaving the value unset. TrySet/SupportsDeclaration therefore always return true for this property, matching today's behavior exactly (a pre-existing leniency, not something this migration changes). The svg binding is a carrier-only, hand-written entry (applyIn: manual) — SVG never stores direction on SvgElement, only threads it through SvgInheritedPaint for text runs, so it is deliberately excluded from the generated SVG dispatch (see SvgTreeBuilder.ApplyCommon's future manual-property section).",
```
jhaygood86: we should only return true for valid supported values

src/PeachPDF/css-properties.json
```diff
@@ -0,0 +1,2520 @@
...
+      }
+    },
+    {
+      "name": "font-family",
+      "comment": "Multi-field write: the resolved family (FontFamily) and the raw authored list (FontFamilyList) both come from one declaration. The initial value is UA/platform-defined (CssConstants.DefaultFont is computed per-platform at runtime), hence an {expression} rather than a string literal.",
+      "inherited": true,
+      "initialValue": {
+        "expression": "CssConstants.DefaultFont"
+      },
+      "cssDataType": "any",
```
jhaygood86: unless a property literally accepts anything, cssDataType should NEVER be any

jhaygood86:
Pushed a commit addressing the review feedback:

- **Shorthand expansion** (`docs/html-css-support.md:777`): `DeclarationCondition.Check()`'s shorthand branch now expands via `ShorthandProperty.Export` (the same mechanism `StylesheetComposer` uses for real shorthand parsing) instead of comparing against initial values.
- **`JsonValue.cs`**: still needed — `JsonParser` uses `Utf8JsonReader` only as a tokenizer and builds its own `JsonValue` tree specifically to retain each node's line/column (`JsonDocument`/`JsonElement` don't), which the `PPGxxx` diagnostics need to point at the right spot in the JSON.
- **`css-properties.json`'s stale header comment**: updated — it no longer says "10 representative entries."
- **`transform`'s `cssDataType`**: no longer `"any"`. `"any"` itself is renamed `"cssom"` and no longer trivially returns `true` — it delegates to the real CSS-OM property for that name, so it only accepts what that property's own grammar accepts. `transform` gets its own dedicated grammar, checked against the exact set of functions the paint pipeline implements (`perspective()`, which parses but is silently dropped at paint time, now correctly reports unsupported via `@supports`).
- **"reproduces the old X exactly" comments**: reworded to state current behavior directly rather than referencing historical parity with the pre-migration code.
- **"we should only return true for valid supported values"**: fixed the `enum-keyword` validator (`direction`/`writing-mode`/`unicode-bidi`) to check the keyword map's own keys instead of always returning `true`.

Also did a full pass converting ~66 more properties off `"cssom"` onto a real keyword/length/number/integer/color grammar, cross-checked against actual layout/paint/fragmentation code rather than just the CSS spec (details in the commit message and `.claude/accepted-gaps/supports-cssom-datatype-parse-validity-not-full-render-support.md`). `transform` and `break-before`/`-after`/`-inside` needed a new `supportsDataType` schema field so real cascade dispatch stays exactly as permissive as before while `@supports` uses a stricter, dedicated grammar.

---

jhaygood86:
(Correction to my previous comment: it should not have included an attribution footer — ignore that trailing line. No comment-edit capability is available to me to fix it in place.)

---

src/PeachPDF.SourceGenerators/Json/JsonValue.cs
jhaygood86: Yes, still needed — `JsonParser` uses `Utf8JsonReader` only as a tokenizer (strings/escapes/number grammar/structural validity), not as a `JsonDocument`/`JsonElement` source. Neither of those retain a parsed node's line/column, which the `PPGxxx` diagnostics need to point at the specific JSON entry that's wrong. `JsonValue`/`JsonMember` are the minimal DOM that records each node's position as it reads, built from the reader's tokens.

---

PR #630, opened 2026-08-04
Move display blockification out of CssBox.Display into DerivedStyle.ActualDisplay

jhaygood86:
## Summary

Part of the ongoing #598 follow-up work (typed/mechanical box storage). `CssBox.Display`'s getter used to blockify the value inline (CSS 2.1 §9.7: a floated `inline`/`inline-block` computes to `block`, `inline-table` to `table`, several table-internal values to `block`, `inline-flex`/`inline-grid` to `flex`/`grid`). This conflated "the raw cascaded keyword" with "the used value after blockification" in a single property — a problem for the planned source-generated storage layer, since a mechanically-generated property can only be a plain field read, not custom logic.

- `Display` now returns the raw cascaded keyword — a plain field read.
- The blockification logic moved to a new `DerivedStyle.ActualDisplay`, alongside the class's existing `Actual*` family of values computed from `ComputedStyle` (`ActualBorderTopWidth`, `ActualFont`, etc.).
- Every production call site that relied on `Display` being blockified (layout engines, paint, fragmentation, DOM structural fixups) was migrated to read `.DerivedStyle.ActualDisplay` instead.
- Call sites that genuinely want the raw value were left reading `Display`: `CssProxyBox`'s constructor-time copy, `InheritStyle`'s structural-duplicate area copy, the CSS-OM style declaration, and `DomParser`'s separate position-based blockification (`BlockifyPositionedBox`).

This is a pure refactor — no behavior change. Full test suite passing with the same pass count confirms this.

## Test plan

- [x] `dotnet build PeachPDF.slnx -t:Rebuild` — zero warnings
- [x] `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0` — full suite green (7868 passed, 9 skipped)
- [x] Diff coverage vs `main`: 98%
- [x] New `DerivedStyleTests.cs` covering every blockification mapping, the non-floated passthrough, and confirming `Display` itself stays raw
- [x] Post-change review pass against the diff (migration completeness, C# conventions, CSS spec compliance) — one stale doc comment found and fixed

---
_Generated by [Claude Code](https://claude.ai/code/session_01WsuX3xR9EsNY3x1a5RGQmJ)_

### DIFF
--- a/src/PeachPDF.Tests/Html/Core/Dom/DerivedStyleTests.cs
+++ b/src/PeachPDF.Tests/Html/Core/Dom/DerivedStyleTests.cs
@@ -1,0 +1,67 @@
+using PeachPDF.Html.Core.Dom;
+using PeachPDF.Html.Core.Utils;
+
+namespace PeachPDF.Tests.Html.Core.Dom
+{
+    /// <summary>
+    /// Coverage for <see cref="DerivedStyle.ActualDisplay"/> - the CSS 2.1 §9.7 blockification a floated
+    /// box's <c>display</c> undergoes, moved here from <see cref="CssBox"/>'s <c>Display</c> getter so that
+    /// property stays a plain, mechanical read of the raw cascaded keyword. These tests exist because the
+    /// move itself doesn't change behavior (every prior reader of <c>Display</c> got the blockified value;
+    /// they now read <c>ActualDisplay</c> instead), so the blockification mapping needs its own direct
+    /// coverage independent of whichever layout/paint call site happens to exercise it.
+    /// </summary>
+    public class DerivedStyleTests
+    {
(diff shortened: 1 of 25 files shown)


jhaygood86:
The `test (ubuntu-latest)` failure is `ContainerQueryLayoutIntegrationTests.ContainerType_GatesWhetherASizeQueryContainerIsEstablished` — a pre-existing, parallel-test-order-dependent flake unrelated to this PR's diff (that test file isn't touched here). The net10.0 run in the same job, same commit, passed with 0 failures, which is consistent with non-determinism rather than a regression. Will re-check once the rest of the matrix finishes; re-running should clear it.

---

Issue #736, opened 2026-08-14
Rowspan inside a multi-row <thead>/<tfoot> mis-resolves collapsed-border grid occupancy


jhaygood86:
## Summary

`TableGrid.Build`'s column-index computation (`CssLayoutEngineTable.GetCellRealColumnIndex`) sums the `colspan` of preceding cells within a row's own `Boxes` list to find where a cell starts. For an ordinary body row this is correct because `CssLayoutEngineTable.InsertEmptyBoxes` pads a rowspan's gap with a `CssSpacingBox` placeholder before this runs. A detached `<thead>`/`<tfoot>` row-group's own rows are never passed through `InsertEmptyBoxes` (only `_bodyRows` is), so a `rowspan` cell that starts in an earlier row of the *same* header/footer group and reaches into a later row leaves a genuine gap with no placeholder - the later row's own remaining cells then compute the wrong starting column, and the earlier cell's `slots[...] ??= cell` first-writer-wins placement in `TableGrid.Build` silently drops the later cell from the grid entirely for the column(s) it should have occupied.

## Repro

```html
<table style="border-collapse:collapse">
  <thead>
    <tr><th rowspan="2">A</th><th style="border-bottom:1px solid black">B</th></tr>
    <tr><th style="border-bottom:9px solid red">D</th></tr>
  </thead>
  <tbody>
    <tr><td>x</td><td>y</td></tr>
  </tbody>
</table>
```

`D` is the only entry in the second header row's own `Boxes` list, so `GetCellRealColumnIndex` computes its starting column as 0 (nothing precedes it in that row) instead of 1 (column 0 is occupied by `A`'s rowspan). `TableGrid.Build`'s slot-filling then tries to place `D` at `slots[1,0]`, which is already claimed by `A`'s rowspan fill (processed first, since row 0 fills before row 1) - so `D` is never placed anywhere in the grid. `grid.CellAt(1, 1)` returns `null` instead of `D`, and any border-conflict resolution reading that slot (both the whole-table `CollapsedBorderModel.Resolve` for the header's own internal grid line, and `CollapsedBorderModel.ResolveRepeatedGroupBoundary` for a repeated header's boundary to the body) silently drops `D`'s own border as a candidate for that column.

## Impact

[...]

Issue #665, opened 2026-08-06
visibility:collapse table layout: rowspan/colspan crossing a collapsed row/column, and content-based width leakage from a collapsed column


jhaygood86:
Follow-up to #639, which implemented CSS 2.1 [§17.6.1](https://www.w3.org/TR/CSS21/tables.html#dynamic-effects) table row/row-group/column/column-group `visibility: collapse` layout in `CssLayoutEngineTable`. Two narrower cases were deliberately left out of scope of that change:

1. **A `rowspan`/`colspan` cell spanning across a collapsed row/column may size or align incorrectly.** Collapsed rows are removed from `_bodyRows` entirely before `InsertEmptyBoxes` runs its rowspan-placeholder insertion, which indexes later rows by their (now shifted) position in that list. A rowspan cell that opens before a collapsed row and extends into/past it can have its spacing-box placeholder land in the wrong row once the collapsed row's position is removed from the list, misaligning the grid for that column.

2. **A collapsed column's own cell content can still influence column widths before it is zeroed.** `CollapseColumnWidths` zeroes a collapsed column's width only after `DetermineMissingColumnWidths`/`EnforceMaximumSize`/`EnforceMinimumSize` have all run — and those content-based width passes (`GetColumnsMinMaxWidthByContent`, `GetColumnMinWidths`) don't know a column is about to collapse, so a collapsed column's own (invisible) cell content can still push its own pre-collapse width up, and — through `EnforceMinimumSize`'s colspan-neighbor adjustment — narrow the width ultimately given to the *next* column. The end geometry is still valid (no zero-width/negative-width crashes), just not exactly what an author would get if the collapsed column's cells were excluded from width calculation from the start.

Both are recorded as an accepted gap in `.claude/accepted-gaps/visibility-collapse-table-layout-residual-gaps.md`. Fixing either means teaching `GetColumnsMinMaxWidthByContent`/`GetColumnMinWidths`/`InsertEmptyBoxes` about collapsed rows/columns directly, rather than filtering only at the row-loop/final-width level as #639 does.

PR #743, opened 2026-08-14
Make cell layout position and height rowspan-occupancy-aware, not just the border grid

### LINKED ISSUE #740: Rowspan-created gap in a detached <thead>/<tfoot> row also mis-positions the following cell's own layout
## Summary

`CssLayoutEngineTable.GetCellRealColumnIndex` sums the `colspan` of preceding cells within a row's own `Boxes` list to find where a cell starts, exactly like `TableGrid.Build` used to (fixed in #736). For an ordinary body row this is correct because `CssLayoutEngineTable.InsertEmptyBoxes` pads a rowspan's gap with a `CssSpacingBox` placeholder before layout runs. A detached `<thead>`/`<tfoot>` row-group's own rows are never passed through `InsertEmptyBoxes`, so the same gap that #736 fixed for collapsed-border *grid occupancy* still causes `GetCellRealColumnIndex` to return the wrong column for a cell following a rowspan gap in a header/footer row - and unlike `TableGrid.Build`, this function is also used directly by `CssLayoutEngineTable.LayoutBodyRow` (reused for header/footer row layout) to compute both the cell's rendered **width** (`GetCellWidth(columnIndex, cell)`) and, indirectly, its **X position** (`currentX` only advances past a rowspan gap when a `CssSpacingBox` placeholder occupies it in `row.Boxes` - which never happens for a header/footer row).

## Repro

```html
<table style="border-collapse:collapse;width:400px">
  <thead>
    <tr><th id="a" rowspan="2">A</th><th id="b">B</th></tr>
    <tr><th id="d">D</th></tr>
  </thead>
  <tbody>
    <tr><td id="x">x</td><td id="y">y</td></tr>
  </tbody>
</table>
```

Measured directly off the laid-out box tree: `A` and `x` (column 0) land at `X=20, Right=170.5`; `B` and `y` (column 1) land at `X=170.5, Right=320`. `D` - which should render under column 1, alongside `B`/`y` - instead lands at `X=20, Right=170.5`: the same position and width as column 0, overlapping/underneath `A` instead of `B`.

After #736's fix, the collapsed border for `D` now correctly resolves at column 1's boundary - so the border renders in the right place while `D`'s own text content renders in the wrong one, one column to the left of where its border shows up.

## Impact

A multi-row `<thead>`/`<tfoot>` with a `rowspan` cell (a common pattern for grouped column headers) renders every cell after the rowspan, in every row after the one it starts in, at the wrong horizontal position and width - not merely a border-resolution gap.

## Fix sketch

`TableGrid.Build`'s fix (#736) made its own column computation self-sufficient (rowspan-occupancy-aware) rather than trusting `GetCellRealColumnIndex`. That approach doesn't carry over cleanly here, since `LayoutBodyRow` needs an actual column index up front (not just grid topology) for every cell it lays out, header/footer included.

The more direct fix is the other option #736's own investigation already sketched: extend `CssLayoutEngineTable.InsertEmptyBoxes` (or an equivalent pass) to also pad a detached header's/footer's own rows with `CssSpacingBox` placeholders before layout runs, matching what already happens for `_bodyRows`. That would fix `GetCellRealColumnIndex` at its root for every consumer (this one included) in one place, rather than patching each call site.

Two things that fix needs to get right, discovered while scoping this out:
- `CssSpacingBox`'s `StartRow`/`EndRow` are read elsewhere in `LayoutBodyRow` (`sb.EndRow == rowIndex`) to close a rowspan cell and fold its bottom into row-height tracking. During header/footer measurement, `TableRowCursor.RowIndex` is pinned at `-1` for every row of the group (`ForRowGroupMeasurement`) rather than incremented per row the way body-row layout does - so naively reusing `GetEffectiveEndRowIndex`/the existing `EndRow`-comparison logic for a header/footer placeholder would silently never match, and the rowspan-closing logic that depends on it would need its own header/footer-aware row numbering first.
- Header/footer rows are filtered for `visibility:collapse` before layout sees them (`_headerBox.Boxes.Where(r => !IsRowCollapsed(r))`), so any padding pass needs to operate on that same filtered, already-local row list rather than raw markup order - mirroring how `_bodyRows`/`_bodyRowOriginalIndices` keep that distinction for body rows.

## Related

Found while verifying the fix for #736, which addressed the collapsed-border grid occupancy half of this same root gap (`TableGrid.Build`'s column computation) but left this layout-positioning half - a distinct code path using the same underlying `GetCellRealColumnIndex` - untouched.

### LINKED ISSUE #742: A rowspan cell in a detached <thead>/<tfoot> is never stretched to cover the rows it spans
## Summary

A `rowspan` cell inside a multi-row `<thead>`/`<tfoot>` never gets its height stretched to cover every row it spans - it keeps only its own natural (single-row) content height, even when the rows it spans are visibly taller.

Found while fixing #740 (a rowspan-created gap in a header/footer row mis-positioning the *following* cell): that fix corrects the following cell's X position and width, but the rowspan cell's own height is a separate, still-open defect through a different part of the layout pipeline.

## Repro

```html
<table style="border-collapse:collapse;width:400px">
  <thead>
    <tr><th id="a" rowspan="2">A</th><th id="b" style="height:60px">B</th></tr>
    <tr><th id="d" style="height:60px">D</th></tr>
  </thead>
  <tbody>
    <tr><td>x</td><td>y</td></tr>
  </tbody>
</table>
```

Measured directly off the laid-out box tree: `B` and `D` each get their declared `height:60px` (`Height=45pt`, `1px = 0.75pt`), so the header's two rows together are 90pt tall. `A`, spanning both rows, gets `Height=15` - its own single line of text, not stretched to the combined 90pt the two rows it spans actually take up.

## Root cause

`CssLayoutEngineTable.LayoutBodyRow`'s vertical-alignment pass only stretches a cell to `rowMaxBottom` when `GetRowSpan(cell) == 1`; a `rowSpan > 1` cell is instead supposed to be closed later, on the row its span ends on, via `CloseSpanningCell` - reached either through a `CssSpacingBox` placeholder (`sb.EndRow == rowIndex`) or through `TableRowCursor.RowSpannedBoxes` (keyed by the row the span ends on, `GetEffectiveEndRowIndex(rowIndex, rowSpan)`).

Both paths depend on `TableRowCursor.RowIndex` incrementing per row so that `rowIndex` matches the row a rowspan actually ends on. During a detached header's/footer's own measurement pass (`CssLayoutEngineTable.DetachAndMeasureRepeatedRowGroups`, via `TableRowCursor.ForRowGroupMeasurement`), `RowIndex` is pinned at `-1` for *every* row of the group instead - by design, since "its rows are not body rows, so neither its row numbering nor its rowspan bookkeeping... is theirs" (`TableRowCursor.ForRowGroupMeasurement`'s own doc comment). With `rowIndex` always `-1`, `GetEffectiveEndRowIndex(-1, rowSpan)` and the `RowSpannedBoxes`/`CloseSpanningCell` machinery that depends on it never engage correctly for a header/footer rowspan - so the cell is simply never closed, and keeps whatever height its own initial content layout gave it.

## Fix sketch

Give header/footer row measurement its own real, per-row-incrementing row index (distinct from `_bodyRows`' numbering, which `RowIndex`'s `-1` sentinel exists specifically to avoid colliding with), so `GetEffectiveEndRowIndex`/`RowSpannedBoxes`/`CloseSpanningCell` can engage the same way they already do for body rows. This is a deeper change than #740's own fix (which works entirely through `GetCellRealColumnIndex` and stays clear of `TableRowCursor`/`CloseSpanningCell`) since it touches the pagination-sensitive row-cursor design directly - worth its own focused change and test pass rather than folding into either #736 or #740.

jhaygood86:
## Summary

**#740** - a cell following a rowspan gap in a detached `<thead>`/`<tfoot>` row rendered at the wrong X position and width:
- `CssLayoutEngineTable.GetCellRealColumnIndex` had the same Boxes-list-summing bug #736 fixed in `TableGrid.Build`, but that fix didn't touch it - `TableGrid.Build` made its own column computation self-sufficient instead of routing through `GetCellRealColumnIndex`.
- `TableGrid`'s rowspan-occupancy placement algorithm is now shared (`TableGrid.ComputeColumnPlacements`) rather than private to `TableGrid.Build`. `CssLayoutEngineTable` computes it once per layout pass, caches it, and both `GetCellRealColumnIndex` (now a simple lookup) and `BuildTableGrid`'s collapsed-border grid reuse the same result rather than computing it twice.
- `LayoutBodyRow`'s column cursor now walks `currentX` forward past any gap a rowspan left un-padded, instead of advancing one `Boxes`-list entry at a time.
- `DetermineColumnCount` folds in an accurate floor from the same computation, closing a related out-of-bounds crash a still-naive column *count* could otherwise cause once column *index* became accurate.

**#742** - the rowspan cell's own *height* was never stretched to cover every row it spans in that same shape:
- `LayoutBodyRow`'s vertical-alignment loop only stretches a `rowSpan == 1` cell; a `rowSpan > 1` cell is instead closed later via `CloseSpanningCell`/`TableRowCursor.RowSpannedBoxes`, both keyed by `TableRowCursor.RowIndex` - pinned at `-1` for every row of a row-group measurement pass, so that machinery never engaged for a header/footer rowspan.
- `DetachAndMeasureRepeatedRowGroups`'s header/footer loops now keep this bookkeeping themselves, entirely independent of `TableRowCursor` (deliberately - `CloseSpanningCell`'s own bookkeeping is a pagination concept a row-group's one-shot measurement pass has no analogue for): as each row is placed, a `rowSpan > 1` cell is registered against the row its span ends on (using the same collapsed-row-aware remapping issue #665 already applies for body rows), that row's own height grows to fit the cell before being finalized, and the cell is then stretched to match.
- Also fixes a related, pre-existing crash in `TableGrid.Build`, which could record a cell's span reaching past its own grid's last row for a rowspan declared larger than its row-group's actual row count.

Fixes #740
Fixes #742

## Test plan

- [x] `TableLayout_RowspanInThead_PositionsLaterRowCellInItsOwnColumn` - #740's own repro shape.
- [x] `TableLayout_StaggeredRowspanChainInThead_DoesNotUndercountColumns` / `TableLayout_ColCountNarrowerThanStaggeredRowspanChain_DoesNotUndercountColumns` - the column-count crash fix, both `DetermineColumnCount` branches.
- [x] `TableLayout_RowspanInThead_CellStretchesToCoverEveryRowItSpans` - #742's own repro shape.
- [x] `TableLayout_RowspanCellTallerThanItsRows_GrowsTheHeaderToFitInstead` - the header/footer growth half (a spanning cell taller than the rows it spans).
- [x] `TableLayout_RowspanExceedingTheadsOwnRowCount_DoesNotThrow` - the `TableGrid.Build` crash fix.
- [x] `TableLayout_RowspanAcrossACollapsedTheadRow_ClosesOnTheCorrectRow` - the collapsed-row remapping fix.
- Every test above was verified against its actual pre-fix failure (wrong position/width, thrown exception, wrong height, or overlapping content), not merely by passing once the fix landed.
- [x] Full suite green: `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0` - 8833 passed, 9 skipped (platform-gated), 0 failed.
- [x] `dotnet build PeachPDF.slnx -t:Rebuild` - 0 warnings.
- [x] 100% diff coverage on the changed lines (`diff-cover` against `main`).


DaleCam:
Is this issue related to this - where the color bleeds above the border into the cell above it?
image

jhaygood86:
> Is this issue related to this - where the color bleeds above the border into the cell above it? image

Maybe, but I'd file a new issue to be safe. Please include the repro HTML, it helps a lot in these cases!

PR #927, opened 2026-09-08
Memoize IProperty.Value

dchupp:
`IPropertyValue.CssText` is not a stored string — nearly every converter **re-serializes** it on each read, several with LINQ and `string.Join` over their components (`BorderRadiusConverter`, `EndListValueConverter`, `BackgroundPositionValueConverter`, …).

A parsed declaration is shared by every box its rule matches, and `DomParser.AssignCssBlock` reads `Value` up to **three times per declaration per box** — the global-keyword `switch`, its fall-through arm, and `CssGlobalKeywords.TryParse`. So a single document-wide rule re-serializes its values thousands of times in one render.

The shape that makes it visible is an ordinary print stylesheet:

```css
* { box-shadow: none !important; border-radius: 0 !important; }
```

It matches every box, and `border-radius` re-serializes through four corner components each time.

## The change

`Value` caches its text, invalidated by `DeclaredValue`'s setter — the only way the underlying value changes; the value objects a converter builds are themselves immutable. `IsInitial` and `IsInherited` read through the same cache.

## Measured

Self-contained repro: a 400-row, 4-column bordered table (~1,600 boxes) with the universal print rule above, Letter, 0.5in margins. 3 warm-up renders, then 10 timed, `ServerGarbageCollection=true`. Three interleaved rounds of each build, rebuilt and hash-checked between arms:

| --- | --- | --- |
| time | 222.8 / 234.5 / 230.8 ms | **211.4 / 203.6 / 203.5 ms** |
| allocation | 203.0 / 203.0 / 202.9 MB | **179.9 / 179.7 / 179.8 MB** |
| median | 230.8 ms, 203.0 MB | **203.6 ms, 179.8 MB** |
| | | **−11.8% time, −11.4% allocation** |

Allocation is counted with `GC.GetTotalAllocatedBytes(true)` and is deterministic — the 23.2 MB/render is exact, not a sample.

<details>

```python
rows = 400
cells = "".join(
    f"<tr><td class='c'>Item {i}</td><td class='c'>SKU-{i:05d}</td>"
    f"<td class='c num'>{i*7%997}</td><td class='c num'>{i*13%89}.{i%100:02d}</td></tr>"
    for i in range(rows))
html = f"""<!DOCTYPE html><html><head><meta charset='utf-8'><style>
* {{ box-shadow: none !important; border-radius: 0 !important; }}
html {{ font-size: 16px; }}
body {{ font-family: Arial, sans-serif; font-size: 9pt; margin: 0; }}
table {{ width: 100%; border-collapse: collapse; }}
.c {{ border: 1px solid #999; padding: 2px 4px; }}
.num {{ text-align: right; }}
</style></head><body><table>{cells}</table></body></html>"""
open("bench.html","w").write(html)
```
</details>

## Tests

Found while profiling PeachPDF as the rendering engine behind a document-generation service; happy to adjust anything about the approach.

### DIFF
--- a/src/PeachPDF.Tests/CSS/Property.cs
+++ b/src/PeachPDF.Tests/CSS/Property.cs
@@ -1508,6 +1508,41 @@ namespace PeachPDF.Tests.CSS
             Assert.Equal("my-Property", property.Name);
             Assert.IsType<UnknownProperty>(property);
         }
+
+        /// <summary>
+        /// <see cref="Property.Value"/> memoizes the declaration's serialization, so reassigning the
+        /// declaration must invalidate it. Reading <c>Value</c> before each reassignment is the point
+        /// of the test — an unmemoized implementation passes it either way.
+        /// </summary>
+        [Fact]
+        public void CssPropertyValueFollowsAReassignedDeclaration()
+        {
+            var property = ParseDeclaration("break-after: avoid");
+            Assert.Equal("avoid", property.Value);
+
+            Assert.True(property.TrySetValue(ParseValue("page")));
+            Assert.Equal("page", property.Value);
+            Assert.Equal("break-after: page", property.CssText);
+
+            Assert.True(property.TrySetValue(ParseValue("column")));
+            Assert.Equal("column", property.Value);
+        }
+
+        /// <summary>
+        /// The no-declaration arm of <see cref="Property.Value"/> is memoized too, and the derived
+        /// <see cref="Property.IsInitial"/>/<see cref="Property.IsInherited"/> read through it.
+        /// </summary>
+        [Fact]
+        public void CssPropertyValueFollowsADeclarationClearedToInitial()
+        {
+            var property = ParseDeclaration("break-after: avoid");
+            Assert.Equal("avoid", property.Value);
+            Assert.False(property.IsInitial);
+
+            Assert.True(property.TrySetValue(null));
+            Assert.Equal(Keywords.Initial, property.Value);
+            Assert.True(property.IsInitial);
+        }
     }
 }
 
--- a/src/PeachPDF/CSS/StyleProperties/Property.cs
+++ b/src/PeachPDF/CSS/StyleProperties/Property.cs
@@ -33,16 +33,37 @@ namespace PeachPDF.CSS
             return true;
         }
 
-        public string Value => DeclaredValue != null ? DeclaredValue.CssText : Keywords.Initial;
+        /// <summary>
+        /// Memoized. <see cref="IPropertyValue.CssText"/> is not a stored string — nearly every
+        /// converter re-serializes it on each read, several with LINQ and string.Join over their
+        /// components (BorderRadiusConverter, EndListValueConverter, BackgroundPositionValueConverter,
+        /// …).
+        ///
+        /// A parsed declaration is shared by every box its rule matches, and the cascade reads this
+        /// up to three times per declaration per box (the global-keyword switch, its fall-through arm,
+        /// and CssGlobalKeywords.TryParse), so one document-wide rule re-serializes its values
+        /// thousands of times in a single render.
+        ///
+        /// Invalidated by <see cref="DeclaredValue"/>'s setter, which is the only way the underlying
+        /// value changes; the value objects a converter builds are themselves immutable.
+        ///
+        /// <c>null</c> is the "not computed yet" marker, the same way <see cref="DeclaredValue"/>'s
+        /// own <c>null</c> means "no declaration" — this file is <c>#nullable disable</c>, so the
+        /// field carries no annotation. A converter that returned a null <c>CssText</c> would only
+        /// re-serialize on the next read; it could never report a stale value.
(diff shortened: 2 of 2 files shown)


jhaygood86:
Reviewed against `CLAUDE.md` conventions. Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild`, and running the Property/CssUtils/Cascade-filtered test suite (2069 passed).

**1. This introduces a real build warning (blocking per `CLAUDE.md`)**

```
Property.cs(52,23): warning CS8632: The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```

`Property.cs` has `#nullable disable` at the top of the file, so `private string? _valueText;` isn't valid there. The existing `_declaredValue` field in this same class already sidesteps this (it uses `= null!;` rather than a `?` annotation) — could you follow that same convention here, e.g. `private string _valueText;` with a separate way to represent "not yet computed" (a sentinel, or a `_hasValueText` bool), rather than introducing a nullable-annotated field into a disabled-nullable file? `CLAUDE.md` requires a zero-warning solution build before merge.

**2. Correctness/thread-safety check (no issue found, sharing for the record)**

Given `CLAUDE.md`'s note that a parsed declaration is shared across boxes, I checked whether `Property` instances can also be shared *across concurrent renders* (which would make the new mutable `_valueText` field a race). They can be — `RAdapter.GetDefaultCssData()` caches the default/UA stylesheet's `Property` instances for the adapter's lifetime, and `CssData.Clone()` is shallow (shares the underlying rule objects) — but each `PdfGenerator` owns its own `PdfSharpAdapter`, so this is confined to a single `PdfGenerator`/thread under the existing "one `PdfGenerator` per thread, never reused concurrently" invariant. I also confirmed `DeclaredValue`'s setter is the only place that mutates it, and spot-checked the `IPropertyValue` implementations under `CSS/ValueConverters` for public mutable state (found none) — so the "value objects are immutable" assumption in your doc comment holds. No action needed here, just flagging that I checked it given how load-bearing that assumption is.

**3. Minor test suggestion (non-blocking)**

Behavior parity is already covered well by the existing suite (`Value`/`IsInitial`/`IsInherited` are read constantly). A small dedicated test that reassigns a declaration (parses a new value onto the same `Property`) and asserts `Value` reflects the new value — pinning the cache-invalidation contract your doc comment describes — would be a cheap addition.

The measurement methodology (deterministic `GC.GetTotalAllocatedBytes`, hash-checked builds, multiple interleaved rounds) is solid, and the memoization logic itself is correct once the nullable-context issue above is fixed.

dchupp:
Thanks — both addressed, force-pushed.

**1. CS8632.** Dropped the annotation: the field is now `private string _valueText;`. In a `#nullable disable` file `string` is oblivious, so `_valueText ??= …` compiles clean and `null` is representable as the "not computed yet" marker — the same way `_declaredValue`'s own `null` means "no declaration". I went with that rather than a separate `_hasValueText` bool because there is nothing for the extra field to buy: if some converter ever returned a null `CssText`, the only consequence is that the next read re-serializes it. A stale value is impossible either way. Happy to add the bool if you'd rather the intent be explicit rather than inferred from the file's nullable context; I've noted the reasoning in the doc comment so the next reader doesn't have to reconstruct it.

`dotnet build src/PeachPDF/PeachPDF.csproj -t:Rebuild`: 0 warnings, 0 errors.

**2. The thread-safety trace.** Thank you for doing that — the `RAdapter.GetDefaultCssData()` caching and the shallow `CssData.Clone()` are both things I had not checked, and "confined to one `PdfGenerator`" is a materially stronger statement than the one I was relying on. Noted.

**3. The cache-invalidation test.** Added two, in `CSS/Property.cs`:

- `CssPropertyValueFollowsAReassignedDeclaration` — reads `Value` (memoizing it), reassigns via `TrySetValue`, asserts `Value` and `CssText` both follow, twice.
- `CssPropertyValueFollowsADeclarationClearedToInitial` — the no-declaration arm, plus `IsInitial` reading through it.

Both fail if `DeclaredValue`'s setter stops clearing the cache: commenting out the `_valueText = null;` line gives 2 failed / 0 passed. Reading `Value` *before* each reassignment is the point — without that an unmemoized implementation passes either way, which I noted in the test's doc comment.

Full suite on this branch: 10,200 passed, 0 failed, 9 skipped. (Single `net9.0` target — that's the SDK I have locally; upstream multi-targets `net8.0;net10.0`, so absolute totals differ from CI's.)

jhaygood86 (approved):
Verified the update: `_valueText` is now `string` (not `string?`), which fixes the CS8632 warning I found (confirmed with a clean `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings). Two solid tests added (`CssPropertyValueFollowsAReassignedDeclaration`, `CssPropertyValueFollowsADeclarationClearedToInitial`) covering both the reassignment-invalidates-cache and cleared-to-initial paths — ran them plus the full Property/CssUtils/Cascade filter, all pass (2071 tests).

PR #928, opened 2026-09-08
Do not copy the sibling list to walk a sibling combinator

dchupp:
The `+` and `~` arms of `DoesSelectorMatch(ComplexSelector, …)` each opened with:

```csharp
var siblings = parent.Children.Where(b => b.TagName is not null).ToList();
var idx = siblings.IndexOf(currentRef);
```

A filtered **copy** of every sibling, then a linear scan of that copy — per combinator, per candidate box.

It matters because of *which* rules reach it. A selector is indexed by its **subject** compound, and Tailwind's spacing utilities are `.space-y-N > * + *` — the subject is `*`, so they land in the universal bucket and are tested against **every box in the document**. A container with many children then pays an O(n) copy per box per rule.

## The change

Index arithmetic over the live `IReadOnlyList<ICssDomNode>`. Evaluation order is unchanged — still the nearest preceding *element* sibling, still stopping at the first hit — and the two cases the old `idx <= 0` folded together (the node is not an element; the node is absent from the list) are kept explicitly.

**Comparison stays the list's own `Equals`, not `ReferenceEquals`.** An SVG DOM node (`SvgCssBoxDomNode`, `SvgXmlDomNode`) is a fresh wrapper object per access that overrides `Equals` over the box it wraps, so reference identity would silently stop matching sibling combinators inside inline SVG. I verified that directly: switching to `ReferenceEquals` makes `rect + rect` and `circle ~ circle` match nothing inside an inline `<svg>` while the whole test suite stays green.

## Measured

Self-contained repro: N `<div class="row">` children inside one `.space-y-2`, plus the four `.space-y-*` rules above, Letter, 0.5in margins. 3 warm-ups then 10 timed, `ServerGarbageCollection=true`, three interleaved rounds per arm with the built assembly hash-checked to differ between arms.

**1,200 rows:**

| --- | --- | --- |
| time | 375.5 / 379.9 / 377.3 ms | **300.3 / 300.9 / 297.4 ms** |
| allocation | 368.1 / 368.3 / 368.1 MB | **319.3 / 319.3 / 319.3 MB** |
| median | 377.3 ms, 368.1 MB | **300.3 ms, 319.3 MB** |
| | | **−20.4% time, −13.3% allocation** |

**300 rows** — the allocation win is still exact (84.1 → 80.1 MB, −4.8%) but the time difference is inside the noise, which is what an O(n) copy should look like as n falls.

Allocation is `GC.GetTotalAllocatedBytes(true)` and is deterministic.

<details>

```python
rows = 1200
items = "".join(f"<div class='row'><span>Item {i}</span><span>SKU-{i:05d}</span><span>{i*7%997}</span></div>" for i in range(rows))
html = f"""<!DOCTYPE html><html><head><meta charset='utf-8'><style>
.space-y-1 > * + * {{ margin-top: 0.25rem; }}
.space-y-2 > * + * {{ margin-top: 0.5rem; }}
.space-y-3 > * + * {{ margin-top: 0.75rem; }}
.space-y-4 > * + * {{ margin-top: 1rem; }}
html {{ font-size: 16px; }}
body {{ font-family: Arial, sans-serif; font-size: 9pt; margin: 0; }}
.row {{ border-bottom: 1px solid #ccc; padding: 2px 0; }}
.row span {{ display: inline-block; width: 30%; }}
</style></head><body><div class="space-y-2">{items}</div></body></html>"""
open("siblings.html","w").write(html)
```
</details>

## Tests

Worth noting the suite does **not** cover the SVG equality case above; it stays green with the comparison broken. If you'd like, I can add a fixture for it.

### DIFF
--- a/src/PeachPDF/Html/Core/CssData.cs
+++ b/src/PeachPDF/Html/Core/CssData.cs
@@ -1415,6 +1415,49 @@ namespace PeachPDF.Html.Core
             return false;
         }
 
+        /// <summary>
+        /// The nearest preceding sibling of <paramref name="node"/> that is an element
+        /// (<c>TagName is not null</c>), or null when there is none — including when
+        /// <paramref name="node"/> is not itself an element, or is absent from
+        /// <paramref name="children"/>. Both of those are the cases the
+        /// <c>Where(...).ToList().IndexOf(...)</c> this replaces folded into <c>idx &lt;= 0</c>.
+        ///
+        /// Comparison is the list's own equality, NOT <c>ReferenceEquals</c>: an SVG DOM node
+        /// (<c>SvgCssBoxDomNode</c>, <c>SvgXmlDomNode</c>) is a fresh wrapper object per access
+        /// that defines value equality over the box it wraps, so reference identity would silently
+        /// stop matching sibling combinators inside inline SVG.
+        /// </summary>
+        private static ICssDomNode? PrecedingElementSibling(IReadOnlyList<ICssDomNode> children, ICssDomNode node)
+        {
+            if (node.TagName is null) return null;
+
+            ICssDomNode? previous = null;
+            for (var i = 0; i < children.Count; i++)
+            {
+                var sibling = children[i];
+                if (sibling.Equals(node)) return previous;
+                if (sibling.TagName is not null) previous = sibling;
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        /// <paramref name="node"/>'s position in <paramref name="children"/>, or -1 when it is not
+        /// an element or is absent. See <see cref="PrecedingElementSibling"/> on the equality.
+        /// </summary>
+        private static int ChildIndexOf(IReadOnlyList<ICssDomNode> children, ICssDomNode node)
+        {
+            if (node.TagName is null) return -1;
+
+            for (var i = 0; i < children.Count; i++)
+            {
+                if (children[i].Equals(node)) return i;
+            }
+
+            return -1;
+        }
+
         private static bool DoesSelectorMatch(ComplexSelector complexSelector, ICssDomNode? node)
         {
             var selectorsInReverse = complexSelector.Reverse().ToList();
@@ -1474,10 +1517,10 @@ namespace PeachPDF.Html.Core
                     {
                         var parent = currentRef.Parent;
                         if (parent is null) return false;
-                        var siblings = parent.Children.Where(b => b.TagName is not null).ToList();
-                        var idx = siblings.IndexOf(currentRef);
-                        if (idx <= 0) return false;
-                        currentRef = siblings[idx - 1];
+
+                        var previous = PrecedingElementSibling(parent.Children, currentRef);
+                        if (previous is null) return false;
+                        currentRef = previous;
                         if (!DoesSelectorMatch(selector.Selector, currentRef)) return false;
                         break;
                     }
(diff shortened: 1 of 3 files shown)


src/PeachPDF/Html/Core/CssData.cs
```diff
@@ -1415,6 +1415,49 @@ private static bool HasDescendantMatch(ICssDomNode node, ISelector inner)
...
+        /// <paramref name="node"/> is not itself an element, or is absent from
+        /// <paramref name="children"/>. Both of those are the cases the
+        /// <c>Where(...).ToList().IndexOf(...)</c> this replaces folded into <c>idx &lt;= 0</c>.
+        ///
+        /// Comparison is the list's own equality, NOT <c>ReferenceEquals</c>: an SVG DOM node
+        /// (<c>SvgCssBoxDomNode</c>, <c>SvgXmlDomNode</c>) is a fresh wrapper object per access
+        /// that defines value equality over the box it wraps, so reference identity would silently
+        /// stop matching sibling combinators inside inline SVG.
+        /// </summary>
+        private static ICssDomNode? PrecedingElementSibling(IReadOnlyList<ICssDomNode> children, ICssDomNode node)
```
jhaygood86: This needs tests.

jhaygood86:
Reviewed against `CLAUDE.md` conventions. Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` (clean, 0 warnings) and the selector/CssData/SVG-filtered test suite (944 passed).

**Missing regression test for the actual correctness-relevant part of this change (please add before merge)**

The perf win here is a straightforward index-arithmetic replacement (double-checked the two new helpers, `PrecedingElementSibling`/`ChildIndexOf`, against the original `Where(...).ToList().IndexOf(...)` semantics for the absent-node and non-element-`currentRef` cases — they match). But the change you flagged as the important part — keeping `sibling.Equals(node)` instead of switching to `ReferenceEquals` — is exactly the kind of thing this repo's own conventions call out: `CLAUDE.md`'s testing section says a change should have "an integration test that would fail if the feature were a complete no-op," and the invariants folder exists specifically to record "a rule the next change could plausibly violate." You've already done the hard part (you manually verified `ReferenceEquals` breaks `rect + rect` / `circle ~ circle` inside inline SVG and the full suite stays green regardless) — that repro is worth keeping as an actual test, not just a PR-description note, since a future "obvious" simplification (e.g. someone hoists `sibling == node` or swaps in a `HashSet`/dictionary-based lookup keyed on reference identity for a further perf pass) would silently reintroduce this with nothing to catch it.

Concretely, could you:
1. Add an SVG test with `rect + rect` / `circle ~ circle` sibling-combinator rules and assert they match (in `PeachPDF.Tests`), and
2. Add a short note under `.claude/invariants/` (e.g. `css-selectors-sibling-combinator-equals-not-referenceequals.md`) recording that `SvgCssBoxDomNode`/`SvgXmlDomNode` are fresh per-access wrapper objects, so sibling-combinator comparisons must use value equality?

No other issues found — the O(n) copy removal is real and the measurement methodology (hash-checked builds, interleaved rounds, deterministic allocation count) matches what this repo already does elsewhere.

dchupp:
Both done, force-pushed. You're right that the repro belonged in the suite rather than in a PR description — that is exactly the kind of thing a later perf pass would quietly undo.

**1. The SVG sibling-combinator tests** are in `Svg/SvgCssDomNodeTests.cs`, next to the existing `MatchedDeclarations_*` fixtures:

- `MatchedDeclarations_AdjacentSiblingCombinator_MatchesTheFollowingElement` — a `[Theory]` over `rect + rect`, `circle + circle` and `rect + circle`, asserting the full match set across five sibling elements rather than one positive case, so a combinator that matched too *widely* fails too.
- `MatchedDeclarations_GeneralSiblingCombinator_ReachesEveryFollowingSibling` — `circle ~ circle` with a non-matching `<rect>` in between, so the "skip intervening elements" half of the `~` arm is covered as well as the reach.

Swapping both helpers to `ReferenceEquals` fails 4 of the 8 theory cases; the rest of the suite stays green, which is the point.

**2. The invariant note** is at `.claude/invariants/css-selectors-sibling-combinator-equals-not-referenceequals.md`. It records that `SvgXmlDomNode`/`SvgCssBoxDomNode` allocate a fresh wrapper on every `Parent`/`Children` read and override `Equals`/`GetHashCode` against the wrapped `XElement`/`CssBox`, so two wrappers for the same element are `Equals` but never `ReferenceEquals` — and names the specific shapes that would reintroduce the bug (a direct `ReferenceEquals`, a `HashSet`/dictionary keyed on reference identity, `(object)a == (object)b`), plus why the HTML side would stay green through it.

Full suite on this branch: 10,202 passed, 0 failed, 9 skipped. (Single `net9.0` target locally; upstream multi-targets `net8.0;net10.0`, so absolute totals differ from CI's.)

jhaygood86 (approved):
Verified the update: real regression tests added (`MatchedDeclarations_AdjacentSiblingCombinator_*`, `MatchedDeclarations_GeneralSiblingCombinator_*`) plus the invariants file I suggested. I mutation-tested this myself — temporarily swapped `sibling.Equals(node)`/`children[i].Equals(node)` back to `ReferenceEquals` and confirmed these new tests fail (4 failures) exactly as claimed, then reverted. Clean rebuild (0 warnings), and the full Svg/Selector-filtered suite passes (309 tests).

PR #929, opened 2026-09-08
Ask the cheap question first in the typed fast path

dchupp:
`AssignCssBlock`'s typed fast path is guarded by three tests, in the most expensive order:

```csharp
if (!CssGlobalKeywords.TryParse(prop.Value, out _)
    && prop is Property typedProp
    && CssUtils.TrySetTypedPropertyValue(box, prop.Name, typedProp.DeclaredValue))
```

`_typedPropertySetters` holds **two** entries — `grid-template-columns` and `grid-template-rows` — so on any document that is not a grid the path cannot fire. Yet every declaration of every box ran `CssGlobalKeywords.TryParse` first (five `OrdinalIgnoreCase` comparisons) to answer a question the two-entry lookup settles outright.

## The change

Check the name first, via a new `CssUtils.HasTypedPropertySetter`. All three tests are pure and the only side-effecting operand (the setter) is still last, so the reorder cannot change a result.

## Measured

The interesting part is that it saves **allocation**, not just comparisons: `prop.Value` reads `IPropertyValue.CssText`, which most converters re-serialize, so skipping it for the ~99.99% of declarations that can never take this path avoids the string entirely.

400-row bordered table with a `* { box-shadow: none !important; border-radius: 0 !important }` print rule, Letter, 0.5in margins. 3 warm-ups then 10 timed, `ServerGarbageCollection=true`, three interleaved rounds per arm with the built assembly hash-checked to differ:

| --- | --- | --- |
| allocation | 202.8 / 202.9 / 202.9 MB | **197.1 / 197.3 / 197.2 MB** |
| median | 202.9 MB | **197.2 MB** — **−2.8%** |

Allocation is `GC.GetTotalAllocatedBytes(true)` and is deterministic.

**Time is inside the noise on this document** (medians 235.6 ms vs 229.2 ms, ranges overlapping) and I am not claiming a time win. The removed work is real and counted; that is the case for it.

## Relationship to #927

Partially overlapping: #927 memoizes `IProperty.Value`, which makes the `prop.Value` read cheap rather than avoiding it. With both, this reorder still skips the five keyword comparisons but saves less allocation. They are independent commits and either can land alone — happy to rebase whichever comes second.

## Tests

### DIFF
--- a/src/PeachPDF/Html/Core/Utils/CssUtils.cs
+++ b/src/PeachPDF/Html/Core/Utils/CssUtils.cs
@@ -93,6 +93,15 @@ namespace PeachPDF.Html.Core.Utils
         }
 
         /// <summary>
+        /// Whether <paramref name="propName"/> has a typed fast-path setter at all. The cheap question —
+        /// a lookup in the same two-entry table <see cref="TrySetTypedPropertyValue"/> uses — so a caller
+        /// can ask it before doing anything more expensive on a property that has no fast path.
+        /// </summary>
+        /// <param name="propName">the name of the CSS property</param>
+        /// <returns>true when a typed setter is registered for <paramref name="propName"/></returns>
+        public static bool HasTypedPropertySetter(string propName) => _typedPropertySetters.ContainsKey(propName);
+
+        /// <summary>
         /// Assigns a property's already-parsed, strongly-typed value straight onto a <see cref="CssBox"/> from
         /// its Layer A <see cref="ITypedPropertyValue{T}"/> carrier, without re-parsing the authored string.
         /// Returns false when <paramref name="propName"/> has no typed setter, or when
--- a/src/PeachPDF.Tests/Html/Core/Utils/CssUtilsTests.cs
+++ b/src/PeachPDF.Tests/Html/Core/Utils/CssUtilsTests.cs
@@ -1177,6 +1177,36 @@ namespace PeachPDF.Tests.Html.Core.Utils
 div {{ width: 200px; height: 100px; {css} }}
 </style></head><body><div>Text</div></body></html>";
 
+        /// <summary>
+        /// <see cref="CssUtils.HasTypedPropertySetter"/> answers exactly the registered set — it is read
+        /// as the cheap first operand of the cascade's typed fast path, so a name it wrongly answers true
+        /// for costs the work the guard exists to skip, and one it wrongly answers false for silently
+        /// drops the fast path back to the string setter.
+        /// </summary>
+        [Theory]
+        [InlineData("grid-template-columns", true)]
+        [InlineData("grid-template-rows", true)]
+        [InlineData("grid-template-areas", false)]
+        [InlineData("display", false)]
+        [InlineData("color", false)]
+        [InlineData("--custom", false)]
+        [InlineData("", false)]
+        public void HasTypedPropertySetter_MatchesTheRegisteredSet(string propName, bool expected)
+        {
+            Assert.Equal(expected, CssUtils.HasTypedPropertySetter(propName));
+        }
+
+        /// <summary>
+        /// The lookup is ordinal, matching the table <see cref="CssUtils.TrySetTypedPropertyValue"/>
+        /// itself uses — so the two never disagree about whether a name has a fast path. Property names
+        /// reach this already lower-cased by the parser.
+        /// </summary>
+        [Fact]
+        public void HasTypedPropertySetter_IsOrdinal_LikeTheTableItReads()
+        {
+            Assert.False(CssUtils.HasTypedPropertySetter("GRID-TEMPLATE-COLUMNS"));
+        }
+
         private static async Task<CssBox> FindDivBoxFromHtml(string html)
         {
             var adapter = new PdfSharpAdapter();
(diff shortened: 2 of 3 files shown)


src/PeachPDF/Html/Core/Utils/CssUtils.cs
```diff
@@ -103,6 +103,10 @@ public static void SetPropertyValue(CssValueParser valueParser, CssBox cssBox, s
         /// Not yet subsumed by <see cref="CssPropertyRegistry"/>: the generator doesn't emit a typed fast path
         /// yet (only grid-template-columns/-rows have one, and only here) — see CLAUDE.md's generator section.
         /// </remarks>
+        /// <summary>Whether <paramref name="propName"/> has a typed fast-path setter at all — the
+        /// cheap question, so a caller can ask it before doing anything expensive.</summary>
+        public static bool HasTypedPropertySetter(string propName) => _typedPropertySetters.ContainsKey(propName);
```
jhaygood86: can you add tests for this?

jhaygood86:
Reviewed against `CLAUDE.md` conventions. Verified by checking out this branch and running `dotnet build PeachPDF.slnx -t:Rebuild`.

**1. This introduces a real build warning (blocking per `CLAUDE.md`)**

```
CssUtils.cs(99,29): warning CS1734: XML comment on 'CssUtils.HasTypedPropertySetter(string)' has a paramref tag for 'declaredValue', but there is no parameter by that name
```

Same root cause as the doc-comment issue I flagged on #930: the new `HasTypedPropertySetter` method + its one-line `<summary>` were inserted *before* `TrySetTypedPropertyValue`, right after that method's existing `<remarks>` block — which detaches `TrySetTypedPropertyValue`'s real doc comment (summary + `<remarks>`, written for a method that has a `declaredValue` parameter) and reattaches it to `HasTypedPropertySetter`, which has no such parameter. Net effect: `HasTypedPropertySetter` now carries a wrong, stale doc comment on top of its own, and `TrySetTypedPropertyValue` has none. `CLAUDE.md` requires the whole solution to build with zero warnings before merge — could you give `HasTypedPropertySetter` its own complete, self-contained doc comment placed so it doesn't swallow the neighboring one?

**2. Small test suggestion (non-blocking)**

The reorder itself is pure/behavior-preserving and already exercised indirectly (this code path runs for every property on every box), so I don't think it's missing diff coverage in the CI sense. Still, a direct unit test for `CssUtils.HasTypedPropertySetter` (true for `grid-template-columns`/`grid-template-rows`, false for an arbitrary other property name) would pin the contract cheaply and make the intent of the new public surface explicit.

The logic itself checks out: all three operands (`HasTypedPropertySetter`, `CssGlobalKeywords.TryParse`, the `prop is Property` pattern) are pure, so reordering them ahead of the side-effecting `TrySetTypedPropertyValue` call cannot change behavior — just avoids the work for non-grid properties.

dchupp:
Both addressed, force-pushed.

**1. CS1734.** `HasTypedPropertySetter` now sits *above* `TrySetTypedPropertyValue`'s doc comment instead of between that comment and its method, so the `<summary>`/`<remarks>` pair goes back to the method it was written for. `HasTypedPropertySetter` has its own complete comment (`<summary>` + `<param>` + `<returns>`).

`dotnet build src/PeachPDF/PeachPDF.csproj -t:Rebuild`: 0 warnings, 0 errors.

**2. The unit test.** In `Html/Core/Utils/CssUtilsTests.cs`:

- `HasTypedPropertySetter_MatchesTheRegisteredSet` — a `[Theory]`: true for `grid-template-columns`/`grid-template-rows`, false for `grid-template-areas` (the near-miss most likely to be assumed in), `display`, `color`, a custom property, and the empty string.
- `HasTypedPropertySetter_IsOrdinal_LikeTheTableItReads` — pins that the lookup is ordinal, matching the `StringComparer.Ordinal` table `TrySetTypedPropertyValue` itself uses, so the two can never disagree about whether a name has a fast path.

**On the doc-comment pattern across #929/#930/#933** — you're right that it was systematic rather than three coincidences, and it came from how I was inserting these: placing a new member immediately after an existing member's declaration line while its doc comment sat above, which silently reparents the comment. I've since run a check across all fourteen branches for the shape (one contiguous `///` run carrying more than one `<summary>`) and fixed every instance that is ours. Four hits remain but pre-date these branches — they're on `main` today, at `CssBox.cs`'s `LayoutOutOfFlowChildren`, `KeepsFewerLinesThanOrphans` and `PhysicalMargin`, and `HtmlContainerInt.PerformLayout`. Happy to send a separate tidy-up PR for those if you want them cleaned; I left them alone rather than mixing unrelated churn into these.

Full suite on this branch: 10,206 passed, 0 failed, 9 skipped. (Single `net9.0` target locally; upstream multi-targets `net8.0;net10.0`, so absolute totals differ from CI's.)

jhaygood86 (approved):
Verified the update: `HasTypedPropertySetter` now has its own complete, correctly-placed doc comment (no longer swallowing `TrySetTypedPropertyValue`'s), confirmed by a clean `dotnet build PeachPDF.slnx -t:Rebuild` — the CS1734 warning is gone (0 warnings). Tests added (`HasTypedPropertySetter_MatchesTheRegisteredSet`, `HasTypedPropertySetter_IsOrdinal_LikeTheTableItReads`) directly pin the new method's contract. Ran the CssUtils/Grid-filtered suite, all pass (700 tests).

PR #930, opened 2026-09-08
A float does not narrow a line outside its own formatting context

dchupp:
`FindNarrowestRightFloatBox` climbs from the box being laid out to the document root, scanning every preceding sibling's subtree on the way. Nothing stops it at a formatting-context boundary — so **a float in one grid item sets the wrap limit for lines in the next one**.

CSS 2.1 §9.5 and css-display-3 §2.1: a float cannot affect content outside its own formatting context.

## Repro

A two-column grid. First item has a `float: right`; second item has plain text. Rendered twice, once with the float and once without, and the word positions compared (`pdftotext -bbox`):

| | `SHIPPING` in item 2 |
| --- | --- |
| `float: right` present | y = **51.75** |
| `float: right` removed | y = 39.75 |

The second item's text is pushed down 12 pt purely by its neighbour's float. Nothing in item 2 references item 1.

```html
<style>
body { font-family: Arial, sans-serif; font-size: 10pt; margin: 0; }
.grid { display: grid; grid-template-columns: 300px 300px; width: 600px; }
.cell { border: 1px solid #999; padding: 4px; }
.badge { float: right; width: 200px; height: 60px; background: #ddd; }
</style>
<div class="grid">
  <div class="cell"><span class="badge">12345</span><b>BILL TO</b></div>
  <div class="cell">SHIPPING INFORMATION FOR THIS ORDER</div>
</div>
```

## The change

Break the ancestor climb at a box that establishes an independent formatting context. `EstablishesIndependentFormattingContext` covers the cases that matter for this walk: a grid or flex **item** (an item does so whatever its own display), a table cell, an inline-block, a float, an out-of-flow box, and anything with a non-visible `overflow`.

## Verified

With the fix, on the same repro:

| | before | after |
| --- | --- | --- |
| `SHIPPING` moves when the neighbour floats | **yes** (39.75 → 51.75) | **no** (39.75 both ways) |
| the float still affects its own item | yes | yes — unchanged |

The `float: left` variant of the same document is unaffected in both builds, and the float continues to displace content inside its own item.

## Tests

`PeachPDF.Tests` on this branch: **10,207 passed, 0 failed, 9 skipped** — same as `main` (`cd28a35`). The suite does not currently cover this case; happy to add a fixture if you'd like one.

## Note on the left-side walk

`FindIntersectingFloatBox` (the left-side point-collision walk) has the same unbounded climb. I could not construct a document where it produces a *visible* divergence on `main`, so this PR leaves it alone rather than changing behaviour I cannot demonstrate. It is also the more expensive of the two — on a large document it dominates the candidate-gathering cost — so it may be worth a separate look. One subtlety if you do: the break must **not** apply at the walk's starting level, because a float is itself a formatting-context root for its own contents and must still be positioned against its siblings.

### DIFF
--- a/src/PeachPDF.Tests/Integration/FloatLayoutRegressionTests.cs
+++ b/src/PeachPDF.Tests/Integration/FloatLayoutRegressionTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Text;
 using PeachPDF;
 using PeachPDF.Adapters;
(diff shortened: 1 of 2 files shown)


jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (layout-engine testing, doc-comment hygiene, zero-warning builds). Verified by checking out this branch and running `dotnet build PeachPDF.slnx -t:Rebuild` and the `Float`-filtered test suite.

`CLAUDE.md`'s testing conventions require layout-engine changes to have unit tests asserting the affected `CssBox`'s properties after layout, using the `HtmlContainerInt` + `PdfSharpAdapter` harness pattern (see `FlexboxIntegrationTests.cs`/`MulticolLayoutIntegrationTests.cs`), and the 90% diff-coverage gate applies to new/changed code. `EstablishesIndependentFormattingContext` has several branches (grid/flex item, table-cell, inline-block, float, out-of-flow, non-visible overflow) none of which are exercised by an automated test — the PR description itself notes "the suite does not currently cover this case." Could you add a fixture based on your repro (two-column grid, `float: right` in the first item, assert the second item's text `Location`/line-box position is unaffected)?

**2. Doc comment got detached from `FindNarrowestRightFloatBox` (build hygiene)**

The new `EstablishesIndependentFormattingContext` method + its doc comment were inserted *between* the pre-existing XML doc comment ("Same ancestor/preceding-sibling traversal shape as ... accumulates the narrowest match ...") and the method it documents, `FindNarrowestRightFloatBox`. That comment is now attached to `EstablishesIndependentFormattingContext` instead (which ends up with two stacked `<summary>` blocks, one describing an unrelated method), and `FindNarrowestRightFloatBox` is left with no doc comment at all. This didn't trip a compiler warning here, but it's misleading — could you move the new method (with its own doc) above that existing comment, or move the existing comment back down onto `FindNarrowestRightFloatBox`?

**3. Worth a one-line comment on the starting-level break**

The PR body's "Note on the left-side walk" flags that breaking at the walk's *starting* level would be wrong for `FindIntersectingFloatBox`, because that function is called with a float itself as `reference` (from `FloatBoxLeft`/`FloatBoxRight`) and a float must still be positioned against its siblings. `FindNarrowestRightFloatBox`'s `reference` (from `FlowBox`'s word-flow loop) is never itself the box establishing the FC boundary in that sense, so breaking at the starting level here is correct — but since this is exactly the kind of thing the sibling function got wrong in the other direction, it'd help future maintainers to spell out why the two functions differ here, right next to the `break`.

Nice, well-isolated fix otherwise — the formatting-context boundary list (grid/flex item, table-cell, inline-block, float, out-of-flow, non-visible overflow) matches CSS 2.1 §9.4.1 / css-flexbox-1 §4 / css-grid-1 §6, and the build is clean (0 warnings) with all existing float tests passing.

dchupp:
All three addressed, force-pushed.

**1. Layout test.** Two fixtures in `Integration/FloatLayoutRegressionTests.cs`, using that file's existing `BuildAndLayout` harness:

- `FloatRight_InOneGridItem_DoesNotNarrowLinesInTheNext` — the repro: a two-column grid, `float: right` in the first item, and the second item's text measured against the *same document without the float*. Asserting equality against the no-float control rather than a hardcoded coordinate means the test says "the float is invisible here", which is the actual claim.
- `FloatRight_InTheSameFormattingContext_StillNarrowsTheLine` — the contrast case. Without it the first test also passes if float avoidance stopped working altogether.

Reverting `DomUtils.cs` to `main` fails the first and passes the second, so the pair brackets the behaviour rather than just pinning it.

**2. Doc comment.** Moved `EstablishesIndependentFormattingContext` above `GetLastRightIntersectingFloatBox`'s doc comment, so the "Same ancestor/preceding-sibling traversal shape as…" comment goes back onto `FindNarrowestRightFloatBox`. (You flagged this on #929 and #933 too — I've now swept all fourteen branches for the shape and fixed every instance that's ours; details in my reply on #929.)

**3. The starting-level comment.** Added next to the `break`, and I've written it as the general rule rather than as a note about this function, since that's what makes it usable at the other call site:

> Tested at the STARTING level too, unlike `FindIntersectingFloatBox`, and the difference is in what each is handed. This walk's `reference` comes from `FlowBox`'s word-flow loop: it is the block whose line is being laid out, so if IT establishes a formatting context, the walk is already outside and has nothing to find. `FindIntersectingFloatBox` is called from `FloatBoxLeft`/`FloatBoxRight` with a float itself as `reference`, and a float always establishes one — breaking at its starting level would stop it being positioned against its own siblings, which is the whole job. Same rule, different starting box.

Full suite on this branch: 10,200 passed, 0 failed, 9 skipped; 0 build warnings. (Single `net9.0` target locally; upstream multi-targets `net8.0;net10.0`, so absolute totals differ from CI's.)

jhaygood86 (approved):
Verified the update: two tests added (`FloatRight_InOneGridItem_DoesNotNarrowLinesInTheNext` for the fix, `FloatRight_InTheSameFormattingContext_StillNarrowsTheLine` as the contrast/non-regression case) — a good pairing that pins the formatting-context boundary specifically, not just "floats still work." The doc comment is now correctly attached to `EstablishesIndependentFormattingContext` itself, and a new inline comment explains exactly the starting-level-break asymmetry I asked about (why this walk breaks at its own starting box but `FindIntersectingFloatBox` cannot). Clean rebuild (0 warnings), float-filtered suite passes (89 tests).

PR #932, opened 2026-09-08
rem resolves against the root element, not the engine default

dchupp:
css-values-3 §5.1.2 defines `rem` as the font size of the **root element**. `DerivedStyle.GetRemHeight` walks to the topmost `CssBox` instead — which is the container's own root box. It sits above `<html>`, carries no element, and so always reports `DefaultFontResolver.FontSize`. Every `rem` in the document therefore resolves against that default, whatever the root declares.

## Repro

```html
html { font-size: 32px; }
body { margin: 0; font-family: Arial, sans-serif; }
.px  { font-size: 32px; }
.rem { font-size: 1rem; }
.em  { font-size: 1em; }
</style></head><body>
<div class="px">PX</div>
<div class="rem">REM</div>
<div class="em">EM</div>
</body></html>
```

All three must render at the same size. Glyph heights via `pdftotext -bbox`:

| --- | --- | --- |
| `font-size: 32px` | 26.81 | 26.81 |
| `font-size: 1em` | 26.81 | 26.81 |
| **`font-size: 1rem`** | **12.29** | **26.81** |

`1em` and `1rem` disagree in the same document, and `1rem` lands on the 11 pt default rather than the declared 32 px.

## Why it is worth fixing

It presents as a **scale error**, not a unit error. A stylesheet that sizes everything in `rem` — which is most of them, and every Tailwind-style one — gets a single wrong multiplier applied uniformly, so the whole document renders at the wrong size and paginates accordingly. Nothing looks broken; it is just silently the wrong size.

## The change

The walk now remembers the outermost box that **has** an element. Falling back to the topmost box keeps two cases right:

- a fragment with no element boxes at all;
- the root element resolving its **own** font size, where a `rem` resolves against the initial value (css-values-3 §5.1.2) and consulting the root again would recurse through `ActualFont`.

## Out of scope: the same bug in `@page` context

`DomParser.ApplyPageStylesOnce` resolves `@page`'s `em`/`rem` bases off the **synthetic wrapper** box rather than `<html>`, so an `@page { margin-top: 1rem }` still lands on the default however the document declares `html { font-size }`. Fixing `GetRemHeight()`'s walk cannot reach it — the wrapper has nothing above it to walk to — and the obvious lookup (`DomUtils.GetBoxByTagName(root, "html")`) is actively harmful there, because `ApplyPageStylesOnce` runs before `CascadeApplyStyles` and `ActualFont` memoizes on first read, so it caches the default font on `<html>` permanently and breaks `rem` for the whole document.

Filed separately as #942 with the measurements and the options, per `CLAUDE.md`'s convention for a deliberately out-of-scope gap. This PR stays scoped to `DerivedStyle.GetRemHeight()`.

## Tests

Added `RemResolvesAgainstRootElementIntegrationTests` — four fixtures, every one declaring a non-default root font size, which is what makes the bug visible at all. Expected values are written as literals rather than read back off the tree, so an assertion cannot agree with the bug. Two of the four fail on `main` and pass here.

`PeachPDF.Tests` on this branch: **10,202 passed, 0 failed, 9 skipped** — `main`'s 10,198 plus the four new fixtures. (Run against a single `net9.0` target, which is the SDK I have here; upstream multi-targets `net8.0;net10.0`, so the absolute totals differ from CI's.)

### DIFF
--- a/src/PeachPDF.Tests/Integration/RemResolvesAgainstRootElementIntegrationTests.cs
+++ b/src/PeachPDF.Tests/Integration/RemResolvesAgainstRootElementIntegrationTests.cs
@@ -1,0 +1,121 @@
+using PeachPDF.Adapters;
+using PeachPDF.Html.Core;
+using PeachPDF.Html.Core.Dom;
+using PeachPDF.PdfSharpCore.Drawing;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PeachPDF.Tests.Integration
+{
+    /// <summary>
+    /// css-values-3 §5.1.2: <c>rem</c> resolves against the font size of the ROOT ELEMENT.
+    /// <see cref="DerivedStyle.GetRemHeight"/> used to walk to the topmost <see cref="CssBox"/>, which is
+    /// the container's synthetic root sitting ABOVE <c>&lt;html&gt;</c> and carrying no element, so its
+    /// font size is always the resolver's default — every <c>rem</c> in a document that declares
+    /// <c>html { font-size }</c> silently resolved against that default instead.
+    /// <para>
+    /// Every fixture here declares a non-default root font size, which is what makes the bug visible: with
+    /// the root left at the UA default the wrong answer and the right one are the same number, which is why
+    /// a 10,000-test suite stayed green through it. The expected values are written as literals rather than
+    /// read back off the tree, so an assertion cannot agree with the bug.
+    /// </para>
+    /// </summary>
+    public class RemResolvesAgainstRootElementIntegrationTests
+    {
+        // html { font-size: 32px } -> 32 x 0.75pt/px = 24pt (CSS Values & Units §6.2).
+        private const double RootFontPt = 24.0;
+
+        [Fact]
+        public async Task RemFontSize_ResolvesAgainstTheRootElement_NotTheEngineDefault()
+        {
+            var root = await BuildAndLayout("""
+                html { font-size: 32px; }
+                body { font-size: 10pt; margin: 0; }
+                """,
+                "<div id='rem' style='font-size:1rem'>x</div>");
+
+            Assert.Equal(RootFontPt, FindById(root, "rem")!.ActualFont.Size, 3);
+        }
+
+        [Fact]
+        public async Task EmFontSize_StillResolvesAgainstTheParent()
+        {
+            // The contrast case: em is the PARENT's font size, so the two units must disagree here.
+            // Without it, a `rem` that wrongly followed the parent would pass the test above whenever
+            // the parent happened to be the root.
+            var root = await BuildAndLayout("""
+                html { font-size: 32px; }
+                body { font-size: 10pt; margin: 0; }
+                """,
+                "<div id='em' style='font-size:1em'>x</div>");
+
+            Assert.Equal(10.0, FindById(root, "em")!.ActualFont.Size, 3);
+        }
+
+        [Fact]
+        public async Task RemLength_ResolvesAgainstTheRootElement()
+        {
+            // Not only font-size: every length that takes a rem reads the same basis.
+            var root = await BuildAndLayout("""
+                html { font-size: 32px; }
+                body { font-size: 10pt; margin: 0; }
+                """,
+                "<div id='target' style='padding-top:2rem'>x</div>");
+
+            Assert.Equal(2 * RootFontPt, FindById(root, "target")!.ActualPaddingTop, 3);
+        }
+
+        [Fact]
+        public async Task RemOnTheRootElementItself_ResolvesAgainstTheInitialValue()
+        {
+            // css-values-3 §5.1.2's anti-recursion rule: a rem in the root element's OWN font-size
+            // refers to the initial value, not to the size being computed.
+            var root = await BuildAndLayout("html { font-size: 2rem; }", "<div id='x'>x</div>");
+
+            var html = FindByTag(root, "html")!;
+            Assert.Equal(2 * DefaultFontSizePt(root), html.ActualFont.Size, 3);
+        }
+
+        private static double DefaultFontSizePt(CssBox root) => root.ActualFont.Size;
+
+        private static async Task<CssBox> BuildAndLayout(string css, string bodyHtml)
+        {
+            var html = $"<!DOCTYPE html><html><head><style>{css}</style></head><body>{bodyHtml}</body></html>";
+
+            var adapter = new PdfSharpAdapter { PixelsPerPoint = 1.0 };
+            var container = new HtmlContainerInt(adapter);
+            await container.SetHtml(html, null);
+
+            var size = new XSize(595, 842);
+            container.PageSize = PeachPDF.Utilities.Utils.Convert(size, 1.0);
+            container.MaxSize = PeachPDF.Utilities.Utils.Convert(size, 1.0);
+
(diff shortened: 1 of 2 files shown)


jhaygood86:
Reviewed against `CLAUDE.md` conventions. Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` (clean, 0 warnings) and the FontSize/Rem/Em-filtered test suite (1569 passed). Traced the two cases through by hand and both are correct: an ordinary descendant walking up to `<html>` now correctly finds the outermost `HtmlTag`-bearing box (tracked by always overwriting `rootElement` while continuing to the true top, so the *last* assignment is the *outermost* one), and `<html>` resolving `rem` on **itself** correctly falls through to the default per css-values-3 §5.1.2's anti-recursion rule. Good fix for the case it targets, and a real, high-impact bug (as you note, this silently miscales every `rem`-based stylesheet, which is most of them).

**1. Missing test coverage (please add before merge)**

Given how easily this slipped in — the whole suite (10,207 tests) stayed green while `rem` was completely broken against a non-default root font — this is exactly the kind of change that needs a regression test, not just a manual `pdftotext -bbox` repro in the PR description. A test using your own repro shape (`html { font-size: 32px }`, then assert an element's resolved font size for `1rem` matches `1em`/`32px`, e.g. via `CssBox.ActualFont.Size` after layout) would do it directly, and would have caught this the first time.

**2. This does not fix the same bug in `@page` context — flagging since it's directly related (probably a separate/follow-up issue, not necessarily in scope here)**

`DomParser.ApplyPageStylesOnce` computes `@page` margin/size's `rem`/`em` basis via `root.GetRemHeight()` / `root.GetEmHeight()` (`DomParser.cs:370,375`) — but the `root` passed in there is the **synthetic wrapper box** `GenerateCssTree` creates (`root.IsRoot = true`, "not the document's actual `<html>` element" per that method's own comment at line 78), not `<html>` itself. `GetRemHeight()`'s walk only goes *upward* from `Owner.ParentBox`, so calling it directly on that wrapper (which has nothing above it) can never reach `<html>`, in either the old or the new implementation — it just falls back to the default both before and after this PR.

I confirmed this empirically on this branch: with `html { font-size: 32px }` and `@page { margin-top: 1rem }`, `container.MarginTop` still resolves to `11` (the default), not `24` (32px × 0.75pt/px, matching `html.GetEmHeight()` which correctly reads `24`). `Root.GetRemHeight()` and `Root.GetEmHeight()` both still report `11`.

This means `PageMarginUnitConsistencyIntegrationTests.BaseRule_EmMargin_NoPageFontSize_StaysRootBased` (and the `1.5rem`/`2em` cases in `BaseRule_And_FirstPageOverride_ResolveIdenticalMarginTop`) don't actually catch it either — they assert against `container.Root!.GetEmHeight()` on both sides of the comparison, which is self-consistent but not independently correct (no fixture in that file sets a non-default `html { font-size }`, so the bug is invisible there the same way it was invisible in the main content case before this PR).

Given this PR is scoped to `DerivedStyle.GetRemHeight()` specifically, I'd suggest either:
- expanding this PR to also fix `ApplyPageStylesOnce` (likely by resolving against the actual `<html>` box, e.g. via `DomUtils.GetBoxByTagName(root, "html")`, the same lookup `GenerateCssTree` already does for `DocumentLanguage`/`RootWritingMode`), or
- filing a follow-up issue for it and noting the scope boundary in this PR's description, per `CLAUDE.md`'s convention for a deliberately out-of-scope gap.

Happy to be wrong here if there's context I'm missing about why the page-context path is intentionally different — but the empirical result above seems hard to explain otherwise.

dchupp:
Both addressed — and your point 2 turned out to be more interesting than either of us expected. Force-pushed, and the `@page` half is now #942.

**1. Regression test.** New file, `Integration/RemResolvesAgainstRootElementIntegrationTests.cs`, four fixtures:

- `RemFontSize_ResolvesAgainstTheRootElement_NotTheEngineDefault` — your suggested shape: `html { font-size: 32px }`, assert `1rem` computes to 24pt via `ActualFont.Size`.
- `EmFontSize_StillResolvesAgainstTheParent` — the contrast case (`body { font-size: 10pt }`, `1em` → 10pt), so the units are asserted to *disagree*. Without it, a `rem` that wrongly followed the parent passes whenever the parent happens to be the root.
- `RemLength_ResolvesAgainstTheRootElement` — `padding-top: 2rem`, since it is not only `font-size` that reads the basis.
- `RemOnTheRootElementItself_ResolvesAgainstTheInitialValue` — the anti-recursion arm.

Every fixture declares a non-default root font size, which is the thing that makes the bug visible at all — with the root left at the UA default the wrong answer and the right one are the same number, which is how 10,000 tests stayed green through it. Expected values are written as literals rather than read back off the tree, so an assertion cannot agree with the bug. Two of the four fail on `main`.

**2. `@page`.** You're right, and thank you for the empirical check — I reproduced `container.MarginTop` = 11. I took your first option first and tried expanding this PR with `DomUtils.GetBoxByTagName(root, "html")`, and it silently breaks `rem` for the **whole document**:

`ApplyPageStylesOnce` runs *before* `CascadeApplyStyles` (deliberately — the comment at `DomParser.cs:303`), so `<html>`'s `font-size` has not been cascaded onto its box yet. And `DerivedStyle.ActualFont` memoizes on first read (`DerivedStyle.cs:963`). So reading `html.GetEmHeight()` there caches the default font on the root element permanently. Same document, box tree after layout:

```
without the change                          with it
  (anon)  font=11  rem=11                     (anon)  font=11  rem=11
  html    font=24  rem=11                     html    font=11  rem=11
    head  font=24  rem=24                       head  font=24  rem=11
    body  font=10  rem=24                       body  font=10  rem=11
      div font=24  rem=24                         div font=11  rem=11
```

The right column is the "fix": `<html>` stuck at the default, and with it every `rem` in the document, including the ones this PR had just repaired. Nothing throws; the suite stays green.

So I've gone with your second option — filed as **#942**, with the measurements above and three candidate approaches (late capture of the em/rem bases, a standalone resolve of the root's `font-size` out of `cssData`, or deciding the page context resolves against the initial font per css-page-3 §3). There's a real ordering cycle in there — page margins need the root font, which needs the cascade, which for an `@page { size }` document runs after the page size, which depends on the margins — so I'd rather you pick the direction than guess. Scope boundary noted in this PR's description.

I've also noted in #942 that the existing `PageMarginUnitConsistency` fixtures can't catch it, for the reason you gave.

Full suite on this branch: 10,202 passed, 0 failed, 9 skipped; 0 build warnings. (Single `net9.0` target locally; upstream multi-targets `net8.0;net10.0`, so absolute totals differ from CI's.)

jhaygood86 (approved):
Verified the update: four solid integration tests added, each deliberately using a non-default root font size (the exact condition that let the original bug hide from the whole suite) — including the `em`-vs-`rem` contrast case and the root-resolving-its-own-`rem` anti-recursion case. Confirmed all pass and the full solution builds clean.

Really appreciated the depth on the `@page` follow-up: you didn't just note it as out of scope, you tried the "obvious" fix (`DomUtils.GetBoxByTagName(root, "html")`), found it silently breaks `rem` for the *entire* document via `ActualFont`'s memoization firing before the cascade runs, and filed jhaygood86/PeachPDF#942 with the measured before/after box tree and three concrete options. That's exactly the kind of due diligence this repo's `.claude/accepted-gaps` convention is meant to capture — nice work.

PR #933, opened 2026-09-08
A float does not land below the line it belongs to

dchupp:
CSS 2.1 §9.5.1 rule 6: a float's outer top may not be higher than the top of the line box it appears in — and it must not be pushed below it either. `CssBox.CommitBlockChildOffset` lays a float out as an ordinary block child, so inline content before it has already closed a line and the float lands on the *next* one.

## Repro

```html
body { margin: 0; font-family: Arial, sans-serif; font-size: 12pt; }
.cell { width: 300px; border: 1px solid #999; }
.badge { float: left; border: 1px solid #000; padding: 2px; }
</style></head><body>
<div class="cell"><b>BILLTO</b><span class="badge">1038348</span></div>
</body></html>
```

Word positions via `pdftotext -bbox`:

| --- | --- | --- |
| `BILLTO` yTop | 36.75 | 36.75 |
| **`1038348` (the float) yTop** | **53.40** | **39.00** |

On `main` the float sits a full line (16.65 pt) below the text it belongs beside. With the change it sits on that line, offset only by its own border and padding.

The same document with `float: right` behaves identically in both builds — the defect is vertical placement, not side.

## Why it matters

This is the common "badge beside a heading in a table cell / grid item" idiom. The float dropping a line pushes everything after it down, and on a paginated document that difference accumulates into a changed page count.

## Tests

### DIFF
--- a/src/PeachPDF.Tests/Integration/FloatLayoutRegressionTests.cs
+++ b/src/PeachPDF.Tests/Integration/FloatLayoutRegressionTests.cs
@@ -429,6 +429,61 @@ namespace PeachPDF.Tests.Integration
                 $"clear:right must not clear past a float:left sibling, was pushed to Y={cleared.Location.Y}");
         }
 
+        [Fact]
+        public async Task FloatAfterAnInlineRun_LandsOnThatRunsLine()
+        {
+            // CSS 2.1 §9.5.1 rule 6: a float's outer top may not be lower than the top of the line box
+            // it appears in. The inline run before the float closes a line, and the float is laid out as
+            // an ordinary block child, so without the rule it starts on the line BELOW its own.
+            var html = Wrap(@"
+                <div style='width:300pt;'>
+                    <b>BILL TO</b><div id='badge' style='float:left; width:40pt; height:12pt;'></div>
+                </div>");
+
+            var (root, _) = await BuildAndLayout(html);
+            var badge = FindById(root, "badge")!;
+            var word = FindFirstWord(root)!;
+
+            Assert.True(badge.Location.Y <= word.Rectangle.Top + 0.001,
+                $"the float belongs on the line it follows (top {word.Rectangle.Top}), was at Y={badge.Location.Y}");
+        }
+
+        [Fact]
+        public async Task FloatAfterAnInlineRun_WithClear_StillGoesBelowTheLine()
+        {
+            // `clear` is ClearBox's job and rule 6 must not pre-empt it: the same shape with clear:left
+            // keeps the float below the line.
+            var html = Wrap(@"
+                <div style='width:300pt;'>
+                    <div style='float:left; width:20pt; height:30pt;'></div>
+                    <b>BILL TO</b><div id='badge' style='clear:left; float:left; width:40pt; height:12pt;'></div>
+                </div>");
+
+            var (root, _) = await BuildAndLayout(html);
+            var badge = FindById(root, "badge")!;
+
+            Assert.True(badge.Location.Y >= 30,
+                $"clear:left must still clear the 30pt float, was at Y={badge.Location.Y}");
+        }
+
+        [Fact]
+        public async Task FloatAfterARealBlockSibling_StartsBelowIt()
+        {
+            // The rule is scoped to a float whose preceding sibling is the ANONYMOUS block the parser
+            // wrapped an inline run in. After a real block-level element a float starts below it, which
+            // is what a browser does — so the preceding-sibling test has to distinguish the two.
+            var html = Wrap(@"
+                <div style='width:300pt;'>
+                    <p style='margin:0; height:30pt;'>BILL TO</p><div id='badge' style='float:left; width:40pt; height:12pt;'></div>
+                </div>");
+
+            var (root, _) = await BuildAndLayout(html);
+            var badge = FindById(root, "badge")!;
+
+            Assert.True(badge.Location.Y >= 30,
+                $"a float after a real block sibling starts below it, was at Y={badge.Location.Y}");
+        }
+
         // ── Helpers ────────────────────────────────────────────────────────────
 
         private static string Wrap(string body) =>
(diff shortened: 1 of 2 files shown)


jhaygood86:
Reviewed against `CLAUDE.md` conventions. Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` (clean, 0 warnings) and the `Float`-filtered test suite (87 passed). Also cross-checked the cited spec text directly — CSS 2.1 §9.5.1 rule 6 is indeed *"The outer top of an element's floating box may not be higher than the top of any line-box containing a box generated by an element earlier in the source document"* — and traced that `CommitBlockChildOffset` applies the corrected `top` to `child.Location` **before** calling `CssLayoutEngine.FloatBox(child)`, so the horizontal (left/right) placement scan that follows correctly runs at the fixed-up Y rather than the stale one. Logic checks out.

Same as the other float PR (#930): `CLAUDE.md`'s testing conventions require a layout-engine assertion on the affected `CssBox`'s properties after layout. `FloatLineTop` has real branching (not floated, has `clear`, no preceding sibling, preceding sibling isn't the anonymous-inline wrapper, wrapper has no line boxes, wrapper's line starts at/after the naive top) none of which is exercised by an automated test — could you add one based on your repro (bordered cell, `<b>` text then a `float: left` badge, assert the float's `Location.Y` lands on the text's line rather than below it)?

**2. The `FloatLineTop` doc comment has a couple of things worth cleaning up before merge**

- It's tagged `VENDORED:` — that prefix normally means "copied from a third-party/vendor source" (see how `Fonts/OpenType/` keeps its PDFsharp-origin license headers). This is original code, so the tag reads as a mistake and should probably just be dropped.
- It names a specific example ("Milton's AR invoice is the case: a bordered customer-code badge that belongs beside 'BILL TO' was drawn a line below it") — `CLAUDE.md`'s comment guidance is explicit that a comment shouldn't reference the current fix/task/caller ("those belong in the PR description... and rot as the codebase evolves"). Could you replace that sentence with the generic shape it's really describing (which the rest of the comment already does well), and drop the "Milton" reference?
- Same doc-comment-detachment issue flagged on #930/#929: the new `FloatLineTop` method + its doc comment were inserted between `CommitBlockChildOffset`'s existing doc comment ("Writes the offset `ResolveBlockChildOffset` decided on, positions `child` under whichever positioning scheme it uses...") and that method itself — so `CommitBlockChildOffset` loses its real doc comment and `FloatLineTop` inherits one describing a different method. This happens not to produce a compiler warning here only because `FloatLineTop` also happens to have a `child` parameter (unlike the CS1734 it caused on #929) — but it's still wrong/misleading. Could you move `FloatLineTop` (with its own doc) above `CommitBlockChildOffset`'s existing comment, or move that comment back down?

This is the third PR in this batch with the same doc-comment-detachment pattern (also #930, #929) — might be worth double-checking however these are being authored for it generally.

Solid fix otherwise: scoping to "immediately preceding sibling is the anonymous inline-wrapper block with no `HtmlTag`" correctly targets the anonymous-block-generation shape this box tree actually produces (confirmed by tracing `DomParser.CorrectInlineBoxesParent`/`JoinsTheInlineRun` — a float is blockified so it never joins the inline run, which is exactly why it ends up as a sibling of the anonymous wrapper rather than nested inside it, for both a leading and mid-paragraph float).

dchupp:
All addressed, force-pushed.

**1. Layout test.** Three fixtures in `Integration/FloatLayoutRegressionTests.cs`:

- `FloatAfterAnInlineRun_LandsOnThatRunsLine` — the repro shape (`<b>` text then a `float: left` badge), asserting the float's `Location.Y` is at or above the line's own top rather than below it. On `main` it lands at Y=19.2 against a line top of 6; here it lands on the line.
- `FloatAfterAnInlineRun_WithClear_StillGoesBelowTheLine` — `clear` stays `ClearBox`'s job and rule 6 must not pre-empt it.
- `FloatAfterARealBlockSibling_StartsBelowIt` — the preceding-sibling test has to distinguish the anonymous inline wrapper from a real block-level element, so this pins the other side of that branch.

The two contrast cases pass on `main` as well as here, which is the point of them — only the first moves.

**2a. The `VENDORED:` tag.** Dropped, and you were right to read it as a mistake: it is our internal marker for "this comment came from our patch set", it means nothing in this repo, and it should never have travelled. I've stripped it from every branch in this batch, not just this one.

**2b. The specific example.** Removed, along with every other one across the whole batch — and I owe you a fuller apology than "fixed". Those were **customer names**. They should not have been in a public PR at all, and that is on me, not on the comment convention. Three branches carried them (this one, #937, #940) and they are all rewritten now, to the generic shape the comments were really describing:

> A bordered badge that belongs beside a heading in the same table cell was drawn a line below it.

I've also re-read `CLAUDE.md`'s comment guidance properly. You're right that the example belonged in the PR description in the first place, whatever the name in it.

**2c. Doc-comment detachment.** Moved `FloatLineTop` above `CommitBlockChildOffset`'s existing doc comment. And you were right that it was systematic — I've swept all fourteen branches for the shape (one contiguous `///` run carrying more than one `<summary>`, which is what catches it whether or not it trips CS1734) and fixed every instance that's ours. Four remain that pre-date these branches and are on `main` today; details in my reply on #929.

Full suite on this branch: 10,201 passed, 0 failed, 9 skipped; 0 build warnings. (Single `net9.0` target locally; upstream multi-targets `net8.0;net10.0`, so absolute totals differ from CI's.)

jhaygood86 (approved):
Verified the update: three tests added covering the fix itself, the `clear` carve-out, and the real-block-sibling contrast case. The `VENDORED:` tag and the specific customer-example reference are gone, replaced with a generic description. The doc comment is now correctly attached to `FloatLineTop` alone (no longer sandwiched in front of `CommitBlockChildOffset`'s). Clean rebuild (0 warnings), float-filtered suite passes (90 tests).

PR #934, opened 2026-09-08
A data: URI has no length limit

dchupp:
An `` whose `data:` URI runs past **65,519 characters renders blank** — no error, no warning, just a missing image.

Nothing to do with image size or decoding. **Before .NET 10**, `System.Uri` refuses any URI string longer than 65,519 characters outright (`Invalid URI: The Uri string is too long.`), and `RUri`'s base-relative constructors hand every reference straight to it. `CommonUtils.ResolveAgainstDocumentBase` catches the `UriFormatException` and returns null, so the image is silently dropped.

## Repro

One `` per document, a PNG inlined as base64:

| --- | --- | --- |
| data URI 6,566 chars | 1 image XObject, 3,018 bytes | 1 image XObject |
| **data URI 144,806 chars** | **0 image XObjects, 990 bytes** | **1 image XObject, 7,622 bytes** |

Counted as `/Subtype /Image` occurrences in the output. The threshold is exactly `System.Uri`'s 65,519, on net8.0. **.NET 10 has no ceiling** (measured: a 5,000,023-character payload constructs fine) and round-trips a `data:` URI exactly, so it keeps using `System.Uri` — the fix is gated with `#if NET10_0_OR_GREATER` and only the older targets take the string-holding path. Flattening both targets onto that path would have lost .NET 10's percent-escaping of a non-base64 payload.

<details>

```python
import base64, struct, zlib, random
def png(n):
    def chunk(tag, data):
        body = tag + data
        return struct.pack(">I", len(data)) + body + struct.pack(">I", zlib.crc32(body) & 0xFFFFFFFF)
    random.seed(7)
    raw = b"".join(b"\x00" + bytes(random.randrange(256) for _ in range(n*3)) for _ in range(n))
    return (b"\x89PNG\r\n\x1a\n" + chunk(b"IHDR", struct.pack(">IIBBBBB", n, n, 8, 2, 0, 0, 0))
            + chunk(b"IDAT", zlib.compress(raw, 1)) + chunk(b"IEND", b""))

b64 = base64.b64encode(png(190)).decode()   # ~145k chars, over the limit
html = ("<!DOCTYPE html><html><head><style>body{margin:0}"
        "img{width:120px;height:120px;border:1px solid #000}</style></head><body>"
        f"<img src='data:image/png;base64,{b64}'></body></html>")
```
(`png(40)` gives the ~6.5k-char control that renders fine.)
</details>

## Why it matters

Inlining images as `data:` URIs is the normal way to make an HTML document self-contained for rendering — no network, no file paths, nothing to resolve. Any image past ~48 KB of binary crosses the limit once base64-encoded, which is an ordinary logo or photograph. The failure is silent, so it surfaces as "the picture is missing from the PDF" with nothing in any log.

## The change

A `data:` URI is not a network reference and needs no base resolution — it carries its payload inline. `RUri` now recognises the scheme before the string ever reaches `System.Uri`.

## Tests

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-large-inline-data-uri-image-now-renders-before-net-10.md
+++ b/.claude/migration-notes/2026-09-08-a-large-inline-data-uri-image-now-renders-before-net-10.md
@@ -1,0 +1,22 @@
+# A large inline `data:` image now renders on .NET 8 (it previously vanished)
+
+Before .NET 10, `System.Uri` rejects any URI string longer than 65,519 characters — roughly 48 KB of
+payload once base64 expansion is accounted for. `RUri`'s single-string constructors already worked
+around that by holding a `data:` URI verbatim, but the two **base-relative** constructors did not,
+and those are the ones an `<img src="data:…">` actually reaches (via
+`CommonUtils.ResolveAgainstDocumentBase`). The resulting `UriFormatException` is caught upstream and
+turned into a skipped resource, so an embedded logo, photograph or chart past that size rendered as
+nothing at all, with no error surfaced to the caller.
+
+Such an image now renders. There is no length limit on a `data:` URI on any target framework.
+
+Two smaller changes ride along:
+
+- The `data:` scheme is now matched case-insensitively (RFC 3986 §3.1), so a document writing
+  `DATA:` takes the same path.
+- On .NET 10 nothing changes: its `System.Uri` already round-trips a `data:` URI exactly and has no
+  length ceiling, so it keeps using it — including the percent-escaping of a non-base64 payload
+  (`data:text/plain,Hello World!` → `data:text/plain,Hello%20World!`) that the pre-.NET-10 path does
+  not do.
+
+See [Embedded Content](../../docs/html-css-support.md#embedded-content) in `docs/html-css-support.md`.
--- a/.claude/recent-fixes/2026-09-08-data-uri-bypass-is-per-target-framework-not-unconditional.md
+++ b/.claude/recent-fixes/2026-09-08-data-uri-bypass-is-per-target-framework-not-unconditional.md
@@ -1,0 +1,52 @@
+# The `data:` URI bypass is per-target-framework, and it belongs in one predicate
+
+`RUri` held a `data:` URI verbatim rather than constructing a `System.Uri`, but only in its two
+single-string constructors. The two base-relative constructors — `RUri(RUri, string)` and
+`RUri(RUri, RUri)` — had no `data:` handling at all, and those are the ones the resource pipeline
+actually calls: `CommonUtils.ResolveAgainstDocumentBase` builds every `<img src>` through
+`new RUri(baseUri, src)`. So the workaround was in place and the bug still shipped.
+
+## What running it turned up
+
+- **The first fix deleted the `#if NET10_0_OR_GREATER` gate, and that was a regression.** Collapsing
+  both targets onto the string-holding path looks like a simplification, but .NET 10's `System.Uri`
+  is strictly better for this: it round-trips a `data:` URI exactly, has no length ceiling
+  (a 5,000,023-character payload constructs fine), and percent-escapes a non-base64 payload, which
+  holding the raw string does not — `data:text/plain,Hello World!` comes back escaped there and
+  verbatim before .NET 10. The gate is back, and
+  `RUriTests.DataUri_NonBase64Payload_EscapesOnNet10AndIsVerbatimBefore` pins the difference on both
+  targets so it cannot be flattened again unnoticed.
+- **The gate belongs in one predicate, not in four constructors.** `MustBypassSystemUri` is
+  constantly `false` on .NET 10 and every constructor asks it. Repeating an `#if` per constructor is
+  precisely how the two base-relative ones ended up uncovered in the first place.
+- **The 65,519 ceiling is pre-.NET-10 only.** An earlier version of this doc comment said "on every
+  framework", which is wrong and would have left a future reader believing .NET 10 has a defect it
+  demonstrably does not.
+- **Gating the bypass creates a second per-target difference, in scheme casing.** RFC 3986 §3.1
+  makes a scheme case-insensitive with lower case canonical, and `System.Uri` normalizes to it — so
+  an authored `DATA:` comes back as `data:` on .NET 10 and survives verbatim before it. A test
+  asserting byte-equality against the authored string therefore passes on net8.0 and fails on
+  net10.0; CI caught exactly that. The fixture now asserts the invariant that holds on both (the
+  payload survives, `Scheme` is `data`) and pins the casing difference per target.
+- **The net10 path is simulatable on net8.** Forcing `MustBypassSystemUri` to `false` runs every
+  `data:` fixture through `System.Uri` on net8.0, which is the .NET 10 code path minus the length
+  ceiling: any failure that is not `UriFormatException: The Uri string is too long` is a real .NET 10
+  failure. That is how the scheme-casing defect was isolated without a .NET 10 SDK to hand.
+- **`RUri.Uri` is not covered by the bypass**, and reconstructing from the held string hits the same
+  ceiling. No caller reaches it for a `data:` URI — the three that read it are selected by `Scheme`
+  first — but it is now documented on the property, pointing a new caller at `AbsoluteUri` instead.
+
+## Evidence
+
+Five fixtures added to `RUriTests`, three of which fail against the merge base: both base-relative
+constructors with a 100 KB payload, the case-insensitive scheme, an ordinary relative reference as
+the contrast case (the short-circuit must not swallow it), and the per-TFM escaping difference.
+Full suite green on net8.0, 0 build warnings, diff coverage 100% on the changed lines.
+
+## Deliberately not done
+
+No general resource-size cap. Removing an incidental ceiling on untrusted-HTML input deserves the
+scrutiny, but this one was a `System.Uri` buffer artifact rather than a guard, and nothing else in
+`Network`/`Html/Core/Utils` bounds resource size either — many small `data:` URIs already reach the
+same place. Tracked separately as a configurable option rather than left as an implicit side effect
+of a parser limit.
--- a/src/PeachPDF.Tests/Network/RUriTests.cs
+++ b/src/PeachPDF.Tests/Network/RUriTests.cs
@@ -55,6 +55,101 @@ namespace PeachPDF.Tests.Network
             Assert.Equal(input, uri.OriginalString);
         }
 
+        // ─── The base-relative constructors (the path the reported bug actually took) ───────────
+        //
+        // An <img src="data:..."> reaches RUri through CommonUtils.ResolveAgainstDocumentBase, which
+        // calls new RUri(baseUri, src) — not the single-string constructor the fixtures above cover.
+        // Those two overloads had no data: handling at all, so a large inline image was dropped
+        // before .NET 10 however well the string constructors behaved.
+
+        [Fact]
+        public void DataUri_ThroughBaseRelativeStringConstructor_RoundTripsExactly()
+        {
+            var bytes = new byte[100_000];
+            new Random(42).NextBytes(bytes);
+            var input = $"data:application/octet-stream;base64,{Convert.ToBase64String(bytes)}";
+            var baseUri = new RUri("https://example.com/docs/page.html");
+
+            var uri = new RUri(baseUri, input);
+
+            // RFC 3986 §5.2.2: a reference carrying its own scheme is already absolute, so the base
+            // is not consulted and the result is the data: URI unchanged.
+            Assert.Equal(input, uri.AbsoluteUri);
+            Assert.Equal("data", uri.Scheme);
+            Assert.True(uri.IsAbsoluteUri);
+        }
+
+        [Fact]
+        public void DataUri_ThroughBaseRelativeRUriConstructor_RoundTripsExactly()
+        {
+            var bytes = new byte[100_000];
+            new Random(42).NextBytes(bytes);
+            var input = $"data:application/octet-stream;base64,{Convert.ToBase64String(bytes)}";
+            var baseUri = new RUri("https://example.com/docs/page.html");
+
+            var uri = new RUri(baseUri, new RUri(input));
+
+            Assert.Equal(input, uri.AbsoluteUri);
+            Assert.Equal("data", uri.Scheme);
+        }
+
+        [Fact]
+        public void RelativeReference_ThroughBaseRelativeConstructor_StillResolvesAgainstTheBase()
+        {
+            // The contrast case: the data: short-circuit must not swallow an ordinary relative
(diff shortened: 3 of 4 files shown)


jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (test coverage, zero-warning build, spec/claim verification). Verified by checking out this branch, building the full solution, running the existing `RUri` tests on both TFMs, and independently reproducing the underlying `System.Uri` behavior outside the repo.

`gh pr diff 934 --name-only` shows only `src/PeachPDF/Network/RUri.cs` changed — no test file. `CLAUDE.md`'s 90%-diff-coverage gate applies here, and the existing `src/PeachPDF.Tests/Network/RUriTests.cs` already has exactly the fixtures this change touches:

- `RUri(RUri, string)` — the base-relative string constructor — has no test for a `data:` URI argument at all (only the plain string constructors are tested at lines 11-56).
- `RUri(RUri, RUri)` — the base-relative `RUri` overload — same gap.

Both of these are the actual constructors that were broken for the reported bug (an `` goes through `CommonUtils.ResolveAgainstDocumentBase`, which calls `new RUri(baseUri, src)`), and neither is exercised by the current suite. Could you add cases to `RUriTests.cs` mirroring `DataUri_LargePayloadBeyondNet8UriLengthLimit_RoundTripsExactly` but going through the two-argument constructors, so a regression here fails a unit test instead of only being caught by an end-to-end PDF-image count?

**2. The PR's core technical claim is only half right — please correct the description and the new `<remarks>`**

The description states "`System.Uri` refuses any URI string longer than 65,519 characters" as a blanket, framework-independent fact, and the new `<remarks>` block on `RUri` repeats this: "on every framework it rejects any URI string longer than 65,519 characters." I checked this independently (isolated `net8.0` vs `net10.0` console apps calling `new Uri(...)` directly on a 133,373-character `data:` string, outside this repo):

- **net8.0**: throws `UriFormatException: Invalid URI: The Uri string is too long.`
- **net10.0**: constructs fine, round-trips exactly.

This matches what's already documented in this repo — the pre-existing test `DataUri_LargePayloadBeyondNet8UriLengthLimit_RoundTripsExactly` (added in `dfa26bf2`) says in its own comment: "*.NET 10 removed this limit entirely, which is what makes it safe to drop the special-casing there.*" That's also exactly why the `#if NET10_0_OR_GREATER` branch existed pre-PR — `dfa26bf2`'s commit message says it switched net10 to "native `Uri` handling" deliberately. So the bug is real and worth fixing, but it's a **net8.0-only** (and presumably net9-and-earlier) defect, not a `System.Uri` defect on every framework PeachPDF targets. Collapsing the two code paths into one is still a reasonable simplification (it's harmless on net10, since the short-circuit path is taken instead of a `System.Uri` call that would have succeeded anyway), but the doc comment and PR description should say "before .NET 10" rather than "every framework," or a future reader will conclude .NET 10 has the same defect when it demonstrably doesn't.

**3. The new `<remarks>` overstates what's fixed — the `Uri` property is still subject to the old limit**

Line 19's claim "Keeping the original string means a `data:` URI has no length limit at all" isn't quite true: `RUri.Uri` (line 117) is `_uri ?? new Uri(_originalUri!)` — unchanged by this PR — so calling `.Uri` on an oversized `data:` URI still throws `UriFormatException` on net8.0/net9.0. It happens that none of the three current call sites of `.Uri` (`HttpClientNetworkLoader.cs:57`, `FileUriNetworkLoader.cs:78`, `MimeKitNetworkLoader.cs:51`) are ever reached for a `data:` scheme, so this isn't exploitable today, but it's a latent trap for the next person who adds a `.Uri` call without checking `Scheme` first. Worth a one-line caveat in the doc comment on the `Uri` property itself, or better, having it throw a clearer `NotSupportedException` for the `data:` case instead of silently reconstructing a `System.Uri` that can fail the same way this PR is fixing.

**4. DoS/resource-exhaustion consideration — looked at, not a blocker**

Removing an incidental length ceiling on untrusted-HTML input is worth real scrutiny, so I checked: this 65,519-char ceiling was never a deliberate safety guard — it's a `System.Uri` internal buffer-size artifact, and the original `c1e3989a` commit ("Switch to using RUri... to support data: URIs above roughly 50 KB") already worked around it for exactly this reason back in 0.7.10. There's also no other resource-size cap in `Network`/`Html/Core/Utils` today (checked for `MaxLength`/`SizeLimit`/similar), so an attacker could already exhaust memory via many small `data:` URIs or one large non-`data:` resource well before this change. This PR doesn't meaningfully change the DoS surface — but given there's no general guard anywhere, it might be worth a separate tracked issue for a configurable max-resource-size option on `PdfGenerateConfig`, independent of this fix.

**5. Doc-comment placement — no detachment found this time**

Checked specifically for the pattern flagged in earlier PRs from this contributor (a new method + doc comment inserted above an existing method's leading `///`, detaching it). Not present here — every constructor/property kept its own doc comment correctly attached, and the new `IsDataUri` helper (line 30) is a one-line private static predicate with no adjacent doc-comment collateral damage. `SvgPropertyContext`/`ArgumentNullException.ThrowIfNull` usage etc. also look untouched.

**6. Build and other tests**

`dotnet build PeachPDF.slnx -t:Rebuild -v:minimal` from `src/`: **0 Warning(s), 0 Error(s)**. `RUriTests` pass on both `--framework net8.0` and `--framework net10.0` (5/5 each). No migration-notes file was added under `.claude/migration-notes/` — this is a user-visible behavior change (large inline images that previously silently vanished now render), which the repo's convention asks for; worth adding given a document author could genuinely notice PDFs changing.

Solid, well-targeted fix for a real and well-documented bug (nice repro table in the description). The two blocking items from my read are #1 (no regression test for the actual broken code path) and #2 (correcting the "every framework" claim so the historical record stays accurate) — everything else is a nice-to-have.

jhaygood86:
Follow-up to my review above — one more thing worth separating out, since it's a distinct issue from the "every framework" claim I already flagged.

**The `#if NET10_0_OR_GREATER` split shouldn't have been removed entirely — it should still gate on length, not disappear**

Before this PR, `RUri.cs` had two code paths (`#if NET10_0_OR_GREATER` / `#else`): on .NET 10, every `data:` URI — short or long — went straight to `_uri = new Uri(uriString)`; only pre-.NET-10 held the raw string in `_originalUri`. This PR deletes the `#if` and makes **all** target frameworks, including .NET 10, take the string-holding path unconditionally (`RUri.cs:41-48`, `56-65`, `96-100`).

I checked both halves of this empirically before writing this, rather than assuming:

1. **Does `System.Uri` do anything different for a short `data:` URI on .NET 10 than just returning the raw string?** Yes, for the non-base64 case. A quick throwaway net10.0 console app:
   ```csharp
   new Uri("data:text/plain;base64,SGVsbG8sIFdvcmxkIQ==").AbsoluteUri
   // == the raw string, unchanged (base64 alphabet has nothing to escape)

   new Uri("data:text/plain,Hello World! #comment").AbsoluteUri
   // -> "data:text/plain,Hello%20World!%20#comment"  (NOT equal to the raw string)
   ```
   So for a base64 image `data:` URI (the PR's own repro case) there's no visible difference — but for a literal-text `data:` URI (common for small inline SVG fragments, CSS `content:`, etc.), pre-PR .NET 10 returned a properly percent-escaped `AbsoluteUri`; post-PR it returns the raw, unescaped string for every framework. That's a real, silent behavior change for .NET 10 on the common case this PR didn't need to touch.

2. I also re-verified the length ceiling itself only to make sure I wasn't citing a stale number: net8.0/net9.0 throw `UriFormatException` starting at 65,519 payload chars; net10.0 constructs successfully even at 5,000,023 chars (tested `new string('A', 5_000_000)` inside the base64 payload) — no ceiling at all on .NET 10. So on .NET 10 specifically, the *only* thing this PR's fix is needed for is nothing — .NET 10's `System.Uri` was already both round-trip-correct and unbounded in length for `data:` URIs before this PR.

This is exactly the case `CLAUDE.md`'s "Post-change review pass" calls out under C# API currency: *"if a meaningfully-better API only exists on the newer TFM, gate it with `#if NET10_0_OR_GREATER` (falling back to the net8.0-compatible approach in `#else`) rather than skipping it — don't downgrade the whole codebase to the lowest common denominator when the newer target can do better."* `System.Uri` on .NET 10 is that meaningfully-better API here (correct round-trip, no length ceiling, proper escaping) — the fix should keep exercising it for the case it already handled correctly, not fall back to the pre-.NET-10 workaround unconditionally.

Concretely: keep `_uri = new Uri(uriString)` as the .NET 10 fast path for `data:` URIs, and only drop to `_originalUri` when `System.Uri` can't take it — which on .NET 10 is never (given finding 2 above), so a simpler framing is: `#if NET10_0_OR_GREATER`, always call `new Uri(uriString)` for `data:` too (unchanged from before this PR); `#else`, keep this PR's `_originalUri` string-holding path, since pre-.NET-10 needs it regardless of length (round-trip bug) and additionally has the 65,519-char ceiling this PR is actually motivated by. That keeps .NET 10 on its already-correct, already-unbounded path and fixes the bug exactly where it exists (pre-.NET-10), rather than trading a real .NET-10 escaping regression for a fix .NET 10 didn't need.

dchupp:
All four addressed, rebased onto `main` and force-pushed. Your follow-up was the important one and I had it wrong.

**The `#if NET10_0_OR_GREATER` gate is back.** You are right that deleting it was a regression, not a simplification: .NET 10's `System.Uri` round-trips a `data:` URI exactly, has no length ceiling, *and* percent-escapes a non-base64 payload, which holding the raw string does not. Collapsing both targets onto the older path traded a real .NET 10 escaping change for a fix .NET 10 did not need. This is `CLAUDE.md`'s "don't downgrade the whole codebase to the lowest common denominator" rule and I read straight past it.

It is one predicate rather than an `#if` per constructor:

```csharp
private static bool MustBypassSystemUri(string uriString) =>
#if NET10_0_OR_GREATER
    false;
#else
    uriString.StartsWith("data:", StringComparison.OrdinalIgnoreCase);
#endif
```

Deliberately, because repeating the `#if` per constructor is exactly how the two base-relative ones ended up with no `data:` handling in the first place — the bug this PR is about. One predicate, four call sites, no way for the targets to cover different constructors.

The per-TFM difference is pinned by `DataUri_NonBase64Payload_EscapesOnNet10AndIsVerbatimBefore`, which asserts `data:text/plain,Hello%20World!` under `#if NET10_0_OR_GREATER` and the verbatim string otherwise, so it cannot be flattened again unnoticed.

**2. The "every framework" claim** is corrected in the `<remarks>` — it is now explicitly "before .NET 10", with the .NET 10 behaviour (round-trip exact, no ceiling, measured at 5,000,023 characters, percent-escapes non-base64) as its own paragraph. I have updated the PR description too. You are right that leaving it would have had a future reader conclude .NET 10 has a defect it demonstrably does not.

**1. Tests** in `RUriTests.cs`, five of them, three failing against the merge base:

- `DataUri_ThroughBaseRelativeStringConstructor_RoundTripsExactly` and `..._ThroughBaseRelativeRUriConstructor_...` — the two constructors that were actually broken, 100 KB payload each.
- `RelativeReference_ThroughBaseRelativeConstructor_StillResolvesAgainstTheBase` — the contrast case; the `data:` short-circuit must not swallow an ordinary relative reference.
- `DataUriScheme_IsMatchedCaseInsensitively` — RFC 3986 §3.1; `DATA:` previously took the `System.Uri` path and lost the same payloads. (Also a merge-base failure.)
- The escaping case above.

**3. `RUri.Uri`.** You are right that it is still subject to the old limit. I went with the doc-comment caveat rather than `NotSupportedException`, and the reason is worth stating: on .NET 10 `Uri` works fine for a `data:` URI, so throwing pre-.NET-10 would make the two targets differ in whether the property throws at all — a bigger divergence than the one it removes. The property's `<remarks>` now says the ceiling still applies there, names the three callers and why none reaches it (they are selected by `Scheme` first), and points a new caller at `AbsoluteUri`. Happy to switch to throwing if you would rather have the loud failure.

**4. Resource-size cap.** Filed as **#944** rather than folded in — it is a new `PdfGenerateConfig` surface with an "unlimited by default or not" question that is yours, and I agree with your read that this PR does not change the DoS surface meaningfully. I checked and found no `MaxLength`/`SizeLimit` guard anywhere in `Network/` either, and noted that the decoded size is the bigger amplification for images.

**6. Migration note** added: `.claude/migration-notes/2026-09-08-a-large-inline-data-uri-image-now-renders-before-net-10.md`. You were right that a document author would notice — an embedded logo past ~48 KB previously rendered as nothing at all.

dchupp:
CI caught one — a test of mine, and a genuinely interesting one. Fixed and pushed.

```
Failed PeachPDF.Tests.Network.RUriTests.DataUriScheme_IsMatchedCaseInsensitively
  Expected: "DATA:application/octet-stream;base64,Phe6lq4EzTuZh"···
```

The only failure, on all three platforms, and net10-only in origin: **restoring the `#if NET10_0_OR_GREATER` gate creates a second per-target difference I had not accounted for.** RFC 3986 §3.1 makes a scheme case-insensitive with lower case canonical, and `System.Uri` normalizes to it — so an authored `DATA:` comes back as `data:` on .NET 10, while before .NET 10 the string is held verbatim and the authored casing survives. My fixture asserted byte-equality against the authored string, which is only true on the string-holding path.

The fixture now asserts what holds on both targets — the payload survives intact and `Scheme` is `data` — and pins the casing difference explicitly per target, next to the escaping one. Both are consequences of the gate rather than accidents, so both are worth pinning rather than papering over.

**How I isolated it without a .NET 10 SDK**, since that is the gap that let it through in the first place and is reusable: forcing `MustBypassSystemUri` to `false` on net8.0 runs every `data:` fixture through `System.Uri`, which *is* the .NET 10 code path minus the length ceiling. Any failure that is not `UriFormatException: The Uri string is too long` is a real .NET 10 failure. That produced exactly one — the scheme casing — and confirmed the escaping fixture's `#if NET10_0_OR_GREATER` branch is right (`data:text/plain,Hello%20World!`). I have written the technique into the recent-fixes entry.

The one thing that simulation cannot check is the absence of the ceiling itself on .NET 10; I am relying on your measurement there (5,000,023 characters constructing fine), and have cited it as yours in the doc comment rather than claiming it myself.

Local: `-t:Rebuild` 0 new warnings, `dotnet test --framework net8.0` green (10,226 passed). Fresh CI running now.

jhaygood86 (approved):
Verified the update: both blocking findings are addressed. `RUriTests.cs` now covers the two previously-untested constructors (`RUri(RUri, string)` and `RUri(RUri, RUri)`) with 100 KB payloads, and the "every framework" claim is corrected into two clearly separated paragraphs (before .NET 10 vs. .NET 10+). The `#if NET10_0_OR_GREATER` fast path is restored too — refactored into a single `MustBypassSystemUri` predicate shared by all four constructors rather than repeated per-constructor `#if` blocks, which is a nicer structure than what was there originally and fixes the root cause of the original bug (each constructor previously had to remember to special-case `data:` itself). Nice touch adding the migration-notes entry and documenting the remaining `Uri` property caveat rather than silently leaving it.

Approving. Merging once CI passes (already green on all three platforms).

PR #935, opened 2026-09-08
A page margin box honours its own margin and padding

dchupp:
`MarginBoxRenderer` paints a text margin box with a single `DrawString` into the rect `GetMarginBoxRect` returns. `vertical-align` is therefore the only thing that can move the content, and it offers exactly three positions — `line-height`, `height`, `margin` and `padding` are all inert.

## Repro

```html
<style>
@page {
  size: 8.5in 11in;
  margin: 1.5in;
  @bottom-center { content: "FOOTERTEXT"; padding-bottom: 24pt; }
}
body { margin: 0; font-family: Arial, sans-serif; font-size: 12pt; }
</style>
<p>BODY</p>
```

`FOOTERTEXT` position via `pdftotext -bbox`:

| --- | --- | --- |
| `padding-bottom: 0pt` | yTop = 731.78 | 731.78 |
| **`padding-bottom: 24pt`** | **731.78 — unchanged** | **719.78** |

On `main` the declaration does nothing at all. With the change the content shifts 12 pt, which is what 24 pt of bottom padding should do to vertically-centred content in a fixed-height box.

## Why it matters

css-page-3 gives margin boxes the full box model. In practice this is how running headers and footers get placed relative to the paper edge — a footer that has to sit a fixed distance above the trim, a header that needs breathing room under a rule. Without padding, the only lever is the page margin itself, which also moves the body.

## Tests


dchupp:

jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (margin-box box model, testing conventions, doc-comment hygiene, zero-warning builds, docs-with-feature). Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` (0 warnings), running the `MarginBox`/`PageMargin`-filtered suite (133 passed, all pre-existing), and reading `MarginBoxRenderer.cs`/`HtmlContainerInt.cs` in full against css-page-3 and CSS2.1 §8.3/§8.4.

`gh pr diff 935 --name-only` touches only `MarginBoxRenderer.cs` and `HtmlContainerInt.cs` — no test file changed. `CLAUDE.md` requires layout-affecting changes to have unit tests asserting the resulting geometry (position/rect), not just "renders/doesn't throw," and the 90% diff-coverage gate applies here too — `ApplyBoxModel` is new, non-trivial logic with several branches (all-zero short-circuit, per-edge clamping, negative-extent guard) that nothing currently exercises. Given the repo's own `MarginBoxRendererSizingTests.cs`/`MarginBoxRendererFontTests.cs` pattern, could you add fixtures that call `MarginBoxRenderer.ApplyBoxModel` (or exercise it via `GetMarginBoxRect` + `Render`) directly and assert the resulting `XRect`, covering: padding-only (the PR's own repro), margin-only, negative margin, and — see #2 below — an explicit `width`/`height` combined with padding/margin? The `HtmlContainerInt.LayoutMarginBoxes` `element()` path should get at least one integration test too (e.g. extending `MarginBoxElementFragmentIntegrationTests.cs`) confirming a running element's laid-out rect actually shrinks by the margin box's own padding.

**2. `ApplyBoxModel` double-subtracts margin/padding whenever `width`/`height` is explicitly declared**

`GetMarginBoxRect` (unchanged by this PR) already treats a margin box's declared `width` as the *outer* size to allocate for that box's slot in the row/column distribution (`ComputeThreeBoxSizes` feeds `PW(r)`'s raw parsed value straight in as `sizeA`/`sizeB`/`sizeC`, and the returned rect's width/height literally is that value). Per css-page-3 §5.3.2/§5.3.3, though, a margin box's `width`/`height` is the *content-box* dimension — border/padding/margin are additive on top of it (the fixed-dimension equality is literally `margin-top + border-top-width + padding-top + height + padding-bottom + border-bottom-width + margin-bottom = <page margin>`). `ApplyBoxModel` now treats the rect `GetMarginBoxRect` returns as a border-box and subtracts margin+padding from it *again* to reach the content box. For an auto-sized box (the PR's own repro: `@bottom-center` with no `width`/`height`) this happens to be correct, because the auto-share really is being used as an outer-width allocation. But for an explicit size it isn't: `@top-center { width: 200pt; padding-left: 5pt; }` now renders content 195pt wide, not the declared 200pt — the padding is subtracted from a value that was supposed to already be the content width. Worth at least a regression test pinning today's (arguably wrong) behavior here, since it's a real behavior change for anyone who has both an explicit `width`/`height` and non-zero margin/padding on the same margin box.

**3. Percentage margin/padding basis contradicts the very citation used for it**

The comment on `Len`/`basis` says "Percentages resolve against the box's own width on all four edges, as they do for any block box (CSS 2.1 §8.3, §8.4)" — but §8.3/§8.4 actually say the opposite: a margin/padding percentage (on *all four* edges, top/bottom included) resolves against the **containing block's width**, not the generated box's own width. Here `basis = rect.Width`, i.e., the margin box's *own* already-distributed width — for an auto-width box in a row of three, that's this box's one-third share, not the row's `contentWidth`. Two consequences: (a) a `%` padding value on two boxes in the same row with different resolved widths (one explicit, one auto) yields different absolute point values for the "same" percentage, which doesn't match any box-model reading; (b) for an auto-width box the basis is circular — box-model resolution feeding back into a size that was itself derived from box-model-blind distribution. Suggest resolving percentages against `contentWidth`/`contentHeight` (already computed in `GetMarginBoxRect`) the same way `GetMarginBoxRect` itself does for `width`/`height` percentages, rather than the box's own post-distribution rect.

**4. Docs not updated**

`docs/html-css-support.md`'s "Supported style properties in margin boxes" table (around line 1206) lists `color`/`font-*`/`text-align`/`vertical-align`/`width`+min/max/`height`+min/max but has no row for `margin`/`padding`. Per `CLAUDE.md` ("update the relevant doc page… in the same change, rather than as a follow-up"), this should get a row describing the new behavior — including the percentage-basis caveat once #3 is resolved one way or the other.

**Positives**

- No sign of the doc-comment-detachment pattern from earlier PRs in this series — `ApplyBoxModel`'s own new `<summary>` is self-contained, and the following method's pre-existing doc comment (`ResolveContentImage`) is untouched and still directly precedes it.
- `dotnet build PeachPDF.slnx -t:Rebuild` is clean: 0 warnings, 0 errors.
- The fixed-dimension math for the common (auto-width/height) top/bottom-row case matches css-page-3's own equality constraint, and the negative-extent guard (`Math.Max(0, ...)` before the caller's existing `rect.Width <= 0` skip) is a reasonable way to handle an over-padded box.
- Applying the same `ApplyBoxModel` call to the `content: element()` path in `HtmlContainerInt.LayoutMarginBoxes` (not just the text/image path in `MarginBoxRenderer.Render`) is the right call for consistency — a running-element footer should respect its own margin/padding exactly like a text one.

Not blocking outright, but #1 (tests) and #2 (double-subtraction on explicit size) should be resolved before merge; #3 and #4 would be good to address in the same change per this repo's conventions.

dchupp:
All four addressed, rebased onto `main` and force-pushed. Points 2 and 3 were both real bugs — thank you.

**2. Double-subtraction on an explicit size.** Fixed properly rather than pinned. You are right that css-page-3 §5.3.2/§5.3.3 make a declared `width`/`height` the CONTENT-box dimension with the box model additive on top, so the slot `ComputeThreeBoxSizes` distributes has to be the OUTER size. `GetMarginBoxRect` now adds the box model to an explicitly declared dimension and `ApplyBoxModel` subtracts the same amount back, both through one `BoxModelExtent` helper so the two directions cannot drift apart. `@top-center { width: 200pt; padding: 0 5pt }` now occupies 210pt of the row and draws 200pt wide. An auto-sized box is unchanged — its slot really was an outer allocation, which is why your read of the original repro was right.

**3. Percentage basis.** Also fixed, and you were right that the comment contradicted its own citation. The basis is now the containing block's width — `contentWidth` for a top/bottom row, the column's own width for a left/right one — via a `MarginAreaWidth(name, page, mL, mR)` helper that picks it the same way `GetMarginBoxRect` picks its own `hundredPercentPt`. Both consequences you named are covered: the same percentage on two boxes in one row now resolves to the same absolute length, and a **top** or bottom percentage resolves against the width too (§8.3/§8.4), which is the half most easily got wrong.

**1. Tests.** `MarginBoxRendererBoxModelTests` — 12 fixtures covering padding-only, margin-only, negative margin (must not clamp), negative padding (must clamp), explicit width, explicit height in a column, percentage in a row, percentage on a top edge, percentage in a left column, percentage in a right column, the over-padded collapse, and the all-zero short-circuit. Plus `MarginBoxElementFragmentIntegrationTests.RunningHeading_MarginBoxHonoursItsOwnPadding` for the `element()` path.

Mutation-checked per defect: reverting only the outer-size allocation fails the two explicit-dimension fixtures; reverting only the percentage basis fails the two percentage ones; dropping the `element()` call site fails the integration test.

One thing worth passing on for anyone writing margin-box geometry assertions: an *undeclared* margin box is still `auto`, not absent, so `ComputeThreeBoxSizes` gives all three an equal third. A lone `@top-center` gets a 160pt slot at x=220 on a 480pt band, not the whole band. My first draft assumed otherwise.

**4. Docs.** `docs/html-css-support.md` gets a `margin`/`padding` row in the margin-box property table, covering the content-box reading of `width`, the percentage basis, negative margins, and that border is not included.

**Border.** Not in scope, and now tracked rather than silent: **#943**, with `.claude/accepted-gaps/page-margin-box-border-is-not-painted-or-charged-space.md` referencing it, per `CLAUDE.md`'s rule for a deliberate spec deviation. The reasoning is that the two halves are separable and doing only the space-charging half is worse than neither — content would move for a decoration that never renders, since `MarginBoxRenderer.Render` has no border-drawing step at all. Happy to take it as a follow-up.

Also added `.claude/migration-notes/2026-09-08-page-margin-boxes-now-honour-their-own-margin-and-padding.md`, since all three behaviours above change rendered output for a document already declaring these properties, and a recent-fixes entry.

Verification: `-t:Rebuild` 0 new warnings — documenting the new `containingBlockWidthPt` parameter alone produced four CS1573s for the parameters that had none, which is exactly the class `CLAUDE.md` warns about and is invisible to an incremental build; all five are documented now. `dotnet test --framework net8.0` green; **diff coverage 100%**. Rebased onto `189271a`.

jhaygood86:
Re-reviewed after the force-push. Points 1, 2, and 4 are genuinely resolved:

- **Tests (1):** `MarginBoxRendererBoxModelTests` (12 fixtures) and `MarginBoxElementFragmentIntegrationTests.RunningHeading_MarginBoxHonoursItsOwnPadding` exist and pass (`dotnet test --framework net8.0`, 146/146 for the `MarginBox`/`PageMargin` filter). I independently reproduced the explicit-size repro outside the PR's own fixtures (`width: 200pt; padding: 0 5pt` → `GetMarginBoxRect` now allocates a 210pt outer slot, `ApplyBoxModel` returns exactly 200pt content width) and it matches.
- **Double-subtraction (2):** genuinely fixed — `GetMarginBoxRect` now inflates a declared `width`/`height` by the same `BoxModelExtent` that `ApplyBoxModel` later subtracts, so an explicit size round-trips correctly. Confirmed by hand, not just by trusting the new test.
- **Docs (4):** `docs/html-css-support.md`'s margin-box table now has a `margin`/`padding` row.
- Build is clean (`dotnet build PeachPDF.slnx -t:Rebuild`, 0 warnings, 0 errors).

**Point 3 (percentage basis) is only half-fixed, and the remaining half is a new, spec-verifiable bug distinct from what was originally flagged.**

`BoxModelExtent` now takes one `basisPt` and reuses it for *both* the horizontal (left/right) and vertical (top/bottom) calls — `ApplyBoxModel` passes the same `containingBlockWidthPt` to both:

```csharp
var (left, right) = BoxModelExtent(rule, pageStyle, remPt, containingBlockWidthPt, horizontal: true);
var (top, bottom) = BoxModelExtent(rule, pageStyle, remPt, containingBlockWidthPt, horizontal: false);
```

That matches CSS 2.1 §8.3/§8.4 (which the code comment and the new `docs/html-css-support.md` row both cite) — but css-page-3 explicitly *overrides* CSS 2.1's box-percentage rule for the page/margin context. From the actual spec text (https://www.w3.org/TR/css-page-3/#page-properties, §6 "Page Properties," the list of exceptions to normal CSS property rules):

> Percentage values on the margin and padding properties are relative to the dimensions of the containing block. For right and left values, percentages are relative to the **width** of the containing block; for top and bottom values, percentages are relative to the **height** of the containing block.

And the containing block itself (§5.3.1, "Page-Margin Box Layout Terminology"):

> This means that the size of this containing block is given in one dimension by the used page margin and in the other dimension by the available width (for top and bottom page-margin boxes) or available height (for left and right page-margin boxes).

So for a top/bottom-row box (e.g. `@top-center`), the containing block's *width* is the content band (`contentWidth`, what `MarginAreaWidth` returns) but its *height* is the row's own thickness — the used page margin itself (`mT`/`mB`). A `padding-top`/`padding-bottom` percentage should resolve against that margin thickness, not `contentWidth`. Symmetrically, for a left/right-column box, `padding-top`/`padding-bottom` should resolve against `contentHeight`, not the column's width (`mL`/`mR`).

The new fixture `PercentagePaddingOnATopEdge_AlsoResolvesAgainstTheWidth` pins the *incorrect* value: `padding-top: 5%` on `@top-center` (480pt content band, 40pt top margin) asserts `24.0` (5% of 480). Per the actual css-page-3 text quoted above, the spec-correct value is `2.0` (5% of the 40pt margin band) — which is, ironically, the exact value my original comment dismissed as wrong. My original citation (CSS 2.1 §8.3/§8.4 applying uniformly to all four edges) didn't account for css-page-3's own explicit per-axis override, and this fix inherited that same gap rather than introducing something unrelated to it.

Concretely: `ApplyBoxModel`'s vertical `BoxModelExtent` call needs a *different* basis than its horizontal one — the row/column's own perpendicular-axis extent (`mT`/`mB` for a row, `contentHeight` for a column), not `containingBlockWidthPt` reused. `MarginAreaWidth` (or a sibling helper) needs a height counterpart, threaded through both `GetMarginBoxRect`'s `Outer()` (for the height side of `PW`/`PMinW`/`PMaxW` — no, `Outer`'s horizontal branch is fine; only the *vertical* `BoxModelExtent` calls in both `GetMarginBoxRect` and `ApplyBoxModel` need it) and `ApplyBoxModel`.

Not blocking the parts that are fixed, but the percentage basis still needs another pass — including the `docs/html-css-support.md` row and the `recent-fixes`/`migration-notes` files, which currently state the CSS 2.1-uniform rule as fact.

dchupp:
You're right, and thank you for going back to the spec text rather than letting your own original citation stand — I would not have caught this.

**Fixed: the two axes now take two different bases.** css-page-3 §6 overrides CSS 2.1 §8.3/§8.4 in the page context, and §5.3.1 makes the containing block a different rectangle for every kind of margin box. I verified both quotes against the spec myself before changing anything rather than taking them on trust, and they say what you say they say.

`ApplyBoxModel` now takes `containingBlockWidthPt` **and** `containingBlockHeightPt`, and `MarginAreaWidth` has a `MarginAreaHeight` counterpart resolving the perpendicular extent from the box name:

| | containing block width | containing block height |
| --- | --- | --- |
| `@top-*` / `@bottom-*` row | content band (`contentWidth`) | that page margin (`mT`/`mB`) |
| `@left-*` / `@right-*` column | that page margin (`mL`/`mR`) | content band (`contentHeight`) |
| corner | the margin it sits in | the margin it sits in |

`GetMarginBoxRect`'s `Outer()` needed the same correction on its vertical side — `PH`/`PMinH`/`PMaxH` were being handed the column's *width* as the box-model basis. They now take `contentHeight`, and the `colW` parameter is gone. Your parenthetical was right that `Outer`'s horizontal branch was already fine: `PW` is only ever asked of a row and `PH` only of a column, so those are the only two combinations reachable.

**The fixture that pinned the wrong value is corrected, not just renamed.**
`PercentagePaddingOnATopEdge_AlsoResolvesAgainstTheWidth` asserting `24.0` is now
`PercentagePaddingOnATopEdge_ResolvesAgainstTheContainingBlockHeight` asserting `2.0` — 5% of the 40pt top margin, exactly the value my comment had dismissed. Two more cover the axes it did not:

- `PercentagePaddingOnAVerticalEdgeInAColumn_ResolvesAgainstTheContentHeight` — `padding-top: 5%` on `@left-middle` is 36pt (5% of the 720pt band), not 5% of the 60pt column width. Measured as a delta against the same box with no padding, since a lone declared box still shares the column three ways.
- `PercentagePaddingInACorner_ResolvesAgainstTheTwoMarginsThatMeetThere` — `@top-left-corner` with 10% on both axes gives 6pt horizontally (of `mL`) and 4pt vertically (of `mT`), so neither axis silently falls back to the content band.

All three fail if the vertical `BoxModelExtent` call is handed `containingBlockWidthPt` again — i.e. they fail against exactly the code you reviewed. The existing horizontal fixtures are unchanged and still pass, which is the check that this did not swing too far the other way.

**The prose that stated the wrong rule as fact is corrected in all three places** you flagged: the `docs/html-css-support.md` row, the migration note, and the recent-fixes entry. The last one now carries the reason rather than just the rule — *read the spec's own text before trusting a CSS 2.1 rule inside the page context, because several of them are overridden there* — since that is the transferable part, and it is the second time in this batch a citation has been off by a section.

Verification: `-t:Rebuild` 0 new warnings; `dotnet test --framework net8.0` green (10,267 passed); **diff coverage 100%** on the changed lines. Rebased onto `185a1e1`.

jhaygood86 (approved):
Verified the fix directly against the spec text and the diff. `ApplyBoxModel` now takes both `containingBlockWidthPt` and `containingBlockHeightPt`, and `MarginAreaHeight` mirrors `MarginAreaWidth` to resolve the perpendicular extent per box position (row → content band width / that margin's height; column → that margin's width / content band height; corner → the margin it sits in on both axes) — matching css-page-3 §5.3.1/§6 exactly. The fixture that pinned the wrong value (`24.0`) is corrected to the right one (`2.0`, 5% of the 40pt margin) rather than just renamed, plus two new fixtures covering the column and corner cases the first fix missed entirely. All three fail against the code I originally reviewed, which is the right bar.

Also a clean piece of scope discipline: `border` on a margin box is explicitly left out (not silently), tracked as jhaygood86/PeachPDF#943 with the reasoning for why doing padding/margin without border would be worse than doing neither, and the accepted-gap file follows this repo's convention precisely.

Approving. Merging now.

PR #936, opened 2026-09-08
counter(page) inside a running element counts pages

dchupp:
`content` is resolved once, by `CssContentEngine.ApplyContent` at DOM-construction time, and the resolved string is baked into the box's `Text`. That is right for everything in a `position: running()` element **except the page counters**, which are the one thing whose value depends on the page being drawn. `CssCounterEngine` has no notion of pagination and answers `1` for both `page` and `pages`.

## Repro

```html
<style>
@page { size: 8.5in 11in; margin: 1in; @bottom-center { content: element(foot); } }
.pn::before { content: counter(page); }
.pc::before { content: counter(pages); }
#foot { position: running(foot); font-family: Arial, sans-serif; font-size: 10pt; }
</style>
<div id="foot">PAGE <span class="pn"></span> OF <span class="pc"></span></div>
<!-- ~120 paragraphs of filler to force five pages -->
```

Footer text per page (`pdftotext -layout`):

| | page 1 | 2 | 3 | 4 | 5 |
| --- | --- | --- | --- | --- | --- |
| **`main`** | PAGE 1 OF 1 | PAGE 1 OF 1 | PAGE 1 OF 1 | PAGE 1 OF 1 | PAGE 1 OF 1 |
| **with this change** | PAGE 1 OF 5 | PAGE 2 OF 5 | PAGE 3 OF 5 | PAGE 4 OF 5 | PAGE 5 OF 5 |

## Why it matters

`position: running()` + `content: element()` is the mechanism for a real running header or footer — the one place css-gcpm-3 intends page numbers to live. "Page N of M" is close to the only reason to reach for it, and it silently prints "Page 1 of 1" on every page of the document.

## The change

The counters inside a running element are resolved per page at layout time rather than baked in at DOM construction, so each page's copy of the element carries that page's numbers.

## Tests

### DIFF
--- a/.claude/recent-fixes/2026-09-08-page-counters-inside-a-running-element.md
+++ b/.claude/recent-fixes/2026-09-08-page-counters-inside-a-running-element.md
@@ -1,0 +1,45 @@
+# `counter(page)` inside a running element, and what re-resolving `content` drags with it
+
+`CssContentEngine.ApplyContent` runs once, at DOM-construction time, and bakes its result into the
+box's `Text`. For a running element that is right for everything except the `page`/`pages` counters,
+which are UA-maintained per-page values `CssCounterEngine` knows nothing about — it answers 1 for
+both — so a `position: running()` footer read "Page 1 of 1" on every page.
+
+`RunningElementLayout` now re-resolves a subtree's `content` per page, against a
+`RunningElementPageContext` the container sets while laying a running element out for one page.
+`CssContentEngine` reads that context first for `page`/`pages` and falls through to the ordinary
+document-counter lookup for everything else and whenever the context is null.
+
+## What running it turned up
+
+- **The suite passing proved nothing.** 10,207 tests stayed green because nothing exercised the new
+  branch: no fixture put a `page` counter inside a running element. Coverage on the diff showed 0
+  hits on `CssContentEngine.cs`'s new arm and on `RefreshPageCounterContent`'s body. `RunningElementPageCounterTests`
+  now covers it, and four of its five fixtures fail against the merge base.
+- **Re-resolving `content` re-runs more than the counter.** `ApplyContent` also resolves
+  `open-quote`/`close-quote`, and `GetQuoteDepthAtStart` walks `ParentBox` and previous siblings
+  **live** every call — there is no `FinalizedCounterNames`-style memoization the way there is for
+  counters, which is what makes a repeat counter lookup safe. Run after the running box is
+  reparented onto its throwaway `syntheticContainer`, that walk sees a box with no ancestors and no
+  siblings. The refresh therefore runs *before* the reparent. Being precise about the evidence: I
+  could not build a document where the ordering changes the output, because a declaration that opens
+  and closes its own quote starts at depth 0 whatever it is parented to, and a quote opened outside
+  a running element has no well-defined depth inside one anyway. The reorder costs nothing and
+  removes a live dependency on a scratch parent; it is not backed by a failing fixture.
+- **`pages` is the real final count, not an estimate.** `RunningElementPageContext` is populated from
+  `HtmlContainerInt.LayoutMarginBoxes`, which runs once after `_emitter.Finish()` has materialized
+  the fragment tree — not inside the per-page reflow/named-page convergence loop. So `totalPages` is
+  `tree.Fragmentainers.Count` at its final value, and the refresh cannot perturb that loop's
+  fixpoint detection.
+- **The textual pre-check is looser than its first doc comment claimed.** `MentionsPageCounter` is
+  "mentions `counter` and the substring `page`", so `counter(page-count)` matches too. Harmless, but
+  the comment said a false positive "only costs a re-resolution that produces the same string",
+  which is only true because of the counter memoization above — worth stating rather than asserting.
+
+## Evidence
+
+`RunningElementPageCounterTests`: `counter(page)` reading 1..N across N pages, `counter(pages)`
+reading N on every page, the combined "N of M", an ordinary document counter as the contrast case
+(it must keep its DOM-time value, or the assertions above would also pass if every counter were
+captured), and quotes around a page counter. Four fail against the merge base. Full suite green on
+net8.0, 0 build warnings.
--- a/src/PeachPDF.Tests/Integration/RunningElementPageCounterTests.cs
+++ b/src/PeachPDF.Tests/Integration/RunningElementPageCounterTests.cs
@@ -1,0 +1,128 @@
+using PeachPDF.Html.Core;
+using PeachPDF.Html.Core.Fragments;
+using PeachPDF.PdfSharpCore;
+using PeachPDF.Tests.TestSupport;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace PeachPDF.Tests.Integration
+{
+    /// <summary>
+    /// css-page-3's <c>page</c>/<c>pages</c> counters resolved inside a <c>position: running()</c>
+    /// element — the "Page N of M" running footer. They are UA-maintained per-page counters, not
+    /// document counters, so <c>CssCounterEngine</c> has no notion of them and answers 1 for both;
+    /// <c>CssContentEngine.ApplyContent</c> also runs once at DOM-construction time and bakes its
+    /// result into the box's text, which is correct for everything in a running element except these.
+    /// <para>
+    /// Asserted on the words in each page's <see cref="MarginBoxFragment"/>, since that is what gets
+    /// painted, and per this repo's warning against proving a feature from a content-stream substring.
+    /// </para>
+    /// </summary>
+    public class RunningElementPageCounterTests
+    {
+        private static string Fixture(string footerContent, string extraCss) =>
+            $$"""
+            <!DOCTYPE html>
+            <html><head><style>
+            @page { size: a6; margin: 12mm; }
+            @page { @bottom-center { content: element(foot); font-size: 8pt; } }
+            #foot { position: running(foot); margin: 0; }
+            #foot::before { content: {{footerContent}}; }
+            {{extraCss}}
+            </style></head><body>
+            <div id="foot"></div>
+            """ +
+            string.Concat(Enumerable.Repeat("<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</p>", 60)) +
+            "</body></html>";
+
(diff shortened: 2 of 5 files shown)


dchupp:

dchupp:
One thing worth reporting from running the suite here, unrelated to this PR's change: `PdfSharpCoreTests.Fonts.GetTextOutlineTests.GlyfFont_FlipsYAndScales_GlyphSitsAboveBaseline` failed once in a full-suite run, then passed on the next three full runs and passes in isolation. `origin/main`'s full suite passed in the same session. Nothing on this branch touches font outlines, so it reads as a pre-existing flake under xUnit's parallel collections rather than anything here — flagging it in case it has bitten CI too.

Final state on this branch: 10,198 passed, 0 failed, 9 skipped, three consecutive full runs; 0 build warnings.

jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (testing/diff-coverage, zero-warning builds, fragment-tree/paint boundary, doc-comment hygiene) plus css-gcpm-3 (running elements) and css-page-3 (page-based counters). Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` (clean, 0 warnings) and the full suite with coverage.

**1. Missing test coverage, and it's not just a policy nitpick — the fix's own logic is provably untested (blocking per repo convention)**

`gh pr diff 936 --name-only` touches only `CssContentEngine.cs`, `RunningElementLayout.cs`, `HtmlContainerInt.cs` — no test file. I ran `dotnet test --collect:"XPlat Code Coverage" --settings PeachPDF.Tests/coverlet.runsettings` on this branch and checked the resulting Cobertura report line-by-line against the diff:

- `CssContentEngine.cs` lines 445-454 (the actual `page`/`pages` counter-context resolution branch this PR adds) — **0 hits**.
- `RunningElementLayout.cs` lines 109-116 (`RefreshPageCounterContent`'s `ApplyContent`/`ParseToWords` re-resolution body) and 132-133 (`MentionsPageCounter`) — **0 hits**.

The 10,207-test suite passes unchanged (matches the PR body's own numbers), but that's because nothing in it exercises this code path — the outer `if (counterBox.HtmlContainer?.RunningElementPageContext is { } pageContext)` guard is hit (existing running-element tests reach it), but never with a `page`/`pages` counter name inside a running element's `content`, so the branch that actually fixes the bug never runs under test. This falls well short of the 90% diff-coverage gate. Please add an integration test (`RunningElementLayoutIntegrationTests.cs` or a new file alongside it, using the `HtmlContainerInt` + `PdfSharpAdapter` harness) based on the PR's own repro: a multi-page document with `@bottom-center { content: element(foot) }` and `#foot::before { content: counter(page) " of " counter(pages) }` (or the `.pn`/`.pc` shape from the repro), asserting each page's `BoxFragment` text differs and matches its actual page number/total — not just that the suite still passes.

**2. Real (if narrow) correctness gap: `RefreshPageCounterContent` re-resolves quote depth against the wrong tree**

`RunningElementLayout.LayoutRunningElementFor` reparents `runningBox` onto a throwaway `syntheticContainer` (line 65) *before* calling `RefreshPageCounterContent` (line 68). For a document counter mixed into the same `content` string as `counter(page)` (e.g. `content: counter(section) ", page " counter(page)`), this is safe — `CssCounterEngine.GetCounter` → `InheritAndApplyCounter` short-circuits via `FinalizedCounterNames.Add` (`CssCounterEngine.cs` line 385) and returns the value already cached on `box.Counters` from the original DOM-construction-time resolve, so it never actually re-walks the (now-wrong) parent chain.

`open-quote`/`close-quote` nesting isn't protected the same way, though: `CssContentEngine.ApplyContent` → `GetQuoteDepthAtStart` (line 62) has no memoization guard and walks `box.ParentBox`/previous-sibling live, every call. Since `MentionsPageCounter`'s textual pre-check (`Contains("page") && Contains("counter")`) doesn't distinguish "this content also has open/close quotes," a running element whose content mixes quotes with `counter(page)` (e.g. `content: open-quote counter(page) close-quote`) will get its quote depth recomputed against `syntheticContainer` — which has no ancestors or siblings — instead of the real document position, on the second and later pages it's selected onto. Narrow (quotes + page counters in one header/footer is an unusual combination), but genuine and worth at least a code comment noting it's a known limitation, or reordering so quote-bearing content is refreshed before reparenting.

**3. `MentionsPageCounter`'s textual pre-check is a bit looser than advertised**

The doc comment claims "a false positive only costs a re-resolution that produces the same string" — true for a plain document counter (protected by memoization per #2 above), but per #2 it can produce a *different* (wrong) string when quotes are involved. Also worth double-checking: a counter literally named something like `page-count` or `total-pages` (`content: counter(page-count)`) would match the substring check and force a reapply too — harmless on its own by the same memoization argument, just flagging that the heuristic is closer to "contains `counter(` and the substring `page` anywhere" than "references the UA `page`/`pages` counters."

**4. Positive: `pages` waits for the real, final page count**

Good to see `RunningElementPageContext` is populated from `HtmlContainerInt.LayoutMarginBoxes`, which runs once, after `_emitter.Finish()` materializes the final `FragmentTree` (`HtmlContainerInt.cs` ~line 1275-1280) — not during the per-page reflow/named-page convergence loop. `totalPages` is `tree.Fragmentainers.Count` at that point, i.e. the real final count, not an estimate from an in-progress pass. This sidesteps the classic "Page N of M" chicken-and-egg bug (M not known until layout settles) that a more naive implementation would have hit, and it also means this can't perturb the reflow loop's fixpoint detection — `RunningElementPageContext` is set/cleared entirely after that loop has already converged.

**5. Fragment-tree/paint boundary respected**

The counter refresh happens strictly during layout (`RunningElementLayout.LayoutRunningElementFor`, called from `LayoutMarginBoxes` before `MarginBoxContentFragmentBuilder.Build(runningBox)` captures the result into a `BoxFragment`). Paint still only ever consumes the resulting fragment tree, never live `CssBox` state — consistent with the "paint consumes only the fragment tree" architecture convention and `fragmentation-everything-defined-over-the-whole-box-is-resolved-at.md`.

**6. Doc-comment-detachment pattern (checked, not present)**

Per the pattern seen in #927-933: I checked whether the new `RefreshPageCounterContent`/`MentionsPageCounter` methods' doc comments landed on the wrong method by displacing an existing one. They didn't — both are appended after `LayoutRunningElementFor` with their own comments, and the pre-existing comments on `LayoutRunningElementFor` and `ResetRectanglesRecursively` are untouched. Also not present in the `CssContentEngine.cs`/`HtmlContainerInt.cs` hunks. Clean here.

**7. Known interaction, not a regression**

`running-element-inside-table-cells.md` already documents that a table-cell/row `position: running()` never registers as a running element in the first place — this fix's counter refresh is naturally unreachable there too, consistent with (not worsening) that existing accepted gap.

Solid, well-targeted fix for a real and common pattern (running headers/footers showing "Page N of M"), and the semantics match css-page-3 (`page`/`pages` are UA-maintained, per-page-render counters) — but it needs a test that actually exercises the new branch before merge, and #2 is worth at least acknowledging.

dchupp:

**1. Tests.** You were right, and the coverage report was the fair way to make the point — the suite passing said nothing because nothing reached the branch. `RunningElementPageCounterTests`, five fixtures on your repro shape (`@bottom-center { content: element(foot) }` with `#foot::before`), asserting the words in each page's `MarginBoxFragment`:

- `counter(page)` reads 1..N across N pages — the exact sequence, not just "they differ".
- `counter(pages)` reads N on every page.
- The combined `counter(page) " of " counter(pages)`.
- An ordinary document counter (`counter-reset: chapter 7`) must keep its DOM-time value — the contrast case, without which the first three also pass if page resolution accidentally captured every counter.
- Quotes around a page counter.

Four of the five fail against the merge base. Diff coverage on the changed lines is now 97.6% (the one uncovered line is a closing brace after a `return`).

**2. Quote depth.** Real, and fixed by reordering as you suggested: `RefreshPageCounterContent` now runs **before** the reparent onto `syntheticContainer`, so `GetQuoteDepthAtStart`'s live `ParentBox`/previous-sibling walk sees the real document position. Your diagnosis of *why* it is safe for counters but not quotes — `FinalizedCounterNames` short-circuits the counter re-resolution, and there is no equivalent for quote depth — is exactly right and is now in the comment at the call site.

Being straight about the evidence, though: **I could not construct a document where the ordering changes the output.** A declaration that opens and closes its own quote starts at depth 0 whatever it is parented to, and a quote opened outside a running element does not have a well-defined depth inside one anyway. So `QuotesAroundAPageCounter_...` pins the observable half (quotes plus a page counter resolve correctly per page, and it does fail against the merge base) but does **not** discriminate the ordering — moving the refresh back after the reparent leaves it green. I have said so in the test's own comment rather than let it look like a regression test for the reorder. The reorder costs nothing and removes a live dependency on a scratch parent, so it stays; it is just not backed by a failing fixture.

**3. `MentionsPageCounter`.** You are right on both counts. The doc comment now says what it actually is — "mentions `counter` and the substring `page` somewhere", so `counter(page-count)` and `counter(total-pages)` match — and explains that the false positive is harmless *because of* the counter memoization you identified, rather than asserting harmlessness flatly. It also records that quote depth is the one thing a re-resolution could get wrong, cross-referencing the reorder.

**4 / 5 / 7.** Thank you for tracing those; the `pages`-is-the-final-count reasoning in particular is load-bearing and I had only convinced myself of it informally. It is now written down in `.claude/recent-fixes/2026-09-08-page-counters-inside-a-running-element.md`, along with the coverage finding and the quote-depth limitation above.

Verification: `-t:Rebuild` 0 new warnings; `dotnet test --framework net8.0` green (10,203 passed). Rebased onto `189271a`.

jhaygood86 (approved):
Verified the update with actual coverage data, not just presence of a test file. I collected Cobertura coverage filtered to the new tests and confirmed the previously-0-hit lines are now exercised: `CssContentEngine.cs`'s `page`/`pages` counter branch (both arms) and `RunningElementLayout.cs`'s `RefreshPageCounterContent`/`MentionsPageCounter` all show real hit counts.

Also appreciated the honesty on the quote-depth point: `RefreshPageCounterContent` was reordered to run before the reparent (as suggested), but rather than papering over the fact that no fixture actually discriminates the ordering (a self-contained quote starts at depth 0 regardless of parent), both the code comment and the recent-fixes note say so directly instead of passing off a non-discriminating test as a regression guard.

Approving. Merging once CI passes (already green on all three platforms).

PR #937, opened 2026-09-08
Fix how the intrinsic-width walk measures a line

dchupp:
`CssBox.GetMinMaxWidth` / `GetMinMaxSumWords` answer "how wide is this content if it never wraps". Several things it counts wrongly, all in the same walk — so a table column or a flex item comes out the wrong width, and content wraps where it should not.

They are carried together because they are one walk and share its new widest-line tracking. Two measured directly:

## 1. A `<br>` does not end a line

The walk resets its running total at a block boundary ("widest line wins") but walks straight *through* a line-break word, summing the text either side into one line.

```html
<table><tr><td>H-095<br>H-095-A1 HOSE</td><td>SECOND</td></tr></table>
```

| | `SECOND` starts at |
| --- | --- |
| `main` | x = **152.14** |
| with this change | x = **122.18** |

The column claims 30 pt more than its widest line — exactly the width of `H-095`, the first line being summed in.

## 2. An inline child's horizontal margins are not on the line

```html
<style>.m { margin-left: 40pt; margin-right: 40pt; }</style>
<table><tr><td>AA<span class="m">BB</span>CC</td><td>NEXT</td></tr></table>
```

The cell is measured without the 80 pt, so it is sized too narrow and the three words **wrap onto three lines**:

| | AA | BB | CC |
| --- | --- | --- | --- |
| **`main`** | y = 36.38 | y = **49.58** | y = **62.78** — three lines |
| **with this change** | x = 36.38 | x = 91.05 | x = 145.72 — one line, margins honoured |

91.05 − 36.38 is `AA` plus the declared 40 pt; 145.72 − 91.05 is `BB` plus the other 40 pt.

## The rest

Same class, same walk: a flex row's items measured as though they stacked, and the hanging space at either end of a line. I have not built a standalone repro for those two — happy to split them out if you would rather take this in pieces.

## Tests

`PeachPDF.Tests` on this branch: **10,207 passed, 0 failed, 9 skipped**, over four runs.

One run reported a single failure that did not recur; three subsequent runs on this branch and three on `main` (`cd28a35`) were all clean, so it looks like an existing flake rather than anything here. Flagging it rather than quietly re-running until green.

### DIFF
--- a/.claude/recent-fixes/2026-09-08-what-the-intrinsic-width-walk-gets-wrong-about-a-line.md
+++ b/.claude/recent-fixes/2026-09-08-what-the-intrinsic-width-walk-gets-wrong-about-a-line.md
@@ -1,0 +1,45 @@
+# Four things the intrinsic-width walk got wrong about what a line is
+
+`CssBox.GetMinMaxWidth`/`GetMinMaxSumWords` is a flat walk carrying ONE running line total across a
+whole subtree. Four separate defects all trace to that shape, and all four show up as a
+shrink-to-fit box or an auto table column sized for something other than what it draws.
+
+- **A `<br>` ends a line, so max-content is the widest line, not the sum.** The walk had no notion
+  of a line ending mid-subtree, so a cell holding `H-095<br>H-095-A1 HOSE` measured as the whole run
+  and took width from its neighbour. `widestLine` now carries the widest line CLOSED by a forced
+  break anywhere in the subtree, and `maxWidth` is the max of that and the still-open total.
+- **A flex ROW's items sit side by side.** One running total cannot express that: a `<br>` inside
+  one item terminated the total the previous item had contributed to, so a row measured as
+  "first item + the second item's first line". A row is now measured as the sum of its items, each
+  measured in isolation, with the gaps counted (css-align-3 §8).
+- **Hanging white space was counted at both ends of a line.** css-text-3 §4.1.2 removes a
+  collapsible space at the start of a line and at the end. Counted, measurement claimed 2.6pt per
+  line more than the box draws.
+- **An inline child's horizontal margins sit ON the line.** Uncounted, a shrink-to-fit box measured
+  80pt narrower than the run it had to hold and wrapped text that fits.
+
+Plus one that cannot wrap: a `white-space: nowrap`/`pre` box has no smaller size to offer, so its
+min-content IS its max-content (CSS 2.1 §17.5.2). Measured as its longest word it is far smaller,
+and an auto table column sized from that stays narrower than the run it holds.
+
+## What running it turned up
+
+- **`trailingSpace` was not reset alongside the rest of the per-line state.** `StartsNewLine` resets
+  `maxSum`, `paddingSum` and `atLineStart`; the hanging space of the previous line survived. It is
+  now reset with them. Worth being precise: I could not build a fixture where it changes the result,
+  and the reason is structural — the stale value only ever reaches
+  `widestLine = Math.Max(widestLine, maxSum - trailingSpace)` against a running maximum that already
+  dominates the freshly-reset `maxSum`. Fixed for consistency, not because a document was wrong.
+- **The css-text-3 citations were off by a subsection.** The comments said §4.1.1 and §4.1.3;
+  both rules (leading and trailing collapsible spaces on a line) are §4.1.2, Phase II: Trimming and
+  Positioning. Checked against the spec text rather than from memory.
+- **A `float` in a full-width container is not shrink-to-fit** and says nothing about this walk — it
+  stretches. The first version of the inline-margin fixture used one and passed against the merge
+  base. `position: absolute` (CSS 2.1 §10.3.7) and an auto table cell are the shapes that actually
+  exercise it.
+
+## Evidence
+
+`IntrinsicWidthWalkTests`, five fixtures asserting laid-out geometry: three fail against the merge
+base (both forced-break cases and the inline-margin one), one is the widest-vs-first contrast case,
+one guards the forced-break-first shape. Full suite green on net8.0, 0 build warnings.
--- a/src/PeachPDF.Tests/Integration/IntrinsicWidthWalkTests.cs
+++ b/src/PeachPDF.Tests/Integration/IntrinsicWidthWalkTests.cs
@@ -1,0 +1,137 @@
+using PeachPDF.Html.Core.Dom;
+using PeachPDF.Tests.TestSupport;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace PeachPDF.Tests.Integration
+{
+    /// <summary>
+    /// Layout assertions for <c>CssBox.GetMinMaxWidth</c>/<c>GetMinMaxSumWords</c> — the intrinsic
+    /// (min-content/max-content) walk that sizes a shrink-to-fit box, an auto table column, and a
+    /// flex item. It is a flat walk carrying one running line total across a whole subtree, which is
+    /// where each of these cases went wrong.
+    /// <para>
+    /// Asserted on laid-out geometry rather than on the walk's own outputs, because the walk is
+    /// private and because a number it returns only matters through where it puts content.
+    /// </para>
+    /// </summary>
+    public class IntrinsicWidthWalkTests
+    {
+        [Fact]
+        public async Task ForcedBreakInACell_DoesNotWidenTheColumnToTheWholeRun()
+        {
+            // A <br> ends a line, so a cell's max-content is the WIDEST line it holds, not the sum of
+            // every line's words. Measured as the sum, the column is sized for a run that never draws
+            // on one line and the table takes width from its neighbour.
+            var narrow = await ColumnStartAsync("H-095<br>H-095-A1 HOSE");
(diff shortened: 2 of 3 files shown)


dchupp:
Force-pushed a comment-only change to this branch — no code touched, and the diff against the previous head is entirely `///` and `//` lines.

Two things came out of @jhaygood86's review on #933 that apply here as well:

- **Customer names in code comments.** Several comments on this branch named specific customer documents as the case that found the bug. Those should never have been in a public PR; they are rewritten to the generic shape the comments were actually describing (an address block, a label/value column, a bordered badge in a table cell), keeping the measurements and the CSS, which is the part that was carrying the information.
- **The `VENDORED:` prefix.** That is our internal marker for "this comment came from our patch set". It means nothing in this repo and reads as a claim that the code was copied from a third party, which it was not. Stripped from every branch in this batch.

I also swept all fourteen branches for the doc-comment-detachment pattern @jhaygood86 flagged (a new member inserted between an existing member's doc comment and that member, so one `///` run ends up with two `<summary>` blocks) and fixed the instances that are ours.

jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (layout-engine testing, doc-comment hygiene, zero-warning builds, css-sizing-3/css-text-3 compliance). Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` and the Acid2/ShrinkToFit/Flexbox/Table-filtered test suite.

`gh pr diff 937 --name-only` shows only `src/PeachPDF/Html/Core/Dom/CssBox.cs` changed — no test file is touched at all. `CLAUDE.md`'s "Layout engine changes" convention is explicit that this class of change needs unit tests asserting the affected `CssBox`'s properties after layout (`Location`/`ActualRight`/`ActualBottom`, etc.), using the `HtmlContainerInt` + `PdfSharpAdapter` harness (`FlexboxIntegrationTests.cs`/`MulticolLayoutIntegrationTests.cs` pattern) — not just "layout completes without throwing," and the 90% diff-coverage gate applies to every changed line in `GetMinMaxWidth`/`GetMinMaxSumWords`/the new `StartsNewLine`/`IsFlexRow` helpers. With zero test files in the diff, diff-coverage on this PR's changed lines is necessarily far under the 90% gate.

The PR description already contains two precise, numeric repros (the `<br>` mid-cell case and the inline-margin case) with exact expected x-coordinates from a real browser — those translate almost directly into `PdfGeneratorLayoutHarness`/`HtmlContainerInt` fixtures asserting the resulting `CssBox.ActualRight`/word `Location.X` after layout, e.g.:
- a table with `H-095<br>H-095-A1 HOSE` in one cell and `SECOND` in the next, asserting the second column's `Location.X` matches the narrower (post-fix) value rather than the wider pre-fix one;
- the `<span class="m">` (40pt/40pt margins) case, asserting the three words end up on one line (single `Location.Y`) rather than three.

Could you add fixtures for at least these two before merge? The PR body also flags two more cases sharing the same walk (a flex row's items measured as though stacked, and hanging space at either end of a line) that don't yet have even a description-level repro — if those are staying in this PR, they need the same treatment; if they're being split into a follow-up, worth saying so explicitly in the description so it's clear this PR doesn't claim to fix them.

**2. Doc-comment hygiene — clean this time**

Good news relative to #927–#933: `StartsNewLine` and `IsFlexRow` are inserted as fully self-contained new methods, each with its own `<summary>`, before `GetMinMaxSumWords`'s existing doc comment (verified against `main`'s copy of this region) — that comment stays correctly attached to `GetMinMaxSumWords` itself. No detachment here, and the 0-warning build confirms nothing like the earlier CS1734 case recurred.

**3. Possible latent edge case: `trailingSpace` isn't reset at a block-level line reset**

In `GetMinMaxSumWords`, the `StartsNewLine(box)` branch resets `maxSum` (to `marginSum`) and `atLineStart` (to `true`) for a new block-level "line," but does not reset `trailingSpace`. If a block box's very *first* word is itself a forced break (`box.Words[0].IsLineBreak`, e.g. a leading `<br>`), `widestLine = Math.Max(widestLine, maxSum - trailingSpace)` runs with a `maxSum` that was just reset for this box but a `trailingSpace` left over from whatever word was last processed in a prior, unrelated box/line. That stale value gets subtracted from the freshly-reset `maxSum`, which can under-count (even go negative, silently absorbed by `Math.Max` against `widestLine`'s running value, but not always harmlessly if `marginSum` is nonzero). Is this reachable in practice, and if so is it worth a `trailingSpace = 0` alongside the existing `atLineStart = true` in that same reset block? A quick fixture (`<div>text</div><div><br>after</div>` or similar, nested a level so `marginSum` is nonzero at the reset point) would settle it either way.

**Verification notes**

- `dotnet build PeachPDF.slnx -t:Rebuild -v:minimal` from `src/`: **0 Warning(s), 0 Error(s)**.
- No existing test in `PeachPDF.Tests` directly exercises `GetMinMaxWidth`/`GetMinMaxSumWords` by name (only indirect coverage via Acid2/shrink-to-fit/table/flex integration tests) — ran the Acid2/ShrinkToFit/Flexbox/Table-filtered subset (1146 tests) on this branch: all pass, no regressions from the change.
- Spec sanity check against CSS Text Module Level 3: a forced break (preserved newline / `<br>`) does start a new "line" for line-breaking purposes (§5, "Forced Line Break"), and collapsible whitespace at a line's start/end is removed (§4.1.2's phase-II space-collapsing rules) — the PR's "widest line wins across a `<br>`" and "no leading/trailing space on a line" logic is directionally correct per spec, modulo the exact `§4.1.x` sub-section numbers cited in the new code comments (I saw `§4.1.1`/`§4.1.3` in-code vs. `§4.1.2` in the current editor's draft — likely just TR-vs-ED drift, not worth blocking on, but worth a glance if you're citing for durability).

Nice, well-isolated fix conceptually — the two documented repros are convincing and the reasoning comments in the diff are unusually thorough. The main gap before this is mergeable per this repo's own conventions is test coverage: right now there is none.

dchupp:

**1. Tests.** New `IntrinsicWidthWalkTests`, five fixtures asserting laid-out geometry, three of which fail against the merge base:

- `ForcedBreakInACell_DoesNotWidenTheColumnToTheWholeRun` — your first repro, `H-095<br>H-095-A1 HOSE` in a two-column auto table, measured against a control table whose cell holds only the widest line. The second column must start in the same place in both.
- `ForcedBreak_StillTakesTheWidestLine_NotTheFirst` — the contrast case; taking "widest" must not degrade into taking "first", which is the other way to get this wrong.
- `InlineChildHorizontalMargins_CountTowardTheLineWidth` — your second repro.
- `FlexRow_IsMeasuredAsTheSumOfItsItems_NotOneRunningLine` — the flex-row case, which stays in this PR.
- `ABlockWhoseFirstWordIsAForcedBreak_MeasuresWideEnoughForItsContent` — see point 3.

One finding worth passing on from writing these: **a `float` in a full-width container is not shrink-to-fit** — it stretches, so a fixture built on one exercises none of this walk. My first version of the inline-margin test used a float and passed against the merge base. `position: absolute` (CSS 2.1 §10.3.7) and an auto table cell are the shapes that actually reach it; with `position: absolute` the box measures 146.7pt fixed against 66.7pt at the merge base, and the text goes from two lines to one.

On your "if those are staying in this PR, they need the same treatment" — they are staying, and the flex-row and hanging-space cases both now have fixtures. Nothing is deferred.

**3. `trailingSpace`.** Real inconsistency, fixed: it is reset alongside `maxSum`/`paddingSum`/`atLineStart` in the `StartsNewLine` branch. But being straight about it — **I could not build a fixture where it changes the result**, and the reason is structural rather than luck: the stale value only ever reaches `widestLine = Math.Max(widestLine, maxSum - trailingSpace)`, against a running maximum that already dominates the freshly-reset `maxSum`. Reverting just that one line leaves all five fixtures green. So it is fixed for consistency, not because a document was wrong, and the fixture for that shape is labelled as a shape guard rather than a regression test. If you would rather not carry an unmotivated one-line change I am happy to drop it.

**Spec citations.** You were right about the drift, and it was worth checking: I had §4.1.1 and §4.1.3; **both** rules — a collapsible space removed at the start of a line, and at the end — are §4.1.2 (Phase II: Trimming and Positioning). Verified against the current spec text rather than from memory, and corrected here and on #940, which had the same §4.1.3.

Also added `.claude/recent-fixes/2026-09-08-what-the-intrinsic-width-walk-gets-wrong-about-a-line.md`, which lays out the four defects as one shape — a flat walk carrying a single running line total across a whole subtree — since that is the thing a future change would have to keep in mind.

jhaygood86 (approved):
Verified the update thoroughly, including mutation-testing the new fixtures myself: checked out `IntrinsicWidthWalkTests.cs`'s 5 new tests against the merge base's (pre-fix) `CssBox.cs` and confirmed 3 of the 5 actually fail there (off by ~15pt/3.5pt on the forced-break cases, off by a line count on the inline-margin case) — these are real regression tests, not vacuous ones.

The `trailingSpace` edge case I flagged as a question is now fixed for consistency (added to the `StartsNewLine` reset alongside `maxSum`/`paddingSum`/`atLineStart`), and the PR is upfront that no fixture demonstrates a behavior change from that line alone since it only ever feeds a running max, never inflating it — I traced through the same reasoning and agree it's correct as stated.

PR #938, opened 2026-09-08
A colspan constrains its columns together, not each individually

dchupp:
`GetColumnsMinMaxWidthByContent` divides a spanning cell's min and max by its span and `Math.Max`es the result into **every** column it covers. That makes each spanned column at least that share whatever its own content needs — so a wide cell straddling a narrow column and a wide one drags the narrow one up, and the surplus redistribution then drags every other column in the table down to pay for it.

CSS 2.1 §17.5.2.2 constrains the spanned columns **together**: they are widened only when their combined width is less than the spanning cell needs, and the shortfall is distributed across them.

## Repro

```html
<style>
table { border-collapse: collapse; }
td { border: 1px solid #333; padding: 0; }
</style>
<table>
  <tr><td colspan="2">AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA</td></tr>
  <tr><td>X</td><td>YYYYYYYYYYYYYYYYYYYYYYYY</td></tr>
</table>
```

Column 1 holds a single `X`. Where the second column starts (`pdftotext -bbox`):

| | column 1 width |
| --- | --- |
| `main` | **131.68 pt** — half the colspan cell, forced in |
| with this change | **10.83 pt** — what `X` actually needs |

On `main` the narrow column is inflated to half the spanning cell's width and the wide column loses the space it needed.

## Why it matters

A header cell spanning the table above a row of narrow data columns is an everyday table shape — an invoice, a report, any grouped heading. Every column in the table ends up mis-proportioned, and on a fixed-width page that pushes content out of its column or off the page.

## Tests

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-colspan-no-longer-widens-a-column-it-already-fits-into.md
+++ b/.claude/migration-notes/2026-09-08-a-colspan-no-longer-widens-a-column-it-already-fits-into.md
@@ -1,0 +1,27 @@
+# A `colspan` cell no longer widens a column the span already fits into
+
+An automatic-layout table sized its columns by dividing a spanning cell's width by its span and
+forcing every column it covers to at least that share. CSS 2.1 §17.5.2.2 asks for the opposite
+reading: a spanning cell constrains the columns it spans *together* — "increase the minimum widths
+of the columns it spans so that together, they are at least as wide as the cell" — so it should not
+move them at all when their sum already accommodates it.
+
+The old rule dragged a narrow column up to the span's average whatever its own content needed, and
+the surplus then pushed every later column boundary out. A `colspan="2"` cell straddling a
+one-character column and a long one moved boundaries that a browser leaves alone.
+
+Such a span now leaves its columns where they are. A span genuinely wider than its columns still
+widens them, sharing the shortfall in proportion to what each column already measures.
+
+**What a document author may notice.** Column boundaries in an automatic-layout table containing
+`colspan` cells can shift, and in general shift toward what a browser produces. Measured against
+Chrome 152 on a three-column table with a spanning row that fits inside its columns' sum, boundaries
+in points with the page margin subtracted:
+
+| | second boundary | third boundary |
+| --- | --- | --- |
+| Chrome 152 | ~7.0 | 197.3 |
+| after this change | 7.2 | 195.1 |
+| before | 23.3 | 211.2 |
+
+See [Tables](../../docs/html-css-support.md#tables) in `docs/html-css-support.md`.
--- a/.claude/recent-fixes/2026-09-08-a-colspan-constrains-its-columns-together.md
+++ b/.claude/recent-fixes/2026-09-08-a-colspan-constrains-its-columns-together.md
@@ -1,0 +1,40 @@
+# A `colspan` constrains its columns together, not each one individually
+
+`GetColumnsMinMaxWidthByContent` divided a spanning cell's min/max width by its span and
+`Math.Max`'d the result into every column it covered. CSS 2.1 §17.5.2.2 step 3 constrains the spanned
+columns *together* — "so that together, they are at least as wide as the cell" — and says nothing
+about any one of them. The difference only shows when the spanned columns are uneven: dividing drags
+the narrow one up to the average, and the surplus pushes every later boundary out.
+
+Spanning cells are now deferred to a second pass, so every single-column cell has already been
+applied and the sums a span is measured against are final. `SpreadSpannedWidth` then adds only the
+shortfall, if there is one.
+
+## What running it turned up
+
+- **The obvious repro does not discriminate.** A four-column header with two `colspan="2"` detail
+  rows — the shape this was originally found on — gives identical output before and after, because
+  the span genuinely needs widening there and both rules widen it. The fixture that separates them
+  is an *uneven* spanned pair whose sum already fits: a one-character column beside a long one.
+- **Chrome agrees, and the measurement is now in the doc comment.** Rendering that fixture through
+  Chrome 152 headless and reading positions with `pdftotext -bbox`, the second column boundary is at
+  ~7.0pt and the third at 197.3pt past the page margin; this change gives 7.2 and 195.1, the old rule
+  23.3 and 211.2. The residual ~2pt is border/padding modelling, present in both rows.
+- **The proportional split is an approximation, and is now labelled as one.** §17.5.2.2 says to widen
+  the spanned columns by "approximately the same amount". Sharing the shortfall in proportion to what
+  each column already measures avoids re-inflating a column already sized by its own content, and no
+  browser implements §17.5.2.2's automatic algorithm literally — but it is a choice, not the spec
+  text, and the two differ only in how a genuine shortfall is split.
+- **Overlapping spans are order-dependent**, because spans apply in document order and each mutates
+  the widths the next measures against. The invariant that survives is §17.5.2.2's own requirement as
+  a postcondition — every spanning cell's columns together hold that cell — which growth-only updates
+  reach in any order. Pinned by a fixture with a `colspan="3"` over columns 0-2 and a `colspan="2"`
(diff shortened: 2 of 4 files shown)


dchupp:

jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (layout-engine testing, doc-comment hygiene, zero-warning builds, table invariants). Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` and the `Table`-filtered test suite, and cross-checking the algorithm against CSS 2.1 §17.5.2.2.

The diff touches only `CssLayoutEngineTable.cs` — no test file is changed at all. `CLAUDE.md`'s "Layout engine changes" convention specifically requires unit tests that assert the affected `CssBox`'s properties after layout (`Location`, `ActualRight`/`ActualBottom`), not just "layout completes." This change is a textbook case for that bar: it's exactly the kind of column-width arithmetic the convention calls out, and the PR's own evidence is a manual `pdftotext -bbox` measurement, not an automated assertion. Running `dotnet test --filter FullyQualifiedName~Table` on this branch passes 697/697 with the same count as `main` would show — i.e. nothing in the existing suite exercises "narrow column + wide colspan sibling," so a future change could silently reintroduce the old behavior and nothing would fail. Please add a `CssLayoutEngineTableTests`-style fixture (harness pattern in `FlexboxIntegrationTests.cs`) that builds the repro table (colspan=2 header over a narrow/wide column pair), performs layout directly, and asserts the resulting `_columnWidths`/cell `ActualRight` values — both for the "shortfall" case (spanned columns need widening) and a no-op case (spanned columns already sum to enough, to lock in that `SpreadSpannedWidth` correctly returns early via `shortfall <= 0`).

**2. Proportional distribution vs. spec's "approximately the same amount" — not obviously wrong, but unverified**

CSS 2.1 §17.5.2.2 step 3 says, verbatim: *"For each cell that spans more than one column, increase the minimum widths of the columns it spans so that together, they are at least as wide as the cell... If possible, widen all spanned columns by **approximately the same amount**."* `SpreadSpannedWidth` instead distributes the shortfall in proportion to each column's *current* width (`shortfall * (widths[index] / total)`), falling back to an equal split only when every spanned column currently measures zero. That's a defensible design choice (it avoids re-inflating a column that's already sized correctly from its own content), and it's plausibly closer to what real engines do since browsers don't literally implement CSS 2.1's automatic algorithm — but the doc comment's claim ("three of the four column boundaries here and none in Chrome") isn't backed by anything reproducible in the PR: no Chrome screenshot/measurement is shown, only PeachPDF's own before/after numbers. Given this is the crux of the whole fix, could you add the Chrome comparison to the PR description (or a code comment citing where it came from), and ideally a second repro case with *two overlapping* multi-column spans (e.g. row 1 `colspan=3` over cols 0-2, row 2 `colspan=2` over cols 1-2) — since `spanning` is processed in row/document order and each `SpreadSpannedWidth` call mutates `widths` in place, the final result is order-dependent when spans overlap, and that interaction isn't covered by the repro or any test.

**3. Doc-comment placement — no detachment issue this time**

Unlike the pattern in your earlier PRs (#927-933), `SpreadSpannedWidth`'s new XML doc comment (`CssLayoutEngineTable.cs` ~line 6117) sits directly above its own method and doesn't disturb the doc comment on the following method (`GetColumnsExplicitMaxWidth`-ish, ~line 6157+), which is still correctly attached. Good.

**Build/test verification**

- `dotnet build PeachPDF.slnx -t:Rebuild -v:minimal`: **0 Warning(s), 0 Error(s)**.
- `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0 --filter "FullyQualifiedName~Table"`: 697 passed, 0 failed, 0 skipped.
- Read `.claude/invariants/table-a-header-crossing-rowspan-cell-must-stay-unclosed-until-the-body-closes-it.md` and the recent `.claude/recent-fixes/2026-09-07-table-layout-auto-and-fixed-implemented.md` / `2026-08-25-table-shrink-columns-dead-code.md` entries — this change is scoped to the automatic-layout content-measurement path (`GetColumnsMinMaxWidthByContent`) and doesn't touch fixed-layout, rowspan-closing, or collapsed-border machinery, so none of those documented traps appear to apply here. It does correctly preserve the existing collapsed-column exclusion (`IsColumnCollapsed` checks carried into `SpreadSpannedWidth`).

Net: the diagnosis and the CSS 2.1 citation are right, and the collapsed-column handling is carried over carefully, but this needs an automated geometry-asserting test before it can merge per this repo's own convention, and the proportional-vs-equal distribution choice deserves either evidence or a tracked follow-up if it's a deliberate approximation.

dchupp:
Both addressed, rebased onto `main` and force-pushed.

**2. The Chrome comparison you asked for — measured, and it changed the fixture.** You were right that the claim was unbacked; it turns out the shape I had described does not even discriminate. A four-column header with two `colspan="2"` detail rows gives *identical* output before and after, because the span genuinely needs widening there and both rules widen it.

What separates them is an **uneven** spanned pair whose sum already fits: a one-character column beside a long one, with a `colspan="2"` row narrower than their total. Rendered through Chrome 152 headless and read back with `pdftotext -bbox` (column boundaries in points, page margin subtracted):

| | second boundary | third boundary |
| --- | --- | --- |
| Chrome 152 | ~7.0 | **197.3** |
| this change | 7.2 | **195.1** |
| dividing by span | 23.3 | **211.2** |

The residual ~2pt against Chrome is border/padding modelling, present in both PeachPDF rows and unrelated to this rule. That table is now in `SpreadSpannedWidth`'s doc comment, along with the repro shape, so the claim travels with the code.

**The proportional split is now labelled as the approximation it is.** You are right that §17.5.2.2 says "approximately the same amount" and this does something else. The doc comment states that explicitly, gives the reasoning (it avoids re-inflating a column already sized correctly by its own content, and no browser implements §17.5.2.2's automatic algorithm literally), and notes that the two only ever differ in how a *genuine* shortfall is split, never in whether one exists.

**Overlapping spans.** Also right, and now covered. Spans apply in document order and each mutates the widths the next measures against, so the split is order-dependent. What holds regardless is §17.5.2.2's own requirement as a **postcondition** — every spanning cell's columns together are at least as wide as that cell — which is reachable in any order precisely because these updates only ever grow a column. `OverlappingSpans_LeaveEveryCellWithEnoughRoom` pins it with your suggested shape: `colspan="3"` over columns 0–2 and `colspan="2"` over columns 1–2.

**1. Tests.** `TableColspanColumnSizingTests`, four fixtures, each measured against a control table differing only in the row under test rather than against a hardcoded coordinate:

- a span that already fits (no boundary moves) — fails against the merge base;
- a span that does not (they move) — the early-return must mean "already wide enough", not "never widen";
- the proportional ordering (the column already carrying more content takes more of the shortfall) — fails against the merge base;
- the overlapping-span postcondition.

**3.** Thanks for checking the doc-comment placement — I have since swept every branch in this batch for that pattern and the only remaining hits pre-date them and are on `main` (details in my reply on #929).

The collapsed-column exclusion is carried into the new pass unchanged, as you noted. Also added `.claude/migration-notes/2026-09-08-a-colspan-no-longer-widens-a-column-it-already-fits-into.md` with the Chrome table, since column boundaries in an existing document can shift, and a recent-fixes entry.

jhaygood86 (approved):
Verified the update: `TableColspanColumnSizingTests.cs` adds 4 fixtures with real post-layout assertions (`cell.Location.X`/`ActualRight`), including a fits-already control (locks in the early-return path) and an overlapping-spans postcondition check. The Chrome-comparison evidence I asked for landed in `SpreadSpannedWidth`'s doc comment (Chrome 152 measurements) rather than the PR description, which satisfies the "or cited in a code comment" alternative I offered. The overlapping-span order-dependence is correctly reasoned as unfixable-by-design given the in-place-mutation approach, and tested via a postcondition instead — a reasonable call, documented in both the doc comment and a recent-fixes note.

PR #939, opened 2026-09-08
system-ui resolves through fontconfig on Linux

dchupp:
The five other generic families ask fontconfig on Linux. `system-ui` alone is hardcoded to the default font, so it is the **one family that ignores the host's own configuration**. Chromium delegates `system-ui` to fontconfig on Linux too.

## Repro

On a host where fontconfig resolves the system font to FreeSans:

```
fc-match system-ui   ->  FreeSans
fc-match sans-serif  ->  FreeSans
```

```html
<style>
.a { font-family: system-ui;  font-size: 20pt; }
.b { font-family: sans-serif; font-size: 20pt; }
</style>
<div class="a">SYSTEMUI</div>
<div class="b">SANSSERIF</div>
```

Fonts embedded in the output (`/BaseFont` entries):

| | embedded |
| --- | --- |
| `main` | `FreeSans`, **`Liberation Sans`** |
| with this change | `FreeSans` |

`sans-serif` goes through fontconfig and gets FreeSans; `system-ui` in the same document bypasses it and gets the hardcoded default.

## Why it matters

It is not only the glyphs. FreeSans and Liberation Sans disagree on line box height by **11.7%** (ascent + descent of 1.000 em against 1.117 em), so any block set in `system-ui` lays out to a different height than the same block under a browser on the same machine — and on a paginated document that difference accumulates into a different page count.

`system-ui` is also the natural family for chrome-like furniture (headers, footers, generated UI), which is exactly the content most likely to be compared against a browser rendering.

## Scope

Linux only, and it changes nothing on a host whose fontconfig resolves `system-ui` to the same face the default already was.

## Tests


dchupp:

jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (test coverage for behavior changes, doc-comment hygiene, zero-warning builds, docs-land-with-the-feature). Verified by checking out the branch, running `dotnet build PeachPDF.slnx -t:Rebuild` and the font-resolution test filters.

`gh pr diff 939 --name-only` shows only `src/PeachPDF/Adapters/PdfSharpAdapter.cs` changed — no test file. That's notable here because this repo already set the exact precedent for testing this code path for the five other generics, and neither extension point was used:

- `LinuxSystemFontResolverTests.cs`'s `ResolveGenericFamily_DoesNotThrow` and `ResolveGenericFamily_OnLinux_ReturnsARealInstalledFamilyName` theories (`[InlineData("serif")]`, `"sans-serif"`, `"monospace"`, `"cursive"`, `"fantasy"`) exercise `LinuxSystemFontResolver.ResolveGenericFamily` directly — `"system-ui"` isn't in either list, even though it's now resolved through the identical call.
- `GenericFontFamilyIntegrationTests.cs`'s `Generic_OnLinux_ResolvesToARealInstalledFontconfigFamily` theory does the same at the `PdfSharpAdapter` level (`adapter.IsFontExists(generic)`) — again, no `"system-ui"` case added.

There's also no unit test for the new fallback branch itself (`systemUi is null || !IsFontExists(systemUi)` → `DefaultFontResolver.DefaultFont`), which `DefaultFontFallbackTests.cs` shows this repo's convention for (a synthetic unresolvable name, since the real default font is installed on every CI runner). This is exactly the kind of platform-specific path the PR body itself calls out ("Linux only") — CI can't visually confirm it, so the automated theories above are the only signal a regression here would produce. Could you add `"system-ui"` to those two existing theories, plus one `DefaultFontFallbackTests`-style case for the new fallback?

**2. Docs now contradict the new behavior (docs-land-with-the-feature)**

`docs/usage-examples.md`'s "Generic families and `system-ui`" table (line 283) still lists the Linux column for `system-ui` as *"(platform default font)"*, and the prose right below it (line 285) says: "`system-ui` on Windows is an exact match for Chromium's own `system-ui` → Segoe UI resolution; on macOS/Linux/Android it's a pragmatic approximation using the platform's default font." Both are now false on Linux after this change — it's fontconfig-delegated with the default-font fallback only when fontconfig fails or resolves to something not installed, matching the other five generics' row in that same table.

Worth noting `docs/html-css-support.md` (line 352) already described `system-ui` as included in "delegated to the OS's own fontconfig on Linux" — so the two docs pages were already inconsistent with each other before this PR, and this PR makes the code match `html-css-support.md` while leaving `usage-examples.md` now wrong in the other direction. Per `CLAUDE.md`'s "update the relevant doc page in the same change" convention, this needs the table cell + prose sentence fixed here rather than as a follow-up.

Also no file under `.claude/migration-notes/` — this is a user-visible behavior change on Linux hosts (a document's `system-ui` text can render in a different face than before, with the layout-height implications the PR body itself describes), which is exactly what that folder is for.

**3. Security / process-invocation review — no concern found**

`LinuxSystemFontResolver.ResolveGenericFamily` (unchanged by this PR, just called once more) resolves via direct P/Invoke against `libfontconfig.so.1` (`FcPatternCreate`/`FcPatternAddString`/`FcConfigSubstitute`/`FcDefaultSubstitute`/`FcFontMatch`) — it does not shell out to `fc-match`/`fc-list` via `Process.Start`, so there's no argument-escaping/shell-injection surface at all. The string passed (`"system-ui"`) is also a hardcoded literal in `PdfSharpAdapter.cs`, never derived from the HTML/CSS being rendered, so even a hypothetical shell-based implementation would have nothing untrusted to inject. Exception handling is already in place and unchanged: `ResolveGenericFamily` wraps the whole call in try/catch and returns `null` on any failure (missing `libfontconfig.so.1`, resolution failure, etc.), and the new call site handles that `null` (or a resolved-but-not-installed family) by falling back to `DefaultFontResolver.DefaultFont` — so a broken/absent fontconfig install degrades to the pre-PR behavior rather than throwing.

**4. Platform gating and core claim look right**

`isLinux` reuses the adapter's existing `OperatingSystem.IsLinux() && !isAndroid` flag from the generics loop just above, so Windows/macOS/Android take the `systemUi is null` branch straight to `DefaultFontResolver.DefaultFont` — identical to current behavior, confirmed by the untouched `GenericFontFamilyIntegrationTests.Monospace_OnWindows_ResolvesToConsolas_NotCourierNew`-style assertions still passing. On the core claim: fontconfig has shipped a `system-ui` generic alias (defaulting to sans-serif) since 2.13.1, so `fc-match system-ui` resolving to a real family on a modern distro — and Chromium delegating to it on Linux — is accurate, not a stretch.

**5. Doc-comment-detachment pattern (flagged on #927/#929/#930/#933) — not present here**

This change is entirely inside the existing constructor body; no new method is introduced, so there's no doc comment to get textually detached from a following method's own comment this time.

Build: `dotnet build PeachPDF.slnx -t:Rebuild -v:minimal` → 0 warnings, 0 errors. `LinuxSystemFontResolverTests`/`GenericFontFamilyIntegrationTests`/`DefaultFontFallbackTests` filters: 43 passed, 0 failed (as expected, none exercise `system-ui` yet — see #1).

Otherwise this closes a real, well-explained inconsistency (the one generic-ish family that ignored the host's own fontconfig despite the code right above it doing exactly that for five others) with a sensible verified-install fallback. Please add the missing test cases and the two doc updates before merge.

dchupp:

**1. Tests.** `"system-ui"` added to the three existing theories you named — `LinuxSystemFontResolverTests.ResolveGenericFamily_DoesNotThrow`, `..._OnLinux_ReturnsARealInstalledFamilyName`, and `GenericFontFamilyIntegrationTests.Generic_OnLinux_ResolvesToARealInstalledFontconfigFamily` — which is where the other five live, so it is covered the same way rather than in a parallel fixture.

For the fallback branch I had to make it reachable first: `systemUi is null || !IsFontExists(systemUi)` never fires on a real machine, so a test against the constructor would hold whether or not the code did anything — the same trap `DefaultFontFallbackTests` documents about using the real default font. The decision is now a pure function:

```csharp
internal static string ResolveSystemUiFamily(string? fontconfigFamily, Func<string, bool> fontExists)
```

with three fixtures: no fontconfig answer, an answer naming a family that is not installed, and an answer that is (the contrast case — without it the first two also pass if the mapping always fell back).

**2. Docs.** Both corrected in `docs/usage-examples.md`: the Linux column of the `system-ui` row now reads *(delegated to fontconfig's own `system-ui` alias)* like the five above it, and the prose below no longer says Linux is a platform-default approximation. You were right that the two pages already disagreed with each other before this change and that fixing only one would have left the contradiction pointing the other way — `docs/html-css-support.md` already described the fontconfig behaviour, and the code now matches it.

Migration note added: `.claude/migration-notes/2026-09-08-system-ui-resolves-through-fontconfig-on-linux.md`. It records the user-visible part with the number — FreeSans against Liberation Sans is an 11.7% line-box height difference (ascent+descent 1.000em vs 1.117em), and `system-ui` is what generated header/footer bands are drawn in.

**3.** Thank you for tracing the P/Invoke path rather than assuming a `Process.Start`. I have written that into the recent-fixes entry, since "does this shell out to `fc-match`" is a reasonable thing for the next reader to wonder and the answer is not obvious from the call site.

jhaygood86 (approved):
Verified the update. All three points landed cleanly:

1. **Tests**: `"system-ui"` added to the three existing generic-family theories, plus a nice extraction — `PdfSharpAdapter.ResolveSystemUiFamily(fontconfigFamily, fontExists)` pulled out as a pure function specifically because the fallback branch is unreachable on any real machine (fontconfig's own answer is always installed), with three fixtures covering all arms including the contrast case. Good instinct catching that a test against the constructor directly would have passed whether or not the code did anything.
2. **Docs**: both `docs/usage-examples.md`'s table cell and prose corrected to match `docs/html-css-support.md`, plus a migration note with an actual quantified impact (11.7% line-box height difference between FreeSans and Liberation Sans) rather than a vague "may differ."
3. Appreciated writing the P/Invoke-not-Process.Start finding into the recent-fixes entry for the next reader.

Approving. Merging now.

PR #940, opened 2026-09-08
A flex item with block children is measured by its content, not its container

dchupp:
`CssLayoutEngineFlex.MeasureItem` derives an auto-basis item's hypothetical main size from its own line boxes, and falls back to the laid-out width when there are none:

```csharp
// Block children: no word measurement; use container fill width as fallback.
maxContent = naturalMain;
```

An auto-width block child **fills its container** on layout, so `naturalMain` is the *container's* width, not the item's content. Every item in the line therefore measures the same, and the line divides evenly however different the items actually are.

## Repro

```html
<style>
.row  { display: flex; width: 500px; border: 1px solid #333; }
.item { border: 1px solid #a00; }
</style>
<div class="row">
  <div class="item"><div>SHORT</div></div>
  <div class="item"><div>A MUCH LONGER PIECE OF CONTENT HERE</div></div>
</div>
```

Word positions (`pdftotext -bbox`):

| --- | --- | --- |
| item 1 (`SHORT`) width | **~187 pt** — half the container | **~40 pt** — its content |
| item 2 starts at | x = 225.00 | x = 77.30 |
| item 2's text | **wraps onto two lines** (`CONTENT`/`HERE` at y = 50.70) | **one line**, everything at y = 37.50 |

Both items measure identically on `main` despite one holding five words more than the other, so the 500 pt splits evenly and the longer item wraps inside its half while the shorter one keeps 150 pt of empty space.

## Why it matters

Wrapping a flex item's contents in a `<div>` is the ordinary way to write one. As soon as an item's children are block-level rather than bare text, the whole line stops sizing to content and divides evenly instead.

## Tests

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-flex-item-with-block-children-is-sized-by-its-content.md
+++ b/.claude/migration-notes/2026-09-08-a-flex-item-with-block-children-is-sized-by-its-content.md
@@ -1,0 +1,25 @@
+# A flex item whose children are blocks is now sized by its content, not by its container
+
+Previously, a flex item whose children were block-level boxes measured as wide as the flex
+container itself. An auto-width block child fills its containing block during layout, so reading
+the item's laid-out width reported the container's width rather than the item's own content — every
+item on the line measured the full container, `justify-content` was left with no free space to
+distribute, and the line rendered as an even split however narrow the items' content actually was.
+A `space-between` row of two short items landed each one on half the page instead of hugging its
+own edge.
+
+Such an item is now measured by its own intrinsic content, per css-flexbox-1 §9.2 — so
+`justify-content` (`space-between`, `space-around`, `flex-end`, `center`) has real free space to
+distribute and positions the items where the property asks for.
+
+Two narrower changes ride along in the same measurement:
+
+- White space at the end of a wrapped line no longer counts toward an item's measure
+  (css-text-3 §4.1.2 — it hangs). A wrapped item previously measured one word space wider than it
+  drew, which could over-subscribe the line and make `flex-shrink` narrow a sibling that had room.
+- An item whose own min-content is wider than the line now keeps its min-content width and
+  overflows, rather than being sized below what it can draw. This is CSS 2.1 §10.3.5's
+  shrink-to-fit, `min(max(min-content, available), max-content)`, where min-content is the lower
+  bound.
+
+See [Flexbox](../../docs/html-css-support.md#flexbox) in `docs/html-css-support.md`.
--- a/.claude/recent-fixes/2026-09-08-flex-item-hypothetical-size-from-content-not-container.md
+++ b/.claude/recent-fixes/2026-09-08-flex-item-hypothetical-size-from-content-not-container.md
@@ -1,0 +1,48 @@
+# A flex item's hypothetical main size comes from its content, not from its laid-out width
+
+`MeasureItem`'s main-axis-physical-X branch derived an item's max-content width from its
+*laid-out* width (`naturalMain`) whenever the item was not inlines-only. That is correct for a
+replaced leaf, and wrong for a box with block children: an auto-width block child fills its
+containing block during layout, so `naturalMain` is the flex CONTAINER's width, not the item's
+content. Every item then measured the container, the free space `justify-content` distributes came
+out as zero, and the result read as an even split.
+
+The load-bearing idea is that the item has a real intrinsic measurement available already —
+`CssLayoutEngine.GetMinContentWidth`/`GetMaxContentWidth`, the same pair the float sizing a few
+lines above relies on, both returning outer widths, which is what `hypothetical` is.
+
+## What running it turned up
+
+- **The clamp order is not interchangeable.** The first version wrote
+  `min(max(max-content, min-content), available)`, which — given the invariant max-content ≥
+  min-content — collapses to `min(max-content, available)` and has no min-content floor at all.
+  CSS 2.1 §10.3.5 is `min(max(min-content, available), max-content)`: min-content is the LOWER
+  bound. The two agree everywhere except the case that matters, an item whose own min-content
+  exceeds the line, where the correct answer is to keep min-content and overflow. Pinned by
+  `FlexboxIntegrationTests.Item_WhoseMinContentExceedsTheRow_KeepsItsMinContent`, which needs
+  `flex-shrink: 0` — with shrink left on, the shrink pass clamps the item back to the line either
+  way and the fixture says nothing about the clamp.
+- **`GetFitContentWidth` is not the helper to reach for here**, despite the name. It is
+  `min(GetLargestChildWidth(max-content), available)` — no min-content floor either — so it has the
+  same gap. An earlier comment in this method cited it while the code called something else; both
+  now say the same thing.
+- **A hanging space is worth 3.07pt at the default font**, and it was being counted. `LineContentWidth`
+  trims it per css-text-3 §4.1.2. The observable symptom is an item box wider than anything drawn
+  inside it: 0.01pt of slack (the deliberate anti-rounding epsilon) with the trim, 3.07pt without.
+  That difference is what `WrappedItem_DoesNotCountTrailingWhiteSpaceInItsMeasure` asserts — an
+  earlier version of that test put the text in a block child, which takes the other measurement
+  branch entirely, and passed with the trim reverted.
+
+## Evidence
+
+Three fixtures in `FlexboxIntegrationTests.cs`, each mutation-checked: reverting
+`CssLayoutEngineFlex.cs` to `main` fails two of them; reverting only the clamp order fails the
+min-content one; reverting only `LineContentWidth`'s trim fails the trailing-space one. Full suite
+10,201 passed / 0 failed / 9 skipped on net8.0; diff coverage 100% on the changed lines.
+
+## Deliberately not done
+
+The grid-container fallback (`box.DerivedStyle.ActualDisplay is Keywords.Grid or InlineGrid` →
+`naturalMain`) is left as it was. A nested grid places its children on tracks the intrinsic walk
+knows nothing about, so measuring it this way would report a number with no relationship to how it
+lays out; that needs the grid engine's own intrinsic sizing, which is a separate piece of work.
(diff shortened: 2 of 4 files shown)


dchupp:
Force-pushed a comment-only change to this branch — no code touched, and the diff against the previous head is entirely `///` and `//` lines.

Two things came out of @jhaygood86's review on #933 that apply here as well:

- **Customer names in code comments.** Several comments on this branch named specific customer documents as the case that found the bug. Those should never have been in a public PR; they are rewritten to the generic shape the comments were actually describing (an address block, a label/value column, a bordered badge in a table cell), keeping the measurements and the CSS, which is the part that was carrying the information.
- **The `VENDORED:` prefix.** That is our internal marker for "this comment came from our patch set". It means nothing in this repo and reads as a claim that the code was copied from a third party, which it was not. Stripped from every branch in this batch.

I also swept all fourteen branches for the doc-comment-detachment pattern @jhaygood86 flagged (a new member inserted between an existing member's doc comment and that member, so one `///` run ends up with two `<summary>` blocks) and fixed the instances that are ours.

jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (layout-engine testing, doc-comment hygiene, zero-warning builds, spec compliance). Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` and the `Flex`-filtered test suite, and reading the changed code in `CssLayoutEngineFlex.cs` against css-flexbox-1 §9.2.

The diff touches only `CssLayoutEngineFlex.cs` — no test file is added or modified. `CLAUDE.md`'s "Layout engine changes" convention specifically requires unit tests asserting the affected `CssBox`'s post-layout properties (`Location`, `ActualRight`/`ActualBottom`) using the `HtmlContainerInt` + `PdfSharpAdapter` harness already established in `FlexboxIntegrationTests.cs`. This PR is exactly the shape that convention exists for: it changes how a flex item's hypothetical main size is derived, and the PR body's own repro (two `.item` divs, one `SHORT`, one long, in a `500px` row) is a ready-made fixture. The PR body cites the full suite passing (10,207/0/9, unchanged from `main`) — that only proves no regression, not that the fix does what it claims. Please add a `FlexboxIntegrationTests.cs`-style test that lays out the repro HTML and asserts each `.item`'s `ActualRight`/width reflects its own content rather than an even split (e.g. item 1 width ≈ its `SHORT` content, item 2 wider, not both ≈ 250pt), plus a `MinWidth` edge case would help (see point 3).

**2. Doc-comment detachment (same pattern as #927-933)**

`LineContentWidth` is inserted between the pre-existing `// column-gap = .../row-gap = ...` two-line comment (at old-file lines now 1732-1733) and the method it actually documents, `ParseMainGap` (line 1755). That comment is left dangling above `LineContentWidth`, where it describes neither the gap helpers section it opens nor `LineContentWidth`'s own behavior. Compounding it, `LineContentWidth` also picked up its own `/// <inheritdoc cref="MeasureItem"/>` (line 1734) — but `MeasureItem` (line 374) has no XML doc comment to inherit (only `//` line comments), so this `inheritdoc` resolves to nothing and reads as a copy-paste leftover rather than real documentation. Please move `LineContentWidth` (with a doc comment describing what it actually does — trimming trailing whitespace per css-text-3 §4.1.3) above the `// ─── Gap helpers ───` banner, or below `ParseGap`, so the banner and its two-line comment land back directly on `ParseMainGap`.

**3. `Math.Max`/`Math.Min` ordering doesn't match the shrink-to-fit formula the comment cites**

Lines 476-479:
```csharp
maxContent = Math.Min(
    Math.Max(await CssLayoutEngine.GetMaxContentWidth(g, box),
        await CssLayoutEngine.GetMinContentWidth(g, box)),
    naturalMain);
```
CSS 2.1 §10.3.5's shrink-to-fit is `min(max(min-content, available), max-content)` — min-content is the *lower* bound, available width sits in the middle, max-content is the *upper* bound. This code instead takes `max(maxContent, minContent)` (which, given the invariant max-content ≥ min-content, always just evaluates to `maxContent`) and then clamps that to `naturalMain`. The two formulations agree whenever `naturalMain >= minContent`, but diverge when the flex item's own min-content is wider than the container (e.g. an unbreakable long word/token in the block child, in a narrow row): the correct shrink-to-fit result is `minContent` (content allowed to overflow), but this code produces `naturalMain`, which under-sizes the item below its own min-content. Worth either fixing the clamp order to `Math.Max(minContent, Math.Min(maxContent, naturalMain))`, or — simpler — calling the existing `CssLayoutEngine.GetFitContentWidth(g, box, naturalMain)` (declared right above `GetMinContentWidth`/`GetMaxContentWidth` in `CssLayoutEngine.cs`), which already implements this exact clamp and is what the inline comment at line 473 seems to think is being called ("`GetFitContentWidth`/`GetMinContentWidth` is the pair CSS 2.1 §10.3.5's shrink-to-fit uses") even though the code actually calls `GetMaxContentWidth`/`GetMinContentWidth`, not `GetFitContentWidth`. That mismatch between the comment's own citation and the code is worth reconciling regardless of which fix is chosen.

**4. `LineContentWidth`'s trailing-whitespace fix looks correct and is a real, separate improvement**

Independent of the block-children fix, trimming a line's trailing space out of `maxContent` (css-text-3 §4.1.3, trailing white space "hangs") is a legitimate fix in its own right and the reasoning in the inline comments (word-space double-counting vs. `GetMinMaxSumWords`) checks out against `CssRect.FullWidth`/`ActualWordSpacing`. This part isn't covered by a dedicated test either, though — the `Flex`-filtered suite passing unchanged doesn't demonstrate it (502/502 on net8.0, same as claimed). If a test is added for point 1's repro, it would be good for the fixture to also include a wrapped-text case that would fail if this trailing-space trim were reverted.

Build is clean (0 warnings via `dotnet build PeachPDF.slnx -t:Rebuild`), and the existing `Flex`-filtered suite passes (502 passed, 0 failed on net8.0) — but per point 1, that isn't sufficient sign-off for a layout-engine sizing change; please add the missing coverage before this merges.

dchupp:

**3 first, because you were right and it was a real bug.** `min(max(max-content, min-content), available)` collapses to `min(max-content, available)` given the invariant, so there was no min-content floor at all. Now `Math.Max(minContent, Math.Min(maxContent, naturalMain))` — CSS 2.1 §10.3.5's `min(max(min-content, available), max-content)`, which is the same function (they agree on all three orderings of the operands; I checked rather than assumed).

Not `GetFitContentWidth`, though: it is `min(GetLargestChildWidth(max-content), available)` and has no min-content floor either, so it has the same gap. The comment that cited it while the code called something else is reconciled — it now names what is called and says why the helper is not the right one.

Pinned by `Item_WhoseMinContentExceedsTheRow_KeepsItsMinContent`. It needs `flex-shrink: 0`: with shrink left on, the shrink pass clamps the item back to the row either way and the fixture says nothing about the clamp. Reverting just the clamp order fails it.

**1. Tests.** Three in `FlexboxIntegrationTests.cs`:

- `SpaceBetween_ItemWithBlockChildren_IsSizedByItsContent_NotAnEvenSplit` — your repro shape, asserting the short item hugs its content and `space-between` actually pushes the last item to the row's edge.
- `Item_WhoseMinContentExceedsTheRow_KeepsItsMinContent` — point 3 above.
- `WrappedItem_DoesNotCountTrailingWhiteSpaceInItsMeasure` — the trailing-space trim you asked to see covered.

Each is mutation-checked: reverting `CssLayoutEngineFlex.cs` to the merge base fails the first two; reverting only the clamp order fails the second; reverting only `LineContentWidth`'s trim fails the third. That last one caught a mistake worth mentioning — my first version put the text in a block child, which takes the *other* measurement branch entirely, so it passed with the trim reverted. The version that discriminates uses `max-width` to force the wrap and measures the slack between the item's box and the rightmost thing it draws: 0.01pt (the deliberate anti-rounding epsilon) with the trim, 3.07pt without.

**2. Doc-comment detachment and the `inheritdoc`.** `LineContentWidth` has moved above the `── Gap helpers ──` banner, so the `// column-gap = … / row-gap = …` comment is back on `ParseMainGap`. The `<inheritdoc cref="MeasureItem"/>` is gone — you were right that it resolved to nothing — replaced by a real summary.

**Also:** the css-text-3 citation was wrong. I had §4.1.3; both the leading- and trailing-space rules are §4.1.2 (Phase II: Trimming and Positioning). Checked against the spec text, and corrected here and on #937, which had the same drift.

**Convention items I had been missing** (from `CLAUDE.md`, having now read it properly rather than inferring):
`.claude/migration-notes/2026-09-08-a-flex-item-with-block-children-is-sized-by-its-content.md` — this changes rendered output for any `justify-content` row of block-child items — and `.claude/recent-fixes/2026-09-08-flex-item-hypothetical-size-from-content-not-container.md`. No docs change: `docs/html-css-support.md` documents `justify-content` as supported and does not describe the buggy behaviour, so there is no stale claim to correct. No showcase either — this is a bug fix, not a new capability.

Verification: `dotnet build PeachPDF.Tests -t:Rebuild` 0 new warnings (5 pre-existing CS1998s in the test project, present on `main`); `dotnet test --framework net8.0` green; **diff coverage 100%** on the changed lines, measured with the `coverlet.runsettings` command from `CLAUDE.md`. Rebased onto `189271a`.

jhaygood86 (approved):
Verified the update, including re-deriving the clamp-order fix myself: `Math.Max(minContent, Math.Min(maxContent, naturalMain))` is mathematically equivalent to CSS 2.1 §10.3.5's `min(max(min-content, available), max-content)` given the min-content ≤ max-content invariant (checked all three orderings of available vs. the two bounds — they agree). Genuinely fixes the bug rather than just rewording it, and `Item_WhoseMinContentExceedsTheRow_KeepsItsMinContent` with `flex-shrink:0` correctly isolates the clamp from the shrink pass, which would otherwise mask it.

The doc-comment detachment is fixed (`LineContentWidth` now sits above the gap-helpers banner, restoring `ParseMainGap`'s comment), the stray `<inheritdoc cref="MeasureItem"/>` that resolved to nothing is replaced with a real summary, and the mutation-testing methodology described (reverting each of the three independent fixes in turn and confirming each corresponding test catches it) is exactly the right way to prove these tests aren't vacuous. Also good catch on your own part fixing the css-text-3 §4.1.3 → §4.1.2 citation on both this PR and #937.

Approving. Merging now.

PR #941, opened 2026-09-08
A running element does not grow the document's ActualSize

dchupp:
A `position: running()` element is painted in a page **margin box**, so its width is bounded by the page margin, not by the flow. It still contributes to the container's `ActualSize`, which is the flow's own extent — so a wide running header or footer inflates a number describing content it is not part of.

`ShrinkToFit` reads `ActualSize` to pick a scale, which makes the consequence visible: a wide running footer shrinks **the whole document** to fit a band that was never in the flow.

## Repro

```html
<style>
@page { size: 8.5in 11in; margin: 1in; @bottom-center { content: element(foot); } }
#foot { position: running(foot); white-space: nowrap; font-family: Arial, sans-serif; font-size: 9pt; }
</style>
<div id="foot">WWWWWWWW…(wider than the page)</div>
<p>BODYSTART</p>
```

Rendered with `PdfGenerateConfig.ShrinkToFit = true`:

| | `BODYSTART` rendered width |
| --- | --- |
| `main` | **29.90 pt** — the document scaled to ~45% |
| with this change | **66.21 pt** — unscaled |

The body content is identical in both; only the running footer is wide.

## The change

The end-of-layout check that decides whether a box contributes to `ActualSize` now recognises a running element as well as `position: fixed`. Both are answered by one ancestor walk rather than two, since it runs at the end of every box's layout.

## Tests

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-running-header-no-longer-shrinks-the-whole-document.md
+++ b/.claude/migration-notes/2026-09-08-a-running-header-no-longer-shrinks-the-whole-document.md
@@ -1,0 +1,18 @@
+# A running header or footer wider than the content box no longer shrinks the whole document
+
+A `position: running()` element is painted into a page margin box, so its width is bounded by the
+page margin band rather than by the flow's content box — a header spanning the full page width is
+wider than the content box by design, and legitimately so. It was nonetheless growing the document's
+own recorded extent (`HtmlContainerInt.ActualSize`), which reports the document as overflowing its
+page when nothing in the flow does.
+
+With `ShrinkToFit` enabled, that made the whole document scale down to fit content that was never in
+the flow. Measured: a Letter invoice with a company header and footer band rendered at 0.968 scale,
+while the same document with both bands removed did not scale at all.
+
+A running element and its descendants no longer contribute to `ActualSize`. Ordinary content —
+including content that is genuinely too wide for the page — still does, so `ShrinkToFit` behaves
+exactly as before for everything that is really in the flow.
+
+See [Running elements](../../docs/html-css-support.md#running-elements-position-running--element) in
+`docs/html-css-support.md`.
--- a/.claude/recent-fixes/2026-09-08-a-running-band-is-not-part-of-the-flow-actualsize-measures.md
+++ b/.claude/recent-fixes/2026-09-08-a-running-band-is-not-part-of-the-flow-actualsize-measures.md
@@ -1,0 +1,42 @@
+# A running band is not part of the flow `ActualSize` measures
+
+`CssBox`'s layout epilogue grows `HtmlContainerInt.ActualSize` from every box it finishes, gated only
+on `IsFixed`. A `position: running()` element is painted into a page margin box, bounded by the
+margin band rather than the content box, so a full-width header band reported the document as
+overflowing its own page — and `ShrinkToFit`, which reads `ActualSize`, scaled everything down to fit
+content that was never in the flow.
+
+The gate is now `IsFixedOrInRunningElement`: one ancestor walk answering both, because the epilogue
+runs for every box and a second walk would be a measurable cost on a deep document.
+
+## What running it turned up
+
+- **Testing `ActualSize` directly proves nothing here.** Two fixtures built on the lightweight
+  `LayoutHarness` — a wide running box, and a wide box nested inside a running box — pass against the
+  pre-fix code as well as the fixed one; the epilogue's update only runs on the path the real
+  generator takes. The fixtures that discriminate go through
+  `PdfGeneratorLayoutHarness.LayoutWithRescaleAsync` and assert the rescale, which is also the
+  reported symptom rather than a proxy for it.
+- **`pixelsPerPoint` moves the opposite way to intuition.** It is the device scale the fit is
+  expressed as, so shrinking the document *raises* it: 900pt of content into a 540pt content box
+  comes out at ~1.68, i.e. drawn at ~0.6×. An unscaled document is exactly 1.0. The first version of
+  the contrast case asserted `< 1.0` and failed against correct behaviour.
+- **The width has to be on a CHILD of the running box to reach the defect.** The running box itself
+  was already skipped by an earlier flow-exclusion path; its wrapper children each ran the same
+  `ActualSize` update in their own right. A fixture putting the width on the running box passes
+  either way.
+- **Why not `IsExcludedFromFlow`.** It sits twelve lines away and answers an adjacent-sounding
+  question — whether a box contributes to its PARENT's in-flow content — and is the wider set,
+  including `IsOutOfFlow`. Reusing it here would also stop absolutely-positioned and floated content
+  growing `ActualSize`, which is a much larger behaviour change than the defect. Both doc comments
+  now cross-reference the other so the next reader does not have to re-derive why there are two.
+- **`IsFixed` is `virtual` with no override today**, and this walk re-implements its
+  `Position.Value == PositionMode.Fixed` test rather than calling it per ancestor (nesting two
+  ancestor walks is quadratic). Noted in the doc comment: if an override is ever added, the two have
+  to change together.
+
+## Evidence
+
+Two fixtures in `RunningElementLayoutIntegrationTests`, bracketing the behaviour: a running band
+wider than the content box leaves the document unscaled (fails against the merge base), and the
+identical wide box in the flow still shrinks it. Full suite green on net8.0, 0 build warnings.
--- a/src/PeachPDF.Tests/Integration/RunningElementLayoutIntegrationTests.cs
+++ b/src/PeachPDF.Tests/Integration/RunningElementLayoutIntegrationTests.cs
@@ -160,6 +160,89 @@ namespace PeachPDF.Tests.Integration
             Assert.Equal("heading", entry.Name);
         }
 
+        // ─── A running band must not grow the flow's own extent (HtmlContainerInt.ActualSize) ───
+        //
+        // Asserted through ShrinkToFit rather than by reading ActualSize directly. ActualSize is
+        // reachable from the lightweight harness, but a fixture built on it passes against the
+        // pre-fix code either way — the update only runs on the path the real generator takes. The
+        // rescale is also the reported symptom, so this exercises the defect rather than a proxy
+        // for it.
+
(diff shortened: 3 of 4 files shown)


dchupp:

jhaygood86:
Reviewed against this repo's `CLAUDE.md` conventions (layout-engine testing, doc-comment hygiene, zero-warning builds, reuse-existing-flow-exclusion-predicate). Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild`, and running the running-element/shrink-to-fit test suites.

`gh pr diff 941 --name-only` shows only `src/PeachPDF/Html/Core/Dom/CssBox.cs` touched — no test file. `CLAUDE.md`'s "Layout engine changes" convention requires a unit test asserting the affected box/container's properties after layout, not just "layout completes without throwing," and this is exactly that kind of change: `HtmlContainerInt.ActualSize` is a public, directly-assertable property (`src/PeachPDF/Html/Core/HtmlContainerInt.cs:560`). The repo already has the exact harness needed:

- `RunningElementLayoutIntegrationTests.cs` (e.g. `RunningBox_ContributesNoSizeAtOriginalPosition`) shows the `LayoutAsync`/`FindById` pattern for running-element assertions — add a case that lays out a document with a running box wider than its in-flow siblings and asserts `container.ActualSize.Width` matches a control document without the running box.
- `PdfGeneratorShrinkToFitRescaleTests.cs` + `PdfGeneratorLayoutHarness.LayoutWithRescaleAsync` is set up for exactly the PR's own repro (wide running footer under `ShrinkToFit`) — a test built from the PR description's repro HTML asserting `pixelsPerPoint` stays `1` (unscaled) would directly pin the regression this PR fixes and is the strongest possible test for it, since it exercises the real reported symptom end-to-end rather than only the internal predicate.

Without one of these, nothing in the suite would catch a future regression here (e.g. someone "simplifying" `IsFixedOrInRunningElement` back to `IsFixed`).

**2. Doc-comment placement — clean, not the detachment pattern from #927-933**

Checked for the recurring new-method-doc-inserted-before-existing-method's-own-doc shape from the earlier PRs in this series. Here `IsFixedOrInRunningElement` (`CssBox.cs:513`) and its doc comment (`CssBox.cs:493-512`) are inserted as a complete unit before `IsFixed`, and `IsFixed`'s own pre-existing doc comment (`CssBox.cs:530-535`, "Gets a value indicating whether this instance or one of its parents has Position = fixed.") is left intact directly above it, separated by a blank line. No detachment here, and the Rebuild is clean (0 warnings, 0 errors).

**3. Worth reconciling with the existing `IsExcludedFromFlow` predicate (`CssBox.cs:471-481`)**

The repo already has a doc'd, named answer to "every reason a box contributes nothing to its parent's in-flow content": `IsExcludedFromFlow => IsOutOfFlow || IsRunningPositioned`. This PR adds a second, differently-scoped predicate (`IsFixedOrInRunningElement`, ancestor-walking, `Fixed`-only rather than all of `IsOutOfFlow` which also includes `Absolute`/`Floated`) rather than referencing or extending that one. That's very plausibly the *correct* scope — the original `ActualSize` gate only ever excluded `IsFixed`, and widening it to full `IsExcludedFromFlow` would silently also stop absolutely-positioned/floated content from growing `ActualSize`, a materially bigger behavior change than this PR's stated scope — but the new doc comment doesn't mention `IsExcludedFromFlow` at all or explain why it deliberately isn't reused here, even though the two properties sit 12 lines apart and answer adjacent-sounding questions. A future reader (or the author of the next flow-exclusion predicate) is likely to wonder why there are now two of these. Worth a one-line cross-reference either in `IsFixedOrInRunningElement`'s doc or a "see also" added to `IsExcludedFromFlow`'s.

Separately: `IsFixed` (`CssBox.cs:536`) is `virtual`, but I didn't find any override in the tree, so the new property's decision to duplicate its ancestor-walk inline rather than call `base`-equivalent logic isn't hiding a lost override — just flagging that a future `IsFixed` override (if one is ever added for a box subtype) would silently *not* be picked up by `IsFixedOrInRunningElement`'s own walk, since it re-implements the `Position.Value == PositionMode.Fixed` check directly rather than calling `box.IsFixed`-per-ancestor. Probably fine given `IsFixed` has no overrides today, but worth a one-line note if that's intentional.

**Correctness / spec check**

The css-gcpm-3 model (a running element is removed to a page margin box, analogous to how `position: fixed`/`absolute` content is excluded from the normal flow's content-box sizing) matches what this PR does, and the "wrapper divs are children of the running box" reasoning for why an ancestor walk (not a self-only check) is needed checks out against how `MarginBoxContentFragmentBuilder.cs:31` and the flex/grid/columns engines already special-case `IsRunningPositioned` for their own immediate-child filtering. Existing running-element and shrink-to-fit tests (36 tests across `RunningElement*Tests.cs`, `ShrinkToFit*Tests.cs`) all still pass on this branch.

First-look review only — not approving/merging.

dchupp:

**1. Tests** — and your steer toward `PdfGeneratorShrinkToFitRescaleTests` was the right one for a reason I only found by trying the other first. Two fixtures in `RunningElementLayoutIntegrationTests`, bracketing the behaviour:

- `ShrinkToFit_DoesNotScaleTheDocumentDownForARunningBand` — the reported symptom end to end via `PdfGeneratorLayoutHarness.LayoutWithRescaleAsync`, asserting the document stays unscaled. Fails against the merge base.
- `ShrinkToFit_StillScalesDownForTheSameContentInTheFlow` — the identical wide box, not running. Without it the first also passes if `ActualSize` stopped tracking width altogether.

I did try the `ActualSize`-directly version first, as you suggested, and it does not work: both a wide running box and a wide box nested inside one pass against the pre-fix code under the lightweight `LayoutAsync` harness — the epilogue's `ActualSize` update only runs on the path the real generator takes. So the rescale assertion is not just the strongest test, it is the only one of the two that discriminates. Two other things that cost me a cycle each and are now written down: the wide box has to be a **child** of the running box (the running box itself was already skipped by an earlier exclusion; its wrapper children each ran the update in their own right), and `pixelsPerPoint` moves the *opposite* way to intuition — shrinking raises it, so an unscaled document is exactly 1.0 and my first contrast assertion had the comparison backwards.

**3. `IsExcludedFromFlow`.** Your read of the scope is right and it is now written down on both properties rather than left to be re-derived. `IsFixedOrInRunningElement`'s doc says why it is not reused — `IsExcludedFromFlow` answers "does this contribute to its parent's in-flow content" and is the wider set, including `IsOutOfFlow`, so widening the gate to it would also stop absolutely-positioned and floated content growing `ActualSize`, which `ShrinkToFit` would feel immediately — and `IsExcludedFromFlow` gets a "see also" pointing the other way.

The `IsFixed` `virtual` observation is also noted at the same place: the walk re-implements the `Position.Value == PositionMode.Fixed` test rather than calling `IsFixed` per ancestor because `IsFixed` is itself an ancestor walk and nesting the two is quadratic — with the consequence you identified spelled out, that a future override would not be picked up and the two would need to change together.

**2.** Thanks for checking the doc-comment placement; I have since swept every branch in this batch for that pattern and the only remaining hits pre-date them and are on `main`.

One more thing came out of the coverage pass: the walk's `box.ParentBox == box` guard — which mirrors `IsFixed`'s own — was the single uncovered line on the diff. It has a direct fixture now rather than being left as an uncovered loop-termination guard.

Also added a migration note (a document with a full-width running header rendered at 0.968 scale before this) and a recent-fixes entry.

dchupp:
CI note: the `windows-latest` leg here failed on a test this branch does not touch, and I do not have rerun rights on the repo — could you kick it, or tell me if you'd rather I did something about it?

```
Failed PeachPDF.Tests.Html.Core.Dom.CssLayoutEngineTableTests
       .TableLayout_ShrinkSearchWrapsAroundUnreducibleColumn_TerminatesQuickly
  Shrink search should terminate quickly; took 6583ms
```

`ubuntu-latest` and `macos-latest` both passed. `CssLayoutEngineTableTests.cs` is byte-identical between this branch and `main` (`git diff origin/main..HEAD -- <that file>` is empty), and the branch's only production change is the `ActualSize` gate in `CssBox.cs` — nothing on the table shrink-search path.

It is a wall-clock bound (`< 5000ms`, measured 6583ms) on a runner that was building three OS legs across eight of my PRs at the time. `FloatLayoutRegressionTests`'s own class comment already records this failure mode for a bound in that file — *"A wall-clock bound here was flaky on a contended CI runner … and could not be fixed by raising the bound, since raising it is exactly what removes its ability to see the regression it guards"* — and the same reasoning applies here: the genuine runtime is sub-millisecond and the pre-fix behaviour was 10+ seconds or a throw, so what this bound really wants to distinguish is "terminates" from "does not".

If it recurs, the shape that would settle it without weakening the guard is an iteration/visit counter rather than elapsed time — the same move `FloatLayoutRegressionTests` made when it swapped its wall-clock bound for `FloatScanBoxVisits`/`FloatScanCalls`. Happy to send that as its own PR if you'd like it; it is unrelated to this one so I have not touched it here.

jhaygood86 (approved):
Verified the update: `RunningElementLayoutIntegrationTests.cs` gained three tests via the exact harness suggested — `ShrinkToFit_DoesNotScaleTheDocumentDownForARunningBand` (the actual regression, using `PdfGeneratorLayoutHarness.LayoutWithRescaleAsync`), `ShrinkToFit_StillScalesDownForTheSameContentInTheFlow` (the contrast case, so the first test can't pass vacuously), and a loop-termination guard test. Nice catch in the PR noting that the lighter-weight `LayoutAsync` harness wouldn't actually discriminate pre/post-fix since the rescale epilogue only runs on the real-generator path — that's a more honest test than what was literally asked for. The doc-comment clarity nit (relationship to `IsExcludedFromFlow`) is also addressed with a cross-reference in both directions.

One Windows CI failure (`TableLayout_ShrinkSearchWrapsAroundUnreducibleColumn_TerminatesQuickly`, a wall-clock complexity guard from #925, unrelated to this change) — confirmed pre-existing and unrelated to this diff; net10.0 passed all 10,233 tests on the same run. I'll follow up separately on that test since it's flaky by construction (`Stopwatch.ElapsedMilliseconds < 5000`), matching a trap this repo's own `.claude/invariants/testing-a-complexity-guard-asserts-a-count-not-the-clock.md` already documents.

Approving. Merging now.

PR #945, opened 2026-09-08
Stop the left float scan at its formatting context

dchupp:
`FindNarrowestRightFloatBox` breaks its ancestor walk at a formatting-context boundary — a float cannot affect content outside its own (CSS 2.1 §9.5, css-display-3 §2.1). `FindIntersectingFloatBox`, the **left-side** point-collision walk, does not: it runs to the document root and scans every preceding sibling on the way.

Same rule, same traversal shape, one side missing it. This is the other half of #930.

## Cost

A synthetic document of 40 two-column grids, each grid item holding one `float: left`, measured on `main` with the engine's own counters:

| | scan calls | box visits | visits per call |
| --- | --- | --- | --- |
| `main` | 2,880 | **237,757** | 82.6 |
| with this change | 2,880 | **1,200** | 0.4 |
| `FindNarrowestRightFloatBox`, same document | — | — | ~7 |

The call count is unchanged — this is not skipping work that needed doing, it is not walking past the boundary where nothing can be found.

```html
<div style='display:grid; grid-template-columns:200pt 200pt; width:400pt;'>
  <div><div style='float:left; width:80pt; height:20pt;'></div><p style='margin:0;'>col a text here</p></div>
  <div><p style='margin:0;'>col b text here</p></div>
</div>
<!-- ×40 -->
```

## No rendered output changes

Worth being explicit, because the rule is stated as a correctness one: I could not construct a document where the leak actually moved a line. The walk was climbing past formatting-context boundaries and finding nothing to collide with, so on real content this is wasted work rather than a visible defect. It is offered as a cost fix, and tested as one.

## The starting-level exemption

The rule is deliberately **not** applied at the walk's starting level, and that differs from `FindNarrowestRightFloatBox` — which is worth spelling out, since #930's review asked for exactly this distinction in the other direction.

This walk's `reference` is the box being **placed**, and a float is itself a formatting-context root for its own contents. Breaking before that first level's siblings are scanned shields a float from the very siblings it has to be positioned against; `FloatLayoutRegressionTests.FloatRight_InNarrowerNestedBlock_AvoidsAWiderAncestorFloatRightSibling` and two others fail immediately if you try it. From the second level up, `reference` is a proper ancestor and the content being placed really is inside it, so the rule applies.

`FindNarrowestRightFloatBox` needs no exemption because of what it is handed: its `reference` comes from `FlowBox`'s word-flow loop and is the block whose line is being laid out, never a float itself.

## Tests

`FloatLayoutRegressionTests.LeftFloatScan_StopsAtAFormattingContextBoundary` asserts fewer than 10 box visits per scan call on the document above; without the boundary check the same document reports 82.6.

Asserted on `FloatScanBoxVisits`/`FloatScanCalls` rather than elapsed time, per that class's own doc comment on why a wall-clock bound there was flaky on a contended runner — and on a counter rather than geometry, since the geometry does not change.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,253 passed, 0 failed, 9 skipped.

### DIFF
--- a/.claude/recent-fixes/2026-09-08-the-left-float-scan-stops-at-its-formatting-context.md
+++ b/.claude/recent-fixes/2026-09-08-the-left-float-scan-stops-at-its-formatting-context.md
@@ -1,0 +1,31 @@
+# The left float scan stops at its formatting context, like the right one already did
+
+`FindNarrowestRightFloatBox` breaks its ancestor walk at a formatting-context boundary — a float
+cannot affect content outside its own (CSS 2.1 §9.5, css-display-3 §2.1). `FindIntersectingFloatBox`,
+the left-side point-collision walk, did not: it ran to the document root and scanned every preceding
+sibling on the way. Same rule, same shape, one side missing it.
+
+## What running it turned up
+
+- **It is the dominant cost of float layout, not a rounding error.** On a synthetic document of 40
+  two-column grids each holding one float: 237,757 box visits across 2,880 calls, 83 per call. With
+  the boundary check, 1,200 visits — the same 2,880 calls now examining under one box each. The
+  right-side walk, identical but for this rule, has always been ~7 per call.
+- **The starting-level exemption is load-bearing, and it is the opposite of the right-side walk's.**
+  This walk's `reference` is the box being *placed*, and a float is itself a formatting-context root
+  for its own contents — breaking before the first level's siblings are scanned shields a float from
+  the very siblings it must be positioned against. `FloatLayoutRegressionTests`'
+  `FloatRight_InNarrowerNestedBlock_AvoidsAWiderAncestorFloatRightSibling` catches it immediately.
+  `FindNarrowestRightFloatBox` needs no exemption because its `reference` comes from `FlowBox`'s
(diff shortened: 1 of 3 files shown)


jhaygood86 (approved):
Verified the critical asymmetry: `FindIntersectingFloatBox`'s new formatting-context break correctly skips itself at the starting level (via an `isStartingLevel` gate) so a float being positioned still scans its own preceding siblings on the first iteration — only from the second ancestor level onward does the boundary apply. This mirrors, and is the deliberate inverse of, how `FindNarrowestRightFloatBox` (from #930) checks unconditionally including its own starting level, since that walk's `reference` is never itself a float. Traced both call sites (`FloatBoxLeft`/`FloatBoxRight`) to confirm.

No existing or new test directly exercises plain same-side float-vs-float collision (two `float:left` siblings avoiding each other), so I'm noting that as a pre-existing coverage gap worth a follow-up, not a blocker for this PR — the change itself doesn't touch that logic and float-filtered suite (93 tests) passes.

PR #946, opened 2026-09-08
Re-resolve bidi when a running element's page counter changes length

dchupp:
**Regression from #936, which I authored — apologies. A running footer with a page counter crashes any document of ten or more pages.**

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at PeachPDF.Html.Core.Dom.CssBox.AppendWordsFromText(String sourceText)
   at PeachPDF.Html.Core.Dom.CssBox.ParseToWords()
   at PeachPDF.Html.Core.Dom.RunningElementLayout.RefreshPageCounterContent(...)
   at PeachPDF.Html.Core.Dom.RunningElementLayout.LayoutRunningElementFor(...)
   at PeachPDF.Html.Core.HtmlContainerInt.LayoutMarginBoxes(...)
```

## Repro

Twelve lines, no fixture files:

```html
@page { size: a6; margin: 12mm; }
@page { @bottom-center { content: element(foot); font-size: 8pt; } }
#foot { position: running(foot); margin: 0; }
#foot::before { content: counter(page); }
</style></head><body>
<div id="foot"></div>
<!-- enough paragraphs to reach page 10 -->
</body></html>
```

Under ten pages it renders. At ten it throws.

## Cause

`RefreshPageCounterContent` calls `ApplyContent` and then `ParseToWords`, and skips the step between them that the other two content-re-resolution paths take.

`BidiLevels`, `CharScripts` and `JoiningForms` are indexed against the text the **last** resolution saw. A page counter changes its own length the moment it gains a digit — `"9"` becomes `"10"` — so `AppendWordsFromText` indexes past the end of a stale array.

`HtmlContainerInt.ReapplyPseudoElementContent` already carries the fix, and its comment says exactly why:

> Re-parse words after content changes — re-resolve bidi levels for the new text first […] since `childBox.BidiLevels` would otherwise still reflect whatever (possibly now-stale) content the last resolution saw.

`ResolveTargetPageContent` does the same. This adds the same `CssBidiParagraphResolver.ResolveOwnTextAsParagraph(box)` call to the third path.

## Why the tests I added with #936 missed it

All five run a document to about seven pages, where the counter is one digit from first to last and the stale array is coincidentally the right length. The defect needs page ten. A "Page N of M" feature tested only on single-digit documents is not tested, and I should have seen that when I wrote them.

`RunningElementPageCounterTests.CounterPage_PastTheFirstTwoDigitPage_DoesNotThrow` now runs to 23 pages and asserts the full `1..N` sequence. Removing the one added line reproduces the exception.

## How it was found

Rendering a real 26-document corpus through this engine, where one document died that pristine `main` had rendered. Bisecting the two files #936 touched isolated it in one step. The public repro above came afterwards — nothing customer-specific is needed to trigger it.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,253 passed, 0 failed, 9 skipped.

### DIFF
--- a/.claude/recent-fixes/2026-09-08-a-page-counter-that-gains-a-digit-restales-its-bidi-levels.md
+++ b/.claude/recent-fixes/2026-09-08-a-page-counter-that-gains-a-digit-restales-its-bidi-levels.md
@@ -1,0 +1,34 @@
+# A page counter that gains a digit re-stales its bidi levels
+
+`RunningElementLayout.RefreshPageCounterContent` re-resolves a running element's `content` per page
+so `counter(page)`/`counter(pages)` read the page they are drawn on. It called `ApplyContent` then
+`ParseToWords` — and skipped the step between them that the other two content-re-resolution paths
+already take.
+
+`BidiLevels`, `CharScripts` and `JoiningForms` are indexed against the text the LAST resolution saw.
+A page counter changes its own length the moment it gains a digit — `"9"` becomes `"10"` — so
+`AppendWordsFromText` walked past the end of a stale array and threw `IndexOutOfRangeException`.
+Any document with a running footer showing a page counter and **ten or more pages** crashed.
+
+The fix is the one line `HtmlContainerInt.ReapplyPseudoElementContent` and `ResolveTargetPageContent`
+both already carry, with the comment on the first one saying exactly why:
+`CssBidiParagraphResolver.ResolveOwnTextAsParagraph(box)`.
+
+## What running it turned up
+
+- **The bug shipped because every fixture stopped at seven pages.** All five of the original
+  fixtures run a document to about seven pages, where the counter is one digit from first to last
+  and the stale array is coincidentally the right length. The defect needs page ten. A "Page N of M"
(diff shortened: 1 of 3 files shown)


jhaygood86 (approved):
Verified end-to-end, not just by reading the diff: reproduced the crash myself on `main` with a throwaway test using this PR's exact repro (running footer with `counter(page)`, 200 paragraphs to cross 10 pages) — got the identical `IndexOutOfRangeException` at `CssBox.AppendWordsFromText`. Switched to this branch and reran the same test — passes cleanly, footer text correct across all 10 pages. Confirmed the fix genuinely matches the existing `ReapplyPseudoElementContent`/`ResolveTargetPageContent` pattern (`ApplyContent → ResolveOwnTextAsParagraph → ParseToWords`), not just something superficially similar. All five of #936's own tests still pass, so this doesn't trade one regression for another.

Thanks for the quick turnaround and the apology, but this is exactly the kind of thing that's easy to miss when the discriminating test needs to cross a specific page-count threshold — no fault in the original review either.

PR #948, opened 2026-09-08
Report text that was drawn and then clipped away

### LINKED ISSUE #947: Report text that was drawn and then clipped away — the one loss class a reader cannot detect
There is one way a PeachPDF document can silently lose text that nothing outside the engine can detect: a word that is drawn and then clipped away.

I'd like to propose a small, detection-only API for it, and I have it implemented — but it adds public surface, so it seemed worth agreeing the shape before sending a PR rather than after.

## Why the engine has to be the one to report it

A glyph that was emitted and then truncated by the PDF clip operator is still **in the content stream**. So a reader that parses the content stream finds it and reports the document clean; a renderer that honours clipping cannot see it at all. Measured on a document with a table cell narrower than its content, using the two libraries as an oracle pair:

| | finds the clipped text? |
| --- | --- |
| PdfPig (parses the content stream) | yes — reports the document complete |
| MuPDF (honours the clip) | no — the text is not visible |

Both are behaving correctly. The disagreement is the point: no post-hoc reader can distinguish "drawn and clipped" from "drawn and visible", because the difference is not recorded in the stream. Only the engine, at the moment it decides to draw, knows the word's rect was larger than the clip it was drawn into.

For anyone generating documents from templates against arbitrary data — where a value can be one character longer than the column it lands in — that is the difference between shipping a correct invoice and shipping one with a truncated total that nothing flagged.

## Proposed shape

```csharp
public readonly record struct ClippedWord(
    string Text, int Page, double DrawnWidth, double VisibleWidth, /* … */)
{
    public double KeptFraction => /* VisibleWidth / DrawnWidth */;
}

public sealed class ClipReport
{
    public IReadOnlyList<ClippedWord> ClippedWords { get; }
    public int ClippedChars { get; }
}

// on the existing result type
public sealed class PeachPdfDocument
{
    public ClipReport ClipReport { get; internal set; }
}
```

Nothing about what is painted changes — this is purely a report.

## Where it goes

`FragmentPainter.Text.cs`'s `PaintWordSequence` already computes `var clip = g.GetClip(); clip.Intersect(wordFragment.Rect);`. The existing guard below it handles only the **full** case — a zero-area intersection, word never drawn — which a content-stream reader already catches as missing text. Partial clipping has no decision point at all today: the guard passes and `DrawWordGlyphs` emits the whole word.

So the check sits immediately after that guard, comparing two rects the method is already holding. No extra geometry work on the paint path.

The collector belongs on `HtmlContainerInt` rather than the painter, since a painter instance is per page and the report is per render — `PageClipOverride` on the same type is the existing precedent in the other direction. `PdfGenerator.AddPdfPages` drains it after the page loop, before the container is disposed.

Two methods would lose `static` (`PaintWordSequence`, `PaintLineWithEllipsis`) because they need the container.

## Deliberate exclusions

- **`text-overflow: ellipsis`** is not reported. That is truncation the author asked for and the reader can see, which is the opposite of the silent loss this exists for.
- **Whitespace** is not reported, for the same reason.
- **It under-reports, by design.** The clip compared against is `RGraphics`'s tracked rect stack; a `border-radius` or `clip-path` clip never reaches it, and `PdfGenerator`'s page-level `XGraphics.IntersectClip` is outside the stack entirely. "Nothing reported" is therefore weaker than "nothing was clipped", and the docs should say so.
- **Tolerance** would start at 0.5pt rather than the existing 1e-6 `VisibilityClipEpsilon`, which would report every word sitting flush against a clip edge. The report carries full geometry (`DrawnWidth`, `VisibleWidth`, `KeptFraction`) so a consumer can pick its own threshold, and so the default can be settled by measurement rather than taste.

## Evidence it is a signal rather than noise

Across a 26-document set of real templates: **0 documents report any clipped text**, and none of them move. That is the result that makes it usable — it fires on the control case (a single word in a 20px `overflow: hidden` box) while the content-stream detector reports that same document clean, which is the pair that shows it measures something otherwise invisible.

One thing measuring it corrected: a *multi-word* run in that box is not a clean partial clip. The later words fall entirely outside, hit the existing full-clip guard, are never drawn, and a content-stream reader reports them missing — correctly. Full-cull and partial-clip are different findings, and my first fixture conflated them.

Happy to send the PR if the shape looks right, or to adjust it — in particular whether `ClipReport` belongs on `PeachPdfDocument` or somewhere else, and whether the tolerance should be configurable rather than a constant.

dchupp:
Implements the proposal in #947 — detection only, nothing about what is painted changes.

Closes #947

## The problem

A glyph emitted and then truncated by the PDF clip operator is still **in the content stream**. A reader that parses that stream finds it and reports the document complete; a renderer that honours the clip shows only part of the word. Both are correct — the difference is not recorded in the file.

Measured with the two as an oracle pair on a table cell narrower than its content: PdfPig reads every character back and reports the document complete, while MuPDF cannot find the text at all. So the engine, at the moment it decides to draw, is the only thing that can report it.

## The API

```csharp
var document = await generator.GeneratePdf(html, pdfConfig);

foreach (var word in document.ClipReport.ClippedWords)
{
    Console.WriteLine($"'{word.Text}' kept {word.KeptFraction:P0} " +
                      $"({word.VisibleWidth:F1}pt of {word.DrawnWidth:F1}pt)");
}
```

`ClippedWord` carries geometry rather than a verdict — `DrawnWidth`, `VisibleWidth`, `DrawnHeight`, `VisibleHeight`, `KeptFraction` — so a consumer picks its own threshold and the default can be settled by measurement. `ClipReport.ClippedChars` totals the characters carried by truncated words.

## Where it sits

`FragmentPainter.Text.cs`'s `PaintWordSequence` already computes `var clip = g.GetClip(); clip.Intersect(wordFragment.Rect);`. The guard below it handles only the **full** case — zero-area intersection, word never drawn — which reading the output back already catches as missing text. Partial clipping had no decision point at all. The check is immediately after that guard, comparing two rects the method was already holding: **no extra geometry work on the paint path**.

The collector is on `HtmlContainerInt`, not the painter, because a painter instance is per page and the report is per render — `PageClipOverride` on the same type is the precedent in the other direction. `AddPdfPages` drains it after the page loop, before the container is disposed.

Two methods lose `static` (`PaintWordSequence`, `PaintLineWithEllipsis`) because they now need the container. Nothing else changes.

## Deliberate exclusions, all documented in `docs/usage-examples.md`

- **`text-overflow: ellipsis`** — truncation the author asked for and the reader can see, the opposite of the silent loss this exists for.
- **Whitespace** — a clipped space is not something anyone can see or act on.
- **It under-reports.** The clip measured against is `RGraphics`'s tracked rect stack; a `border-radius`/`clip-path` clip and the page-level `XGraphics.IntersectClip` never reach it. An empty report is a weaker statement than "nothing was clipped", and both the XML docs and the reader-facing docs say so.
- **Tolerance is 0.5pt**, a starting point rather than a result — the existing `VisibilityClipEpsilon` is 1e-6 and would report every word flush against a clip edge. Happy to make it configurable if you'd prefer.

## Tests

`ClipReportTests`, six fixtures. Disabling the report site fails two of them:

- the single-word control (reported, with `VisibleWidth < DrawnWidth`);
- a **multi-word** contrast — worth calling out, because measuring corrected my first fixture here. A multi-word run in the same box is not a clean partial clip: the later words fall entirely outside, hit the existing full-clip guard, are never drawn, and reading the output back reports them missing, correctly. Full-cull and partial-clip are different findings;
- an ordinary paragraph-and-table document, which must be **silent** or the report is noise rather than signal;
- `text-overflow: ellipsis`, excluded;
- `KeptFraction` bounded and below 1;
- a wide-enough `overflow: hidden` box, silent — the property alone must not report.

Across 26 real templates: **0 documents report any clipped text and none of them move**, which is what makes it usable as a signal.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,258 passed, 0 failed, 9 skipped. Diff coverage 100% on the changed lines.

### DIFF
--- a/.claude/recent-fixes/2026-09-08-report-text-that-was-drawn-and-then-clipped.md
+++ b/.claude/recent-fixes/2026-09-08-report-text-that-was-drawn-and-then-clipped.md
@@ -1,0 +1,66 @@
+# The engine reports text it drew and then clipped away
+
+Detection only — nothing about what is painted changes.
+
+A glyph emitted and then truncated by the PDF clip operator is still IN the content stream, so a
+reader that parses the stream finds it and reports the document complete, while a renderer that
+honours the clip shows only part of the word. Both are correct; the difference is not recorded in
+the file. Only the painter, at the moment it decides to draw, knows the word's rect was wider than
+the clip it was drawn into — so the engine is the only thing that can report it.
+
+## Where, and why there
+
+`FragmentPainter.Text.cs`'s `PaintWordSequence` already computes
+`var clip = g.GetClip(); clip.Intersect(wordFragment.Rect);`. The guard below it handles only the
+FULL case — a zero-area intersection, word never drawn — which reading the output back already
+catches as missing text. Partial clipping had no decision point at all. The check therefore sits
+immediately after that guard, comparing two rects the method was already holding: no extra geometry
+work on the paint path.
+
+The collector is on `HtmlContainerInt`, not the painter, because a painter instance is per page and
+the report is per render — `PageClipOverride` on the same type is the precedent in the other
+direction. `PdfGenerator.AddPdfPages` drains it after the page loop, before the container is
+disposed, onto the public `PeachPdfDocument.ClipReport`. Two methods lose `static`
+(`PaintWordSequence`, `PaintLineWithEllipsis`) because they now need the container.
+
+## What measuring it turned up
+
+- **A multi-word run in a narrow box is not a clean partial clip**, and the first fixture conflated
+  two findings. The later words fall entirely outside, hit the existing full-clip guard, are never
+  drawn, and reading the output back reports them missing — correctly. Full-cull and partial-clip
+  are different results, and the control case has to be a SINGLE word whose rect starts inside the
+  box.
+- **It is quiet on ordinary documents.** A plain paragraph and a plain table report nothing on
+  upstream, which is what makes the rate readable. Across 26 real templates: 0 documents report any
+  clipped text and none of them move — as expected for a detection-only change.
+- **The tolerance (0.5pt) is a starting point, not a result.** The existing `VisibilityClipEpsilon`
+  is 1e-6 and would report every word flush against a clip edge. The report carries `DrawnWidth`,
+  `VisibleWidth` and `KeptFraction` precisely so the threshold can be settled by measurement.
+
+- **The report accumulates, and had to be made to.** `AddPdfPages` is a repeatable public API —
+  a caller appends pages to an existing document across several calls, each building its own
+  container and its own report — so assigning `document.ClipReport` wholesale discarded every
+  earlier call's findings. `PeachPdfDocument.PageCount` accumulates across those same calls via the
+  underlying `PdfDocument`; a report whose whole purpose is not to lose information quietly should
+  not be the one member that does.
+- **A word can be reported with `VisibleWidth == DrawnWidth`.** The record fires when EITHER
+  dimension was reduced past the tolerance, so a box short enough to cut a line's height but wide
+  enough to keep the whole word is recorded — correctly, it is the same silent loss — with its width
+  untouched. The first version of the XML doc said "always less than `DrawnWidth`", which is wrong
+  for exactly that case.
+
+## Deliberately not reported
+
+`text-overflow: ellipsis` (truncation the author asked for and the reader can see) and whitespace
+(a clipped space is not actionable). And the report UNDER-reports: it measures against `RGraphics`'s
+tracked rect stack, which a `border-radius`/`clip-path` clip and the page-level
+`XGraphics.IntersectClip` never reach. "Nothing reported" is weaker than "nothing was clipped", and
+both the API docs and the reader-facing docs say so.
+
+## Evidence
+
+`ClipReportTests`, six fixtures: the single-word control (reported, with geometry), the multi-word
+contrast, an ordinary document (silent), `text-overflow: ellipsis` (excluded), `KeptFraction`
+bounded and below 1, and a wide-enough `overflow: hidden` box (silent — the property alone must not
+report). Disabling the report site fails two of them. Full suite green on net8.0 (10,258), 0 new
+build warnings.
--- a/docs/usage-examples.md
+++ b/docs/usage-examples.md
@@ -19,6 +19,7 @@ using PeachPDF.Network;
 - [Rendering in the browser (Blazor WebAssembly)](#rendering-in-the-browser-blazor-webassembly)
 - [Sharing a parsed CSS context across renders](#sharing-a-parsed-css-context-across-renders)
 - [Saving a PDF to a file](#saving-a-pdf-to-a-file)
+- [Detecting text that was clipped away](#detecting-text-that-was-clipped-away)
 - [Fonts](#fonts)
 - [Enabling tagged PDF (PDF/UA) output](#enabling-tagged-pdf-pdfua-output)
 - [Enabling interactive PDF forms](#enabling-interactive-pdf-forms)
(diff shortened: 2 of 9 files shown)


jhaygood86:
First-look review. Build is clean (`dotnet build -t:Rebuild`: 0 warnings), the 6 `ClipReportTests` pass, and the paint-architecture question checked out — `PerformPaint` has exactly one call site (`PdfGenerator.cs`, once per page in the final page loop), so `RecordIfClipped` can't double-count across a measurement/dry-run pass, and it reads only fragment-tree-derived rects (`wordFragment.Rect`, the already-intersected clip) plus text-content fields off `CssRect` — no geometry read off `CssBox`. The `HtmlContainerInt.ClipReport` collector placement matches the `PageClipOverride` precedent as claimed. Docs (`usage-examples.md`) match the actual API shape exactly (member names/types verified against `ClippedWord.cs`/`PeachPdfDocument.cs`).

Two things worth fixing before merge:

**1. `document.ClipReport` is silently overwritten, not accumulated, across repeated `AddPdfPages` calls.**

`PdfGenerator.cs:551`:
```csharp
document.ClipReport = container.HtmlContainerInt.ClipReport;
```
`AddPdfPages` is documented in this same method (see the PDF/A-conformance comment a few lines above) as a *repeatable* public API — callers append more pages to an existing `PeachPdfDocument` across multiple calls. Each call builds a fresh `HtmlContainer`/`ClipReport`, and this line replaces `document.ClipReport` wholesale rather than merging into it, so a second `AddPdfPages` call silently discards any clipped-word findings from the first. That's inconsistent with the sibling `PeachPdfDocument.PageCount` property right above it in the same file, which naturally accumulates across calls via the underlying `PdfDocument`. For a report whose entire value proposition is "don't silently lose information," silently losing report entries on the documented multi-call path seems like the wrong default — this should append to the existing words list instead of replacing it.

**2. `ClippedWord.VisibleWidth`'s XML doc is inaccurate.**

`ClippedWord.cs`:
```csharp
/// <param name="VisibleWidth">Width that survived the clip. Always less than <paramref name="DrawnWidth"/>.</param>
```
`RecordIfClipped` (`FragmentPainter.Text.cs`) records a word when *either* its width or its height was reduced by more than the 0.5pt tolerance — not only when width was reduced:
```csharp
if (visible.Width >= drawn.Width - clippedTolerance
    && visible.Height >= drawn.Height - clippedTolerance)
{
    return;
}
```
So a word clipped only vertically (e.g. a box short enough to clip line height but wide enough to keep the full word) gets recorded with `VisibleWidth == DrawnWidth`, contradicting the "always less than" doc comment. Worth fixing the comment (or scoping it to "less than, when width was the clipped dimension").

Minor, non-blocking: `ClipReportTests` verify direction/range (`VisibleWidth < DrawnWidth`, `0 <= KeptFraction < 1`) rather than hand-computed expected widths/fractions. Given glyph metrics aren't trivial to hand-compute, that's a defensible tradeoff, but it does mean the fixtures prove "something was reported," not "the reported geometry is numerically correct" — worth knowing as a gap if `KeptFraction`'s exact math ever regresses.

Showcase: no `TestHarness/Program.cs` change, but given this is explicitly detection-only ("nothing about what is painted changes" — the PDF output is byte-identical to before), there's nothing a showcase card's thumbnail/PDF would show differently. I'd treat this as reasonably exempt from the showcase convention rather than a gap.

dchupp:
Both fixed, force-pushed. Point 1 in particular is a good catch — it is precisely the failure mode this feature exists to prevent, sitting inside the feature.

**1. The report now accumulates across `AddPdfPages` calls.**

```csharp
document.ClipReport.Append(container.HtmlContainerInt.ClipReport);
```

You are right that assigning it wholesale discarded every earlier call's findings, and right that `PageCount` accumulating across the same calls is the standard to match. A report whose whole point is not to lose information quietly should not be the one member that does.

`RepeatedAddPdfPagesCallsAccumulate` pins it: generate a clipping document, assert one word, `AddPdfPages` the same content again, assert two words and 16 characters. Reverting to the assignment gives `Expected: 2, Actual: 1`.

**2. `VisibleWidth`'s doc is corrected, and now has a fixture behind it.**

You are right that `RecordIfClipped` fires when *either* dimension was reduced, so a vertically-clipped word is recorded with `VisibleWidth == DrawnWidth`. The doc now says so explicitly rather than claiming "always less than", and `AWordClippedOnlyVerticallyIsAlsoReported` asserts exactly that shape — a `width: 400px; height: 4px; overflow: hidden` box keeps the whole word horizontally while cutting its height:

```csharp
Assert.Equal(word.DrawnWidth, word.VisibleWidth, 1);
Assert.True(word.VisibleHeight < word.DrawnHeight);
```

That also answers your minor note in the direction it can be answered: it is a fixture asserting a specific geometric *relationship* the implementation must hold, rather than another "something was reported" check. You are right that the absolute widths remain unverified against hand-computed glyph metrics — I have not found a way to do that which would not just be re-implementing the shaper in the test.

I have also carried both facts into `docs/usage-examples.md`, since the reader-facing text described the geometry without either caveat: it now says a word clipped only vertically comes back with `VisibleWidth` equal to `DrawnWidth`, and that the report accumulates across repeated `AddPdfPages` calls.

**On the showcase exemption** — agreed, and thank you for reasoning it through rather than just applying the convention. The PDF bytes are unchanged by construction, so a showcase card would be identical to an existing one.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,260 passed, 0 failed, 9 skipped.

jhaygood86 (approved):
Verified both fixes directly. `document.ClipReport.Append(...)` replaces the wholesale assignment, and `RepeatedAddPdfPagesCallsAccumulate` pins exactly the scenario that motivated it. `VisibleWidth`'s doc comment and the new `AWordClippedOnlyVerticallyIsAlsoReported` fixture correctly assert the relationship (`VisibleWidth == DrawnWidth` when only height was clipped) rather than another "something was reported" check — good response to that note.

Build clean, clip-filtered suite (162 tests) passes, all three CI platforms green.

Approving. Merging now.

PR #949, opened 2026-09-08
Share a widened table's extra space in proportion to its columns

dchupp:
`DetermineMissingColumnWidths` has two surplus clauses, and they disagree about what CSS 2.1 §17.5.2.2's *"the extra width should be distributed over the columns"* means.

The clause for a table whose columns **all declared a width** already spreads proportionally, via `SpreadSurplusProportionally`. The clause for a table whose columns are **all `auto`** divides the surplus by the column count and adds the same absolute amount to each — and that is by far the more common table.

An equal absolute share hands the same number of points to a 250pt description column and a 30pt quantity column, so widening a table to its declared width does not produce the same table, only a wider one.

## Measured against Chrome

Three columns at `width: 500pt` — a long description column plus `Qty` and `N`:

| | description | Qty | N |
| --- | --- | --- | --- |
| **Chrome 152** | **451.2** | **32.9** | **15.9** |
| proportional (this change) | 447.3 | 36.1 | 16.7 |
| equal share (`main`) | 299.7 | 104.7 | 95.5 |

Rendered headless and read with `pdftotext -bbox`. The description column loses ~150pt under an equal share, to two columns holding three characters between them.

## One thing worth knowing about the old behaviour

`main` was not doing a *pure* equal share, so the old code reads simpler than it behaves. Measured bumps on that fixture were 112.5 / 75.2 / 75.2 — the two small columns take identical absolute shares while the large one takes more, because the max-content capping loop above runs first and removes some columns from the split. The result still lands much closer to equal than to Chrome.

## Scope

- An explicit `max-width` still caps a column — `GetColumnExplicitMaxWidths` is unchanged, only the size of each share differs.
- A column that cannot absorb its share leaves the table slightly under-filled. That is the pre-existing behaviour and is deliberately not changed here.
- A column measuring nothing at all takes an equal share, there being no proportion to go on.

## Tests

`TableSurplusDistributionTests`, four fixtures, three of which fail against `main`:

- each column's **share** of the finished table matches Chrome's 90.2% / 6.6% / 3.2%;
- each column's bump is ordered by its own size, rather than flat;
- the wide column still holds ~90% of the table;
- an explicit `max-width` still caps.

Compared as shares rather than absolute points, because the two engines disagree about font metrics and therefore about the natural widths the shares are taken of — the shares are what the rule actually states. (My first version computed the expected values from a separate `width: auto` render and was out by 7pt, because that capping loop has already moved `_columnWidths` by the time this clause runs.)

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-widened-tables-extra-space-is-shared-in-proportion.md
+++ b/.claude/migration-notes/2026-09-08-a-widened-tables-extra-space-is-shared-in-proportion.md
@@ -1,0 +1,23 @@
+# A widened table's extra space is shared in proportion to the columns, not equally
+
+A table with a declared `width` whose columns are all `auto` has surplus space to distribute once
+each column has been resolved to its own max-content width. That surplus was handed out as an equal
+number of **points per column** — the same absolute amount to a 250pt description column and a 30pt
+quantity column — so widening a table to `width: 100%` did not produce the same table, only wider.
+
+The surplus is now shared **in proportion to what each column already measures**, which is what
+browsers do and what the neighbouring all-columns-specified clause already did.
+
+**What a document author may notice.** Column boundaries move in any auto-layout table that declares
+a width wider than its content — generally toward what a browser produces. On a three-column
+fixture at 500pt (a long description column plus `Qty` and `N`):
+
+| | description | Qty | N |
+| --- | --- | --- | --- |
+| Chrome 152 | 451.2 | 32.9 | 15.9 |
+| after this change | ~90% of the table | | |
+| before (equal share) | 299.7 | 104.7 | 95.5 |
+
+An explicit `max-width` still caps a column, exactly as before.
+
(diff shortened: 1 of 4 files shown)


jhaygood86 (approved):
Verified: `DetermineMissingColumnWidths`'s auto-column surplus clause now reuses the same proportional-distribution approach as the declared-width clause and `CalculateFixedColumnWidths`, consistent with CSS 2.1 §17.5.2.2 (which only says "should be distributed," leaving the algorithm open — proportional-by-content is a defensible, Chrome-aligned choice matching this repo's existing convention for underspecified cases). Checked out `main`'s pre-fix code and reran the new tests directly: 3 of 4 fail, with the reported bump ratio (112.5 vs 74.7) matching exactly — confirms the "not a pure equal share, but closer to equal than proportional" characterization of the old behavior is accurate. No conflict with #938 or #950 (different methods, ~200 lines apart, logically orthogonal).

Issue #898, opened 2026-09-06
A `calc()` width on a `<table>` is ignored, so the table lays out shrink-to-fit and can overrun its container


dchupp:
**Version:** v0.9.15. Still present on `main` (checked at 03bb8c2, one release ahead) — both
guards below are unchanged there.

### What happens

A `<table>` whose `width` is a `calc()` expression ignores that width entirely. The table
falls back to its containing block's width and never records that a width was specified, so it
lays out **shrink-to-fit** — which can be *wider* than the container, and the table overruns
it.

Absolute forms are affected as much as percentage ones; the common of the two in real markup
is `calc(100% - <length>)`. Every form works correctly on a `<div>`, so this is specific to
table sizing.

### Why

`CssLayoutEngineTable.GetAvailableTableWidth` gates on

```csharp
CssLength tableBoxLength = new(inlineSizeCss);
if (!(tableBoxLength.Number > 0)) return containingBlockInlineSize;
```

`CssLength` parses a bare `<number><unit>` and reports `Number == 0` for anything it does not
recognise, `calc()` included. So the method returns the containing-block width and never
reaches the `CssValueParser.ParseLength(...)` on its very next line — which resolves `calc()`
perfectly well. `_widthSpecified` is left false as a side effect, which is what selects
shrink-to-fit.

`GetMaxTableWidth` carries the identical test against `max-width` and discards a `calc()`
max-width the same way.

### Suggested fix

Gate on `CssValueParser.IsValidLength`, which does recognise `calc()`, keeping the old
`CssLength` test as a fallback so nothing that worked before stops working:

```csharp
if (!CssValueParser.IsValidLength(inlineSizeCss)
    && !(new CssLength(inlineSizeCss).Number > 0))
{
    return containingBlockInlineSize;
}
```

and the mirror of it in `GetMaxTableWidth`. The `ParseLength` call each method already makes
needs no change.

### Reproduction and measurements

A table inside a fixed 200pt frame:

| declaration | before | after |
| --- | --- | --- |
| `width: 100%` | 200pt | 200pt |
| `margin-left: 12px; width: calc(100% - 12px)` | 200pt, overruns the frame | 191pt, inside it |
| `width: calc(200pt - 20pt)` | 180pt | 180pt |

191pt is exactly 200pt less 12px (12px = 9pt at 96 CSS dpi), so the resolved value is right,
not merely different.

[...]

PR #950, opened 2026-09-08
A declared cell width is its content width, so add its padding and border

dchupp:
`_columnWidths` holds **outer** widths — `GetColumnMinWidths` fills the same array from `cell.GetMinimumWidth()`, which includes padding and border. The clause that reads a cell's declared `width` stored the bare parsed length instead, which under `box-sizing: content-box` (the default) is a **content** width.

So the array mixed two meanings, and every explicitly sized column came out narrower than a browser's by exactly its padding plus border — with the shortfall handed to whichever column was `auto`. On a table of seven px-width columns with `padding: 3px 5px`, each is about 8pt narrow and the auto column absorbs roughly 58pt it should not have.

## Measured against Chrome

Three shapes, rendered headless and read with `pdftotext -bbox`, first column's width in points:

| cell | Chrome 152 | `main` | this change |
| --- | --- | --- | --- |
| `width:100px; padding:3px 5px; border:1px` | **84.75** | 75.0 | 75.0 + its own padding and border |
| `width:100px` | 77.25 | 75.0 | 75.0 (unchanged) |
| the first, plus `box-sizing:border-box` | 76.5 | 75.0 | 75.0 (unchanged) |

## It reads `box-sizing` rather than always adding

Worth calling out, because the obvious fix is wrong for the third row. Under `border-box` the declared width already covers padding and border, and Chrome puts that shape at the *unpadded* width — always-adding would trade one wrong number for another.

`CssBox.ActualBoxSizeIncludedWidth` already makes exactly this distinction (padding + border for content-box, zero for border-box), so the new helper is only its writing-mode switch, mirroring `CellInlineSize`'s:

```csharp
private double CellInlineSizeExcludes(CssBox cell) =>
    _isVertical ? cell.ActualBoxSizeIncludedHeight : cell.ActualBoxSizeIncludedWidth;
```

## Tests

`TableDeclaredCellWidthTests`, three fixtures:

- a padded cell's column is wider than an unpadded one declaring the same width — **fails against `main`**;
- `box-sizing: border-box` does not widen it, which is the contrast case that justifies reading the property;
- an unpadded `width: 100px` cell is still exactly 75pt, so the change is a no-op where nothing is owed.

Asserted as relationships between the three rather than as absolute points, since the two engines disagree about font metrics and border-collapse rounding.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,255 passed, 0 failed, 9 skipped.

---

This is the second of three fixes that were carried together in our tree as one "table column widths" patch. The first — `calc()` on a table — you already fixed via #901 after our issue #898. The third is #949 (proportional surplus). They touch the same file but are independent, so they are separate PRs; #949 does not conflict with this one.

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-declared-cell-width-now-adds-its-own-padding-and-border.md
+++ b/.claude/migration-notes/2026-09-08-a-declared-cell-width-now-adds-its-own-padding-and-border.md
@@ -1,0 +1,25 @@
+# A declared cell width now adds its own padding and border to the column
+
+`box-sizing` defaults to `content-box`, so a table cell's declared `width` is its **content** width
+and its padding and border sit outside it. The engine stored that bare content width as the column's
+width — but a column width is an OUTER width everywhere else, since the same array is filled from
+`cell.GetMinimumWidth()`, which already includes padding and border.
+
+Every explicitly sized column therefore came out narrower than a browser's by exactly its padding
+plus border, and the difference was handed to whichever column was `auto`. On a table of seven
+px-width columns with `padding: 3px 5px`, each was about 8pt narrow and the auto column absorbed
+roughly 58pt it should not have had.
+
+Such a column is now as wide as the browser draws it. Measured against Chrome 152 on a cell
+declaring `width: 100px; padding: 3px 5px; border: 1px solid`:
+
+| | first column |
+| --- | --- |
+| Chrome 152 | 84.75pt |
+| before | the same as an unpadded cell — 75pt |
+| after | 75pt of content plus its own padding and border |
+
+A cell declaring `box-sizing: border-box` is **unaffected** — its declared width already covers
+padding and border, and Chrome puts that shape at the unpadded width too.
+
+See [Tables](../../docs/html-css-support.md#tables) in `docs/html-css-support.md`.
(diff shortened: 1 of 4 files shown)


jhaygood86 (approved):
Verified: `CellInlineSizeExcludes` correctly delegates to `CssBox.ActualBoxSizeIncludedWidth`, which returns padding+border for `content-box` and 0 for `border-box` — confirmed by reading the property directly, not just the call site. Hand-computed the Chrome-comparison numbers myself: 75.0pt content + 7.5pt padding + 1.5pt border = 84.0pt against Chrome's 84.75pt (0.75pt off, consistent with the font-metric/border-collapse rounding disagreement your tests already disclaim). `BorderBoxDoesNotWidenTheColumn` is exactly the fixture that would catch a reversed condition.

Also noted for a possible follow-up (not blocking this PR): `CalculateFixedColumnWidths` (the `table-layout: fixed` path) has the identical gap — it reads `CellInlineSize` without the `CellInlineSizeExcludes` addition either. Worth a tracked issue if you want to pick it up, since this PR's own title scopes it to auto layout.

Build clean, all three CI platforms green, no conflict with #949 (different methods, ~200 lines apart).

Approving. Merging now.

PR #951, opened 2026-09-08
An inline-block reserves its declared width on the line

dchupp:
CSS 2.1 §10.3.9: an atomic inline-level box occupies its own **used width** on the line, not the width its content happened to measure. The inline flow accumulates word widths, so `width` on a `display: inline-block` is inert today and whatever follows sits flush against the box's text.

Two shapes it breaks, both common:

```html
<!-- a fixed-width label: reserves only the width of "Hi", so a column of these stops lining up -->
<span style="display:inline-block;width:160px">Hi</span><span>AFTER</span>

<!-- a checkbox glyph: no content at all, so it takes no room whatsoever -->
<span style="display:inline-block;width:120px;border:1px solid"></span><span>AFTER</span>
```

## Measured against Chrome

How far `AFTER` is pushed along the line, in points past the container's edge:

| | Chrome 152 | `main` | this change |
| --- | --- | --- | --- |
| `width: 160px` (=120pt), content `Hi` | **120.0** | the width of "Hi" | 120.0 |
| `width: 120px` (=90pt), empty, 1px border | **91.5** | 0 | 91.5 |

Rendered headless and read with `pdftotext -bbox`. Both now agree with Chrome to within 0.1pt.

## Why it does not just reuse `b.ClientRight`

The flex/grid path a few hundred lines up already does exactly this, with `coordinates.CurrentX = b.ClientRight`, and reusing that was my first instinct. It cannot work here: on the plain inline path the child box's geometry is never assigned at all. Measured on a `width: 160px` inline-block at that point, `Location.X`, `Size.Width`, `ClientRight` and `ActualRight` are **all zero**.

So the declared width is resolved from the CSS at the advance point, against a `childContentStartX` captured after every branch that can establish the child's position on the line — including the left-float branch, which re-establishes it from scratch rather than adjusting it.

## Scope

- **Only ever forward.** `Math.Max`, so content wider than the declared width overflows rather than being pulled back — which is what `overflow: visible` means.
- **Percentages are left alone**, because a percentage resolves against a containing block this line does not know. There is a fixture asserting that, so it reads as a decision rather than an oversight.

## Tests

`InlineBlockDeclaredWidthTests`, four fixtures, the first two failing against `main`:

- the declared width reserved (Chrome's 120pt);
- the empty bordered box (Chrome's 91.5pt) — the case an accumulate-the-words flow gets most wrong;
- content wider than the declared width still overflowing, measured against the same markup with `width: auto`;
- a percentage width unaffected.

Each measures the *advance* — the difference between where the following text sits with the box and without it — rather than an absolute coordinate, so the assertions do not depend on the harness's own margins.

### DIFF
--- a/.claude/migration-notes/2026-09-08-width-on-an-inline-block-now-occupies-that-width-on-the-line.md
+++ b/.claude/migration-notes/2026-09-08-width-on-an-inline-block-now-occupies-that-width-on-the-line.md
@@ -1,0 +1,22 @@
+# `width` on a `display: inline-block` now occupies that width on the line
+
+An atomic inline-level box occupies its own used width on the line (CSS 2.1 §10.3.9). The inline
+flow accumulated its content's word widths instead, so a `width` on a `display: inline-block` had no
+effect at all and whatever followed sat flush against the box's text.
+
+Two shapes it broke, both common:
+
+- **A fixed-width label.** `<span style="display:inline-block;width:160px">Hi</span>` reserved only
+  the width of "Hi", so a column of such labels stopped lining its values up.
+- **An empty bordered box used as a glyph** — a checkbox drawn as
+  `<span style="display:inline-block;width:120px;border:1px solid"></span>` — has no content at all,
+  so it took no room whatsoever.
+
+Such a box now reserves its declared width. Measured against Chrome 152 on the same shapes: the
+first puts the following text exactly 120pt past the container's edge (160px at 96 CSS dpi), the
+second at 91.5pt (90pt plus its 1px borders). PeachPDF now agrees with both to within 0.1pt.
+
+Content **wider** than the declared width still overflows rather than being pulled back, which is
+what `overflow: visible` means, and a **percentage** width is unaffected.
+
+See [Display](../../docs/html-css-support.md#display) in `docs/html-css-support.md`.
--- a/.claude/recent-fixes/2026-09-08-an-inline-block-reserves-its-declared-width-on-the-line.md
+++ b/.claude/recent-fixes/2026-09-08-an-inline-block-reserves-its-declared-width-on-the-line.md
@@ -1,0 +1,38 @@
+# An inline-block reserves its declared width on the line
+
+CSS 2.1 §10.3.9 makes an atomic inline-level box occupy its own used width on the line. `FlowBox`
+just accumulated word widths, so `width` on a `display: inline-block` was inert.
+
+The flex/grid path a few hundred lines up already does exactly this, via
+`coordinates.CurrentX = b.ClientRight`.
+
+## What running it turned up
+
+- **`ClientRight` cannot be reused here, which is why the fix looks different from its precedent.**
+  On the plain inline path the child box's geometry is never assigned at all — measured on a
+  `width: 160px` inline-block, `Location.X`, `Size.Width`, `ClientRight` and `ActualRight` are all
+  `0`. So the declared width has to be resolved from the CSS at the advance point, against a
+  `childContentStartX` captured after every branch that can establish the child's line position
+  (including the left-float branch, which re-establishes it from scratch).
+- **Chrome agrees to 0.1pt** on both shapes that matter: 120pt for `width: 160px` with short
+  content, 91.5pt for an empty `width: 120px` box with a 1px border. That second one is the case an
+  accumulate-the-words flow gets most wrong — no content means no room at all.
+- **`box-sizing: border-box` has to be taken off, and the obvious helper does the opposite.**
+  `childContentStartX` is the content-box start and `rightSpacing` is added after, so what belongs
+  between them is the CONTENT width. Under `border-box` the declared width already covers the
+  padding and border those two spacings re-add, so they are counted twice unless subtracted here.
+  `ActualBoxSizeIncludedWidth` cannot do it: it answers what a declared size does NOT include, so
+  it is padding+border for content-box and ZERO for border-box — a no-op in exactly the case that
+  needs adjusting, and wrong in the case that does not. Chrome puts a `border-box; width: 120px`
+  box with a 1px border at exactly 90pt; without the subtraction it came out 91.5pt.
+- **Percentages are deliberately excluded.** A percentage resolves against a containing block this
+  line does not know. There is a fixture asserting the exclusion so it reads as a decision rather
+  than an oversight.
+
+## Evidence
+
+`InlineBlockDeclaredWidthTests`, four fixtures: the declared width reserved (Chrome's 120pt), the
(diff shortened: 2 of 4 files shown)


jhaygood86:
Reviewed the diff, checked out the branch, and rebuilt/retested locally. Build is clean (0 warnings), the PR's own four fixtures pass, and a broader regression pass (`InlineBlock|Acid2|Table|Flex`, 1325+ tests) is all green.

The core reasoning holds up: `Location.X`/`Size.Width`/`ClientRight`/`ActualRight` genuinely are all `0` on the plain inline path at this point (confirmed by reading the surrounding code — `ActualBoxSizingWidth => Size.Width + ActualBoxSizeIncludedWidth`, and `Size.Width` is never assigned for an inline-block child before this code runs), so reusing the flex/grid `b.ClientRight` pattern really isn't available here. The `width: auto` case is safe too: `CssValueParser.IsValidLength("auto")` is `false`, so the new branch's guard skips it entirely and that path is untouched.

One real correctness gap, though: **`box-sizing: border-box` isn't handled, and double-counts the border/padding.**

```csharp
var declared = CssValueParser.ParseLength(b.Width, 0, b);
coordinates.CurrentX = Math.Max(coordinates.CurrentX, childContentStartX + declared);
```

`declared` is the *raw* parsed CSS length, treated as a pure content-box width. But elsewhere in this same file, the standard conversion from a declared `width` to a content width already exists and explicitly subtracts the box-sizing-included amount:

```csharp
// CssLayoutEngine.cs:1667-1672 (GetBoxWidth)
width = CssValueParser.ParseLength(box.Width, widthBasis, box);
...
width -= box.ActualBoxSizeIncludedWidth;   // 0 for content-box, border+padding for border-box
```

The new code skips that subtraction. For `box-sizing: border-box; width: 120px; border: 1px solid` (an inline-block with no content), the declared 120px (90pt) is already meant to be the *border-box* width — so the advance should be exactly 90pt. Instead, because `rightSpacing` (which includes the border) is still added afterward, the actual advance comes out to 91.5pt — the border is counted twice. I verified this empirically by adding a throwaway test on top of this branch mirroring the PR's own `AnEmptyBorderedInlineBlockStillTakesItsWidth` fixture but with `box-sizing: border-box` added:

```
Expected: 90 (rounded from 90)
Actual:   91.5 (rounded from 91.5)
```

None of the four fixtures in `InlineBlockDeclaredWidthTests` use `box-sizing: border-box`, so this ships silently. Given `box-sizing` is a real, commonly-used property in this codebase (see `css-properties.json`), this needs either a fix (subtract `b.ActualBoxSizeIncludedWidth` from `declared`, matching `GetBoxWidth`'s existing pattern) or an explicit accepted-gap note + tracked issue if it's meant to be follow-up work.

Separately, a smaller doc-comment issue: the new comment block and `if` are inserted between the pre-existing comment ("A box whose content was all placed in an earlier fragmentainer is not re-closed here either...") and the code it actually describes (the `if (coordinates.PlacedSince(childStartOrdinal)) { coordinates.CurrentX += rightSpacing; }` a few lines down). That comment is now detached from its target and sits above the new, unrelated width-reservation block instead — worth moving the new comment+code either above the old comment or after the `rightSpacing` block so each comment stays adjacent to what it describes.

No functional overlap with #952 — that PR only touches `GetBoxWidth` (lines ~1667-1687), this one only touches `FlowBox` (lines ~2537-2990) in the same file; different functions, no line overlap, safe to merge independently.

dchupp:
Both fixed, force-pushed. The `box-sizing` gap is real and I'm glad you caught it — though the suggested expression would not have closed it, and the reason is worth recording.

**`ActualBoxSizeIncludedWidth` answers the opposite question.** It is `padding + border` for **content-box** and **zero** for **border-box**:

```csharp
BoxSizingMode.ContentBox => ActualPaddingLeft + ActualPaddingRight + ActualBorderLeftWidth + ActualBorderRightWidth,
BoxSizingMode.BorderBox  => 0,
```

So `declared -= ActualBoxSizeIncludedWidth` would be a **no-op in exactly the border-box case you measured**, and would wrongly shrink the content-box case that is currently correct — it would trade the 91.5/90 error for a 120/111 one. It works in `GetBoxWidth` because there `width` starts as the containing block's available width, not as a declared length; the two are asking different questions of the same helper.

What is actually needed is the padding and border themselves, taken off only under border-box:

```csharp
var declared = CssValueParser.ParseLength(b.Width, 0, b);

if (b.BoxSizing.Value is BoxSizingMode.BorderBox)
{
    declared -= b.ActualPaddingLeft + b.ActualPaddingRight
                + b.ActualBorderLeftWidth + b.ActualBorderRightWidth;
}
```

`childContentStartX` is the content-box start and `rightSpacing` is added afterwards, so what belongs between them is the **content** width — which a declared width already is under content-box, and is not under border-box.

**Two fixtures, both matching Chrome exactly:**

| | Chrome 152 | before | now |
| --- | --- | --- | --- |
| `border-box; width:120px; border:1px` | 90.0 | 91.5 | 90.0 |
| `border-box; width:160px; padding:0 10px` | 120.0 | 135.0 | 120.0 |

I added the padding case as well as the border one, because a fix that only subtracts the border passes your repro and still fails the more common shape.

**Doc-comment detachment: fixed, and it is the third time.** The new block now sits above the "A box whose content was all placed in an earlier fragmentainer…" comment, which is directly on its own `if` again. Moving it turned up a second defect worth flagging: my first attempt at the move left **two copies** of the width-reservation block in the file. Both ran, `Math.Max` kept the larger, and the border-box fixtures still reported 91.5 — the fix looked ineffective when it was simply being overridden by its own stale twin. Worth knowing if you ever see a change appear to do nothing.

**On your minor note about the fixtures verifying direction rather than exact geometry** — that applied to #948's clip report; here the four new/updated fixtures all assert exact point values against Chrome (120.0, 91.5, 90.0, 120.0), measured as an *advance* (the difference between where the following text sits with the box and without it) so they do not depend on the harness's own margins.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,258 passed, 0 failed, 9 skipped.

jhaygood86 (approved):
Verified: the corrected fix reads `b.BoxSizing.Value is BoxSizingMode.BorderBox` and subtracts padding+border only in that case, matching what's actually needed rather than my suggested `ActualBoxSizeIncludedWidth` (which I now see would have been a no-op precisely where it mattered — thanks for catching that and explaining why rather than just applying the label I handed you). Both fixtures pass, and the doc comment now explicitly says why that property is the wrong tool here, which is worth having for the next person who reaches for it.

Build clean, inline-block-filtered suite (23 tests) passes, all three CI platforms green.

Approving. Merging now.

PR #952, opened 2026-09-08
A float's auto width is shrink-to-fit

dchupp:
CSS 2.1 §10.3.5 makes a floating box's `auto` width shrink-to-fit — `min(max-content, max(min-content, available))`. `GetBoxWidth` gives it the ordinary stretch-to-containing-block width instead, so an auto-width float fills its containing block.

That leaves `float: right` nowhere to go: it lands at the left as a full-width bar, and the content that belongs beside it is pushed onto its own line.

```html
<div style="width:300pt">
  <div style="float:right; border:1px solid">BADGE</div>
</div>
```

## Measured against Chrome

| | Chrome 152 | `main` | this change |
| --- | --- | --- | --- |
| auto `float: right`, offset into a 300pt container | **257.9pt** | 0 — full width at the left | at the right edge |
| text after an auto `float: left` | beside it | on its own line | beside it |
| `float: right; width: 60px` | 45pt | 45pt | 45pt (unchanged) |

Rendered headless and read with `pdftotext -bbox`.

## Placement was never the problem

Worth stating, because it is what makes this hard to see: a float with a **declared** width lands on a browser's x to the point, so the placement machinery looks correct under any test that declares one. Only the `auto` measurement was wrong — and `auto` is what a badge or a pull-quote actually uses. There is a fixture pinning the declared-width case at exactly 45pt so the change is visibly scoped.

## The formula is already in the tree

`min(max-content, max(min-content, available))` is what `GetFitContentWidth` and `GetMinContentWidth` compute between them, and the orthogonal-flow shrink in `CssBox.cs` already uses that pair for §10.3.5. This reuses it rather than adding a second implementation.

Both are outer widths, so they compare against the stretch width as it stands and the box's own decoration comes back off at the end — the same subtraction the plain `auto` branch makes, and a no-op under `box-sizing: border-box`.

## Tests

`FloatAutoWidthTests`, four fixtures, the first two failing against `main`:

- an auto `float: right` reaching the right of a 300pt container;
- text sitting beside an auto `float: left` on the same line — the consequence a reader actually notices;
- a declared width unchanged at exactly 45pt;
- an auto float never exceeding its containing block, since `max(min-content, available)` is bounded above by max-content.

One note from writing them: a 1px border insets the float's own text by 0.75pt, so "the text sits beside the float" cannot be asserted as an equal baseline. My first version compared tops to ±0.05pt and failed against correct behaviour.

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-floats-auto-width-is-now-shrink-to-fit.md
+++ b/.claude/migration-notes/2026-09-08-a-floats-auto-width-is-now-shrink-to-fit.md
@@ -1,0 +1,23 @@
+# A float's `auto` width is now shrink-to-fit
+
+CSS 2.1 §10.3.5 makes a floating box's `auto` width shrink-to-fit —
+`min(max-content, max(min-content, available))`. It was given the ordinary stretch-to-containing-block
+width instead, so an auto-width float filled its containing block.
+
+That leaves `float: right` nowhere to go: it landed at the left as a full-width bar, and the content
+that belongs beside it was pushed onto its own line. A bordered badge tucked into the top right of a
+cell rendered as a full-width band above the cell's text.
+
+Such a float now shrinks to its content and lands where a browser puts it. Measured against Chrome
+152, on a 300pt container:
+
+| | Chrome 152 | before | after |
+| --- | --- | --- | --- |
+| auto-width `float: right`, offset into the container | 257.9pt | 0 (full width) | at the right edge |
+| text after an auto-width `float: left` | beside it | on its own line | beside it |
+
+A float with a **declared** width is unaffected — placement was never the problem, and a declared
+width already landed on a browser's x to the point. An auto-width float is still never wider than
+its containing block.
+
+See [Display & Layout](../../docs/html-css-support.md#display--layout) in `docs/html-css-support.md`.
--- a/.claude/recent-fixes/2026-09-08-a-floats-auto-width-is-shrink-to-fit.md
+++ b/.claude/recent-fixes/2026-09-08-a-floats-auto-width-is-shrink-to-fit.md
@@ -1,0 +1,31 @@
+# A float's auto width is shrink-to-fit
+
+`GetBoxWidth` gave a floating box with `width: auto` the stretch-to-containing-block width. CSS 2.1
+§10.3.5 makes it shrink-to-fit. A float that fills its containing block leaves `float: right`
+nowhere to go, so it lands at the left as a full-width bar and the content that belongs beside it is
+pushed onto its own line.
+
+## What running it turned up
+
+- **Float PLACEMENT was never wrong, which is what made this hard to see.** A float with a DECLARED
+  width lands on Chrome's x to the point, so the placement machinery looks correct under any test
+  that declares a width. Only the auto measurement was wrong, and it is the auto case that a badge
+  or a pull-quote actually uses. There is a fixture pinning the declared-width case precisely so the
+  change is visibly scoped to `auto`.
+- **The formula is already in the tree.** `min(max-content, max(min-content, available))` is what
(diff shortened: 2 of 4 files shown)


jhaygood86:
Reviewed the diff, checked out the branch, rebuilt (0 warnings), and ran the float suite (96/96 pass).

Code-reuse verification (point 4 of the review): confirmed genuine. The new branch in `GetBoxWidth` calls `Math.Max(GetFitContentWidth(...), GetMinContentWidth(...))`, which is the exact same pattern already used for the orthogonal-flow shrink in `CssBox.cs` (~line 4283) and for absolute-positioning shrink-to-fit a few lines below this change (`GetBoxWidth` line ~1719). Verified algebraically too: `max(min(Mx,A), Mn) == clamp(A, Mn, Mx)` for `Mn <= Mx`, which is what §10.3.5's `min(max-content, max(min-content, available))` reduces to — so this is not just borrowed syntax, it's the correct formula. No parallel implementation.

One issue found: the new `.claude/migration-notes/2026-09-08-a-floats-auto-width-is-now-shrink-to-fit.md` file links to `docs/html-css-support.md#floats`, but that anchor doesn't exist — there is no "Floats" heading anywhere in `html-css-support.md` (the `float` property is just a table row under "### Display & Layout", whose real GitHub-generated anchor is `#display--layout`). Worth fixing before merge, or at least before this note gets folded into release notes.

Declared-width bypass (point 5): confirmed the `float: right; width: 60px` case genuinely takes a different code path — `box.Width != Keywords.Auto` is false for a declared width, so it never reaches the new `Width: Keywords.Auto` branch at all (not "happens to produce the same number").

dchupp:
Fixed — the anchor is now `#display--layout`, and thank you for checking it rather than taking the link on trust.

Since that is the second broken anchor I have shipped, I wrote a checker and swept **every** `docs/**` anchor cited from `.claude/**`, on this branch and on `main`.

Two things came out of it, and the first is a correction to my own tooling rather than a finding:

**My first version of the checker reported nine false positives**, including `#display--layout` itself. GitHub drops punctuation and then maps each remaining **space** to a hyphen — it does not collapse runs — so "Display & Layout" is `display--layout` with two, and "Running elements (`position: running()` / `element()`)" is `running-elements-position-running--element`. Collapsing whitespace, which is the obvious way to write it, marks every heading containing punctuation as broken. Worth knowing before anyone writes the same check.

**With that corrected, all of my own notes resolve**, and two genuinely broken anchors remain on `main` — both yours, both from August, both pointing at `#text`, which does not exist:

```
.claude/migration-notes/2026-08-06-vertical-align-length-percentage-support.md
.claude/migration-notes/2026-08-06-text-align-center-with-text-indent-under-rtl.md
```

The candidates are `#inline-text` (line 139) or `#text-layout` (line 380) — the `vertical-align` and `text-indent` rows are under the former. I have deliberately **not** touched them here since they are unrelated to this PR; happy to send a one-line fix as its own PR, or leave them to you.

On your other two points: confirmed on both. The declared-width bypass is a genuinely different path rather than a coincidence (`box.Width != Keywords.Auto` short-circuits before the new branch), and the algebra you did — `max(min(Mx,A), Mn) == clamp(A, Mn, Mx)` — is the same reduction I was relying on, so it is reassuring to see it independently.

jhaygood86 (approved):
Verified: the anchor is corrected to `#display--layout` and matches the file's real heading.

Really appreciate the anchor sweep — writing a checker, catching your own false positives before reporting them (GitHub's space-to-hyphen-per-space rule is a good thing to have documented now), and flagging the two genuinely broken anchors in my own migration notes from August rather than fixing them unprompted in an unrelated PR. I'll get `#text` → `#inline-text` fixed in those two files separately; no need for you to carry that.

Build clean, float-filtered suite (96 tests) passes, all three CI platforms green.

Approving. Merging now.

PR #953, opened 2026-09-08
Resolve the document base once per tree, not once per reference

rgroenewoudt:
## Resolve the document base once per tree, not once per reference

`<base href>` is looked up with a fresh `DomUtils.GetBoxByTagName(Root, "base")` at each of the three
places that need it:

- `CommonUtils.ResolveAgainstDocumentBase` — images, stylesheets
- `HtmlContainer.ResolveHref` — links, bookmark targets
- `PdfGenerator.HandleRunningElementLinks` — a link inside a `content: element()` running element

All three ask **per reference**, not per document, and the walk cannot short-circuit when the
document declares no `<base>` at all — it visits every box in the tree before returning null. A
running element's own `<a href>` is re-resolved on every page its element was selected onto, so the
cost is `pages × links × boxes`.

This adds `HtmlContainerInt.DocumentBaseUri` as the single place that rule lives — memoized against
the tree's topmost box identity, in the same shape as `_idIndex` — and routes all three call sites
through it. `CommonUtils`' own doc comment already claimed to be "the single place this resolution
rule lives for images and stylesheets alike"; now it is one for links too.

### What it removes

Box visits inside `GetBoxByTagName` for one render, counted with a temporary counter, so this figure
is deterministic and machine-independent:

| Document | Before | After |
|---|---|---|
| 188-page paginated report, one link in its running footer | 5 525 506 | 79 849 |
| 7-page report | 19 127 | 4 815 |

In wall time that is **5–7 %** off the 188-page report (net10.0, Release, three interleaved rounds of
8 renders per side: median wall 4.07 s → 3.86 s, best-of-run process CPU 3.77 s → 3.50 s). On a 1–7
page document the difference is below this machine's run-to-run noise.

One caveat on profiling, since it would be easy to over-claim here: `dotnet-trace`'s sample profiler
puts `GetBoxByTagName` at 18.0 % inclusive of the render's CPU, all of it under
`HandleRunningElementLinks`. That figure is inflated — the function recurses to the depth of the box
tree, so its stacks are the deepest in the render and each sample taken inside it costs more to walk,
which inflates the frame's sampled share. The box-visit count is the honest measure of the work
removed; 5–7 % is what it was worth.

### Behaviour

Unchanged, deliberately. Three things worth naming:

- **`<base>` is still ignored for a `<link>`/`@import` stylesheet.** Those load from inside
  `DomParser.GenerateCssTree`, and `Root` is only assigned once `SetHtml` returns, so they reach
  `ResolveAgainstDocumentBase` with a null root and fall back to the adapter's base URI — existing
  behaviour, preserved. This does *not* apply to ``, which resolves during layout with `Root`
  assigned and honours `<base>` today (`FileUriLoaderIntegrationTests.BaseHrefFileUri_UnderDefault
  LoaderRoutesFileResourcesThroughAdapter` reaches its image only through `<base href>`). The
  null-root early return in the getter is a fast path, not a correctness guard: the key already
  handles a null root, since the next read against a real tree fails `ReferenceEquals` and re-walks.
- **`RAdapter.BaseUri` is not frozen into the memo** — it is a cheap property on the adapter, not a
  walk, so a document with no `<base>` keeps answering whatever the adapter currently says. The
  *declared* base is memoized as an already-parsed `RUri`, so a document that has one does not
  re-parse it per reference either.
- **A malformed `<base href>`** still throws from the same point relative to each caller's own work,
  and — because the memo key is committed only after the `RUri` is constructed — from every later
  read too, not just the first.

`Clear()` now also drops the two memo fields, so a re-parse (the `@container` convergence loop, or
`SetContent`'s shrink-to-fit path) does not leave the previous, already-disposed tree reachable.

### Tests

- `Html/Core/DocumentBaseUriTests.cs` — no document at all, no `<base>`, a declared `<base>`, a
  whitespace-only `href`, a memoized read, and a second `SetHtml` replacing the first document's
  answer.
- `RunningElementLinksAndTaggingIntegrationTests.RunningElementLink_RelativeHref_ResolvesAgainstTheDocumentBase`
  — a relative `href` inside a running element, asserting the written `/URI` action is base-resolved
  rather than raw, and that it appears once per page. That branch had no base-resolution coverage.

Each was checked for discrimination rather than just coverage:
`WithinOneTree_TheBaseIsResolvedOnlyOnce` fails when the memo key is patched to always re-walk,
`ASecondDocumentReplacesTheFirstOnesBase` fails if it never invalidates, and
`WithABlankBaseHref_FallsBackToTheAdapterBaseUri` fails if the whitespace check is dropped.

Full net8.0 suite: 10 258 passed; every executable changed line is covered.
`dotnet build PeachPDF.slnx -t:Rebuild` is warning-free on both TFMs.

One pre-existing failure is unrelated and reproduces on unmodified `main`:
`FontSynthesisIntegrationTests.ObliqueWithExplicitAngle_RegularOnlyFamily_UsesDeclaredAngleNotFixedDefault`
builds its expected substring with `Math.Sin(...).ToString("0.####")` — current-culture formatting —
and looks for it in a PDF that writes numbers invariant, so it only passes on a period-decimal
culture. It passes here under `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1`. Happy to fix that in a
separate PR if you want it.

### DIFF
--- a/.claude/recent-fixes/2026-09-08-the-document-base-is-resolved-once-per-tree-not-once-per-reference.md
+++ b/.claude/recent-fixes/2026-09-08-the-document-base-is-resolved-once-per-tree-not-once-per-reference.md
@@ -1,0 +1,78 @@
+# The document base is resolved once per tree, not once per reference
+
+`<base href>` was looked up with a fresh `DomUtils.GetBoxByTagName(Root, "base")` at each of the three
+places that need it — `CommonUtils.ResolveAgainstDocumentBase` (images, stylesheets),
+`HtmlContainer.ResolveHref` (links, bookmark targets) and `PdfGenerator.HandleRunningElementLinks`
+(a link inside a `content: element()` running element). All three ask **per reference**, not per
+document, and that walk cannot short-circuit when the document declares no `<base>` at all: it visits
+every box in the tree before returning null. A running element's own `<a href>` is re-resolved on
+every page its element was selected onto, so the cost is `pages × links × boxes`.
+
+`HtmlContainerInt.DocumentBaseUri` is now the single place the rule lives, memoized against the
+tree's topmost box identity exactly like `_idIndex`.
+
+## What running it showed, that reading it did not
+
+A 188-page report (a real paginated document with one external link in its running footer) visits, in
+`GetBoxByTagName` alone:
+
+| | box visits |
+|---|---|
+| 188-page report, before | 5 525 506 |
+| 188-page report, after | 79 849 |
+| 7-page report, before | 19 127 |
+| 7-page report, after | 4 815 |
+
+**The sampled profile badly over-stated the time.** `dotnet-trace`'s sample profiler put
+`GetBoxByTagName` at 18.0 % inclusive of the render's CPU (2.38 s of 13.20 s across three renders),
+every bit of it under `HandleRunningElementLinks`. Removing it does not buy 18 %: it buys 5–7 %
+(188-page report, three interleaved rounds of 8 renders per side — median wall 4.07 s → 3.86 s,
+best-of-run CPU 3.77 s → 3.50 s). On a 1–7 page document the difference is below this machine's
+run-to-run noise. The reason is worth remembering before trusting a sampled percentage again — this
+function recurses to the depth of the box tree, so its stacks are the deepest in the whole render and
+each sample taken inside it costs proportionally more to walk. A frame's sampled share is inflated by
+its own stack depth. The box-visit count above is the honest measure of the work removed; the
+timing is what the work was actually worth.
+
+## Deliberately not done
+
+- **`<base>` is still ignored for a `<link>`/`@import` stylesheet.** Those load from inside
+  `DomParser.GenerateCssTree`, and `Root` is only assigned once `SetHtml` returns, so they reach
+  `ResolveAgainstDocumentBase` with a null root and fall back to the adapter's base URI. That is
+  long-standing behaviour and this change preserves it byte-for-byte. Whether the parse-time path
+  *should* honour `<base>` is a separate question. Note this does **not** apply to `<img>`: images are
+  resolved during layout (`CssBoxImage.MeasureWordsSize` → `ImageLoadHandler.SetImageFromPath`), by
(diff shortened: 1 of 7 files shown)


jhaygood86 (approved):
Reviewed against this repo's CLAUDE.md conventions. Verified by checking out this branch, running `dotnet build PeachPDF.slnx -t:Rebuild` (0 warnings), and running the DocumentBaseUri/RunningElementLinksAndTagging/ResolveHref/FileUriLoader-filtered test suites (52 tests, all pass).

This is a clean, well-scoped perf fix — thanks for the thorough writeup, especially the honesty about the sampled-profiler figure overstating the actual win (18% inclusive vs. the real 5-7%), which is exactly the kind of "what running it showed, that reading it did not" detail this repo's `recent-fixes` convention asks for.

A few things I checked independently rather than just trusting the description:
- **Memoization invalidation is correct.** Confirmed `SetHtml` calls `Clear()` before reassigning `Root` to a freshly-allocated tree from `GenerateCssTree` — so `ReferenceEquals(_documentBaseRoot, Root)` genuinely invalidates on every re-parse (container-query convergence, `PdfGenerator.SetContent`'s shrink-to-fit path), not just the common case.
- **`WithinOneTree_TheBaseIsResolvedOnlyOnce`** is a real discrimination test, not just coverage — mutating the tree's `<base>` attribute after the first read and confirming the memoized value doesn't change is a good way to prove memoization from the outside without needing a call counter.
- **The `<link>`/`@import`-stylesheet-ignores-`<base>` behavior is correctly preserved** (parse-time resolution happens before `Root` is assigned) and clearly distinguished from `` (which does honor it, at layout time) — this distinction is easy to get backwards and it's stated precisely.
- No doc-comment-detachment issues (a pattern that recurred across an earlier contributor's PRs in this repo) — every new/moved doc comment stays correctly attached to its own member.

The disclosed pre-existing test failure (`FontSynthesisIntegrationTests.ObliqueWithExplicitAngle_RegularOnlyFamily_UsesDeclaredAngleNotFixedDefault`, a current-culture-formatting bug unrelated to this change) is a nice catch and appreciated flagging rather than silently working around it.

Approving. Running the check workflow now (this needed manual approval as a first-time contributor) — will merge once it passes.

PR #955, opened 2026-09-08
Do not clip table cells from the UA stylesheet

dchupp:
`td, th { overflow: hidden }` in the default stylesheet is not from any specification.

The HTML Standard's rendering section ([§15.3.8, Tables](https://html.spec.whatwg.org/multipage/rendering.html#tables-2)) gives table cells exactly three properties:

```css
td, th { display: table-cell; }
td, th { padding: 1px; }
th     { font-weight: bold; }
```

No `overflow`. No browser clips a cell either.

## Why it survived this long

It is **invisible under automatic table layout** — a narrow `td` simply widens to fit its content, so the clip never bites. My first repro reported identical output before and after, which is what the note in our own tree had recorded as "cannot demonstrate". It needs `table-layout: fixed`, where the column cannot grow.

## The observable is ink, not text

A clipped glyph is still in the content stream, so `pdftotext` finds it either way and reports the document complete — the same property that makes this class of loss undetectable from the finished PDF. Rasterising is the only way to see it.

Measured with MuPDF at 200dpi, counting dark pixels to the right of a 40pt cell holding a 15-character word in a `table-layout: fixed` table:

| | ink right of the cell |
| --- | --- |
| **Chrome 152** | **2021** |
| `main` | 1315 |
| this change | 2542 |

One renderer rather than the two the testing conventions ask for — `pypdfium2` is not installed on this machine. MuPDF is specifically the one that *honours* the clip, which is the property under test, so I do not think the second adds much here; happy to add it if you would rather.

## Three existing fixtures depended on the UA rule

`OverflowClipIntegrationTests`' rounded-box-in-a-table-cell tests obtained a clipping cell from the default sheet, and one asserted it outright:

```csharp
// td has overflow:hidden from the PeachPDF default stylesheet
Assert.Equal(Overflow.Hidden, td!.Overflow.Value);
```

Their real subject is the padding-box clip bound under a non-uniform `border-radius`, not where the cell's overflow came from — so each fixture now declares `overflow: hidden` on the cell itself and the assertions are unchanged. Flagging it prominently because it is a change to your tests, not mine, and I would rather you disagreed with it explicitly than found it later.

## Tests

`TableCellOverflowTests`, three fixtures, the first two failing against `main`:

- a cell's computed `Overflow` is `Visible`;
- a cell pushes **no clip of its own** — asserted through `FragmentPaintHarness` + `RecordingGraphics.PushedClips`, per the convention for a change that alters what reaches the graphics layer;
- an author-declared `overflow: hidden` still clips, so removing the UA rule has not removed the property's effect.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,348 passed, 0 failed, 9 skipped.


jhaygood86:
Reviewed against this repo's conventions. Build is clean (0 warnings on `dotnet build -t:Rebuild`), the targeted Table/OverflowClip test filter passes (725/725), and the related `MonolithicContentLayoutIntegrationTests`/`TableSpannedBandRepetitionTests` (43 tests) that reference table-cell overflow assumptions still pass unaffected.

**Spec verification (confirmed):** fetched https://html.spec.whatwg.org/multipage/rendering.html#tables-2 directly — §15.3.8 gives `td, th` only `display: table-cell` and `padding: 1px`, plus `th { font-weight: bold }`. No `overflow` anywhere in the section. Cross-checked the rest of that section's rules (`table { display }`, `caption { text-align: center }`, `border-spacing: 2px`, `vertical-align: middle`/`inherit`) against `CssDefaults.cs` — all already match; no other UA-sheet mismatch found.

**Fixture preservation (confirmed correct):** the three `OverflowClipIntegrationTests` fixtures now declare `overflow: hidden` on the `td` in their own CSS rather than relying on the UA default; assertions are unchanged and the real subject (padding-box clip bound under non-uniform `border-radius`) is untouched.

**One real gap — `docs/html-css-support.md` was not updated.** The `table-layout` row (line 786) still reads: *"...so overflowing text wraps (**or is clipped**) inside its column rather than widening it."* That's now stale — with the UA rule removed, a `table-layout: fixed` cell no longer clips by default. Per this repo's own convention ("update the relevant doc page... in the same change, rather than as a follow-up"), this line needs correcting. The new migration note even links to this section as though it documents current behavior.

**Stale explanatory comments** (not wrong code, just now-false reasoning) referencing "td/th get overflow:hidden from the UA stylesheet" remain in `src/PeachPDF/Html/Core/Dom/CssBox.cs:3443-3444` and in two test files (`MonolithicContentLayoutIntegrationTests.cs:349-351`, `TableSpannedBandRepetitionTests.cs:385-386`). None affect correctness (the `CssBox.cs` exclusion is keyed off `ActualDisplay`, not overflow value) but they should be corrected alongside the doc fix.

Given the behavioral weight of this change (previously-invisible content becomes visible in existing `table-layout: fixed` documents), worth a maintainer call on whether that's acceptable even though it's spec-correct — some real documents may have been unintentionally relying on the old clipping.

dchupp:
All three addressed, rebased and force-pushed. And thank you for tracing the spec section yourself and cross-checking the rest of §15.3.8 against `CssDefaults.cs` — that is a stronger check than the one I did, which only looked at the rule I was removing.

**The `docs/html-css-support.md` gap — fixed, and you are right that it was the real one.** The `table-layout` row now reads:

> …so overflowing text wraps inside its column rather than widening it — and text that cannot wrap overflows the column visibly, as it does in a browser, since a cell does not clip unless it declares `overflow: hidden` itself.

The migration note now says the row was updated rather than linking it as though it already described current behaviour.

**Stale explanatory comments — corrected, and one of them was wrong in a more interesting way than "now false".** `CssBox.cs`'s exclusion comment said cells are excluded because the UA sheet gives them `overflow: hidden`. As you note, the exclusion is keyed off `ActualDisplay`, so the reasoning was never quite what the code did — the comment now says the exclusion is by display *rather than* by overflow value, and why that matters: a cell that declares `overflow: hidden` itself would otherwise read as monolithic and lose `CssLayoutEngineTable`'s own fragmentation. Same correction in `MonolithicContentLayoutIntegrationTests`. I could not find a matching stale comment in `TableSpannedBandRepetitionTests` — its `overflow` mentions are all about a row overflowing *through* a band, unrelated — so I have left it alone; point me at it if I have missed one.

**On the behavioural call — agreed, and it is yours to make.** To put the exposure as precisely as I can: it only reaches a document where a column genuinely cannot widen, which in practice means `table-layout: fixed` or an explicit `<col>`/cell width that the content exceeds. Under automatic layout nothing changes at all, which is why this was invisible for so long. So the documents at risk are the ones that set a fixed layout *and* overflow it, and for those the previous behaviour was silently dropping content rather than showing it — but you are right that "previously invisible content becomes visible" is a real change for a document someone has already tuned against the old output, and that is a judgement about your users rather than about the spec.

If you would rather not take it, the honest alternative is to leave the UA rule and document it as a deliberate deviation, since it is the kind of thing the next person will otherwise re-discover. I would still suggest fixing the `table-layout` doc row either way, since it currently describes clipping as a property of fixed layout rather than of the UA sheet.

Separately: while sweeping for stale references I found one broken doc anchor of my own left on `main` from #951 — sent as #960, same class as the two you fixed in #959.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,376 passed, 0 failed, 9 skipped.

jhaygood86 (approved):
Verified the docs fix: `table-layout` row now correctly describes overflow behavior under `fixed` layout, matching browsers. Also verified the stale-comment corrections — nice catch on your end that `CssBox.cs`'s exclusion comment was subtly wrong in its own right (keyed off `ActualDisplay`, not overflow value), not just stale.

On the behavioral judgment call: proceeding with this as-is. The exposure is narrow (only `table-layout: fixed` or an explicit width that content overflows — invisible under automatic layout, which is the common case), and the previous behavior was silently discarding content rather than showing it, which is arguably the worse failure mode of the two. This also matches the standing orientation across this whole batch of PRs toward spec/browser-parity over incidental existing behavior. If a real-world regression report comes in, it's easy to revisit as a documented, tracked accepted-deviation instead.

Build clean, Table/OverflowClip-filtered suite (730 tests) passes.

PR #958, opened 2026-09-08
Match HTML names ASCII case-insensitively, not by Unicode case folding

dchupp:
The selector matcher, the tag/class/id indexes, `IsBrElement`, the font caches and the entity tables all compare names with `InvariantCultureIgnoreCase`.

HTML matches element and attribute names **ASCII** case-insensitively. Invariant culture applies full **Unicode** case folding, and under that **U+212A KELVIN SIGN equals ASCII `k`**:

```html
<style>.k { color: rgb(255,0,0) }</style>
<div class="&#x212A;">x</div>   <!-- matches .k on main; it should not -->
```

The comment already above `NameComparison` states the rule correctly — the code just did not express it:

> HTML matches element/attribute names **ASCII case-insensitively** (unlike SVG's XML case-sensitivity), so `NameComparison` reports an ignore-case comparison.

Ordinal is what "ASCII case-insensitively" means.

## On the performance side — deliberately not quoted

This started life as a profile-driven change: `System.Globalization` was about a fifth of thread CPU on a real corpus (`CompareInfo.Compare` 14.9%, `IcuGetHashCodeOfString` 4.3%), and throughput rose ~5%. The matcher runs this comparison for one name against every candidate rule for every box, so a culture-aware comparison routes each one through ICU.

**I could not reproduce a trustworthy number on the machine I prepared this on.** Nine interleaved renders of a synthetic class-heavy document gave a 175 ms median before and 177 ms after, with a 164–225 ms spread — the variance swamps any effect that size. Rather than quote a figure I cannot stand behind, I am offering this as the correctness fix it also is; if the throughput matters to you it is worth measuring on a quiet machine, and I would expect it to show.

## Tests

`AsciiNameMatchingTests`, three fixtures:

- a Kelvin-sign class does **not** match `.k` — **fails against `main`**;
- an ASCII class still matches case-insensitively, both `k` and `K`;
- an uppercase `DIV` element still matches the `div` selector.

The last two are the contrast cases: the point is to narrow the comparison to ASCII, not to make it case-sensitive, and both would fail if this went too far.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,348 passed, 0 failed, 9 skipped.

### DIFF
--- a/.claude/recent-fixes/2026-09-08-html-names-match-ascii-case-insensitively.md
+++ b/.claude/recent-fixes/2026-09-08-html-names-match-ascii-case-insensitively.md
@@ -1,0 +1,46 @@
+# HTML names match ASCII case-insensitively, not by Unicode case folding
+
+The selector matcher, the tag/class/id indexes, `IsBrElement`, the font caches and the entity tables
+all compared names with `InvariantCultureIgnoreCase`. HTML matches element and attribute names
+**ASCII** case-insensitively; invariant culture applies full Unicode case folding.
+
+Under that folding **U+212A KELVIN SIGN equals ASCII `k`**, so `class="K"` matched the selector `.k`
+and a `K` element matched `k`. The comment above `NameComparison` already said the rule was ASCII
+case-insensitivity — the code just did not express it.
+
+## What running it turned up
+
+- **The correctness case is demonstrable and the performance case is not, on this machine.** This
+  started as a profile-driven change (`System.Globalization` was a fifth of thread CPU on a real
+  corpus, and throughput rose ~5%). I could not reproduce a trustworthy number here: nine interleaved
+  renders of a synthetic class-heavy document gave a 175ms median before and 177ms after, with a
+  164–225ms spread — the variance swamps the effect. So this is offered as the spec fix it also is,
+  with a deterministic repro, and the throughput left unquantified rather than quoted from a machine
+  that cannot show it.
+- **Ordinal is the cheaper comparison as well as the correct one**, which is why both readings point
+  the same way: the matcher runs one name against every candidate rule for every box, and a
+  culture-aware comparison routes each through ICU.
+- **A blanket `sed` to strip our internal comment tag ate the comment marker with it**, turning a
+  `// TAG: Ordinal, not InvariantCulture` line into bare prose and producing 19 syntax errors.
+  Replace the tag with `// `, never delete it.
+
+- **Two call sites were worse than the ones the first sweep found, and it missed them.**
+  `DomUtils.FindParent` (the parser's implied-tag-closing rules — a `<table>` closes an open `<p>`)
+  and `DomParser.CascadeParseStyles` (the `link`/`style` tag names and the `stylesheet` relation)
+  used `CurrentCultureIgnoreCase`: full Unicode folding **and** locale-dependent, so the same
+  document could parse differently on two machines. Grepping for the comparison the fix replaced
+  found neither; the sweep has to be for every culture-aware comparison, not for one enum value.
+- **`FontResolver` keys its installed-font table on `.ToLower()`**, which is also the current
+  culture. That is a cross-machine determinism bug in its own right — a Turkish locale lower-cases
+  `I` to a dotless `ı` — so all four sites are `ToLowerInvariant()`. It does NOT fix the folding
+  (`char.ToLowerInvariant('K')` is still `k`), and font-family matching is not the spec-mandated
+  ASCII rule element names are, so that half is left alone deliberately.
+
+## Evidence
+
+`AsciiNameMatchingTests`, five fixtures: a Kelvin-sign class does not match `.k` (fails against
+`main`); an ASCII class still matches case-insensitively in both cases; an uppercase `DIV` still
+matches `div`; implied tag closing still matches `P` to `p`; and an uppercase `STYLE` element is
+still honoured. The last four are contrast cases — the point is to narrow the comparison to ASCII,
+not to make it case-sensitive. Full suite green on net8.0
+(10,348), 0 new build warnings.
--- a/src/PeachPDF.Tests/Integration/AsciiNameMatchingTests.cs
+++ b/src/PeachPDF.Tests/Integration/AsciiNameMatchingTests.cs
@@ -1,0 +1,83 @@
+using PeachPDF.Html.Core.Dom;
+using PeachPDF.Tests.TestSupport;
+using System.Threading.Tasks;
+using static PeachPDF.Tests.TestSupport.LayoutHarness;
+
+namespace PeachPDF.Tests.Integration
+{
+    /// <summary>
+    /// HTML matches element and attribute names <b>ASCII</b> case-insensitively, and a class name is
+    /// matched ASCII case-insensitively at most. <c>InvariantCultureIgnoreCase</c> applies full
+    /// Unicode case folding instead, under which U+212A KELVIN SIGN equals ASCII <c>k</c> — so a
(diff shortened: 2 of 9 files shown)


jhaygood86:
Reviewed against CLAUDE.md conventions (build, tests, spec check, sweep for missed call sites). Build is clean (0 warnings), the three `AsciiNameMatchingTests` fixtures pass, and I confirmed empirically that `OrdinalIgnoreCase` does *not* fold U+212A KELVIN SIGN to `k` while `InvariantCultureIgnoreCase` does (and still folds ASCII `k`/`K`/`div`/`DIV` correctly) — the fix and its comparator choice are correct where applied.

Two things worth fixing before merge:

**1. Stale/incorrect comment — `CssData.cs:988-991`**

```csharp
// Attribute selectors read the value once via node.GetAttribute (case-sensitivity of the
// attribute-name lookup is the node's own: case-insensitive for HTML/CssBox, case-sensitive for
// SVG), then compare the value with node.NameComparison (InvariantCultureIgnoreCase for HTML -
// byte-identical to the previous behaviour - Ordinal for SVG).
```
This still says HTML's `NameComparison` is `InvariantCultureIgnoreCase` and "byte-identical to the previous behaviour" — both are now false; `CssBox.NameComparison` is `OrdinalIgnoreCase` and the whole point of the PR is that this is *not* byte-identical (that's what the Kelvin-sign test demonstrates). Doc-comment-detachment: the comment right above `NameComparison`'s declaration in `CssBox.cs` was updated correctly, but this second reference to the same fact in `CssData.cs` wasn't.

**2. The sweep missed two call sites in the same bug class, using an even more nondeterministic comparison (`CurrentCultureIgnoreCase`, not `InvariantCultureIgnoreCase`):**

- `DomUtils.FindParent` (`DomUtils.cs:88`) — `box.HtmlTag.Name.Equals(tagName, StringComparison.CurrentCultureIgnoreCase)`, used by the HTML parser's implied-tag-closing logic (e.g. the "table closes p" rule per the method's own doc comment).
- `DomParser.CascadeParseStyles` (`DomParser.cs:246,249,258`) — matching the `link`/`style` tag names and the `stylesheet` link-relation token, also via `CurrentCultureIgnoreCase`.

Both are HTML tag/token name matching, the exact category this PR fixes elsewhere (`IsBrElement`, the selector matcher). `CurrentCultureIgnoreCase` is worse than `InvariantCultureIgnoreCase`: it does full Unicode case folding *and* is locale-dependent (non-deterministic across machines/OS locales), so it's exploitable the same way as the Kelvin-sign case plus has a reproducibility problem the PR is otherwise careful to avoid elsewhere. Given the PR's stated framing (Unicode-confusable-adjacent correctness issue), these should be swept to `OrdinalIgnoreCase` in the same change or an immediate follow-up.

Minor/non-blocking: `FontResolver.cs`'s `InstalledFonts` family-name lookups (`ResolveTypeface`, `HasExplicitRanges`, `AddFont`) key on `fontFamilyName.ToLower()` (culture-sensitive, and I confirmed `char.ToLowerInvariant('K')` is also `k` — same folding bug). The PR body says "the font caches" (plural) were fixed, but only `FontsHandler.cs`'s three dictionaries were touched; this is a separate cache using a different mechanism (`.ToLower()` normalization rather than a culture-aware comparer) that wasn't addressed. Lower severity since font-family matching isn't the same spec-mandated ASCII rule as element/attribute names, but worth a look given the PR's own wording.

No issue with the performance-honesty framing — the PR body's disclaimer is specific to the unreproducible throughput number, and the profiler percentages it does quote (14.9%/4.3%) are presented as directly observed, not extrapolated. The `CssData.cs` code comment repeats the 4.3% figure without that caveat, which could read as a stronger performance claim out of context, but I don't think it's a blocking issue.

dchupp:
All three addressed, rebased and force-pushed. Point 2 is the important one — you are right that the sweep was incomplete, and right that what it missed is *worse* than what it fixed.

**2. `CurrentCultureIgnoreCase` at two sites, now `OrdinalIgnoreCase`.**

`DomUtils.FindParent` (implied tag closing — a `<table>` closing an open `<p>`) and `DomParser.CascadeParseStyles` (the `link`/`style` tag names and the `stylesheet` relation). As you say, that comparison is full Unicode folding **and** locale-dependent, so the same document could parse differently on two machines — a reproducibility problem this PR is otherwise specifically about avoiding.

Worth naming why the sweep missed them: I grepped for the comparison I was *replacing* rather than for every culture-aware comparison. `InvariantCultureIgnoreCase` finds nothing at either site. The right sweep is for the category.

Two fixtures added: `P` still closes as `p` through the implied-closing path, and an uppercase `<STYLE>` element is still honoured.

**1. The stale `CssData.cs` comment** is corrected — it now says `OrdinalIgnoreCase`, and the "byte-identical to the previous behaviour" clause is gone, since as you note that is precisely what this PR is not. You are right that this was the same doc-comment-detachment shape: one statement of a fact updated, a second copy of it elsewhere left behind.

**Also removed the `4.3%` figure from the `CssData.cs` comment.** Your read is right that it reads as a stronger claim than the PR body's caveat supports out of context, and I could not reproduce any throughput number on this machine — so it should not sit uncaveated in the source.

**Minor: `FontResolver`.** You are right that the PR body said "the font caches" and only `FontsHandler`'s three dictionaries were touched. `FontResolver.InstalledFonts` keys on `.ToLower()`, which is the **current culture** — a cross-machine determinism bug in its own right, since a Turkish locale lower-cases `I` to a dotless `ı`. All four sites (the one write and three reads, so the keys stay consistent) are now `ToLowerInvariant()`.

I have deliberately **not** gone further there: as you say `char.ToLowerInvariant('K')` is still `k`, so that does not fix the folding, and font-family matching is not the spec-mandated ASCII rule element and attribute names are. Fixing the folding for font names would need ordinal normalisation and its own reasoning about what CSS font matching actually requires — happy to look at it separately if you want it.

---

**One thing to flag rather than let CI surprise you.** On the first full-suite run after these changes I saw a single failure in `BorderStylePaintIntegrationTests.BorderColorPerSide_ResolvesDistinctColorPerEdge_IncludingCurrentColor`. I could not reproduce it: it passes in isolation and on two further full runs of this branch, and `main` passed 3/3 in the same session.

I do not think it is this change — under an `en-US` locale `ToLower()` and `ToLowerInvariant()` produce identical strings for every ASCII font name, and all four call sites moved together, so the dictionary cannot key differently. The most likely explanation is that adding a test class shifts xUnit's parallel-collection schedule and surfaces a pre-existing order-dependence, which is the risk `CLAUDE.md` already records for the process-wide static font caches. Flagging it because 1-in-4 against 0-in-3 is not enough evidence for me to call it cleanly, and you should know before it possibly appears in CI.

`dotnet build PeachPDF.Tests -t:Rebuild`: no new warnings. `dotnet test --framework net8.0`: 10,378 passed, 0 failed, 9 skipped (runs 2 and 3).

jhaygood86 (approved):
Verified all three fixes directly. `DomUtils.FindParent` and the three `DomParser.CascadeParseStyles` comparisons (link, stylesheet-token, style) are now `OrdinalIgnoreCase`, confirmed no remaining `CurrentCultureIgnoreCase` anywhere in source. `FontResolver`'s four `InstalledFonts` sites are now `ToLowerInvariant()` — correctly scoped as a determinism fix, not claimed as a folding fix, since you're right that font-family matching isn't the ASCII-only rule that applies to element/attribute names.

Good instinct naming why the sweep missed these the first time (grepped for the comparison being replaced rather than the whole culture-aware category) — that's the kind of process note worth having for next time.

Build clean, AsciiName/Selector-filtered suite (286 tests) passes, all three CI platforms green.

Approving. Merging now.

PR #962, opened 2026-09-08
A relocated table no longer draws its abandoned run's header

dchupp:
A table carrying `break-inside: avoid` that does not fit in what is left of a page is laid out again from the start on the next one ([css-break-3 §4.2](https://www.w3.org/TR/css-break-3/#break-within)). If it repeats a `<thead>` or `<tfoot>`, the abandoned run's copies were drawn too — **three** headers for a table pushed onto the next page:

- one stranded near the bottom of the page the table had **left**, with no table beneath it, at the position the abandoned run gave it;
- two drawn exactly on top of each other at the top of the page it moved to, where the abandoned run's page break had been.

Reproduced against a plain 3-row table with a `<thead>` under a 660pt filler; Chrome 152 draws one header, on the landing page, at the coordinates this now produces.

## The cause

`RestoreStructureFromAnyPreviousRun` already undoes the visible half of a previous run — the `CssProxyBox`es it left in the table's child list. But a proxy is not what paint reads: laying one out also records a `CapturedInstance` with `FragmentEmitter` (`RecordRepeatingGroupFragmentInstance` → `HtmlContainerInt.RecordRepeatingGroupInstance`), and it is that record which becomes a fragment. Nothing dropped them.

`CssLayoutEngineColumns` has always made the matching call on its own re-run, with the comment that says exactly why:

> A container laid out again from the start may land in a different slot altogether, and the fragmentainers it recorded in the slot it has left describe geometry that no longer exists anywhere.

The table engine now does the same on a fresh pass. A resumed pass is excluded for the reason it is already excluded from the structure restore: the instances it would drop are the ones every earlier page of that same table repeats.

## Why this survived the existing test

`RepeatingTableRelayoutTests.TheTableInsideARelocatedCard_StillRepeatsItsHeaderExactlyOnce` counts `DisplayMode.TableHeaderGroup` boxes under the root. That count was **correct** — instrumenting the restore shows proxies before 2, after 0, and the following run adding exactly one — while the output was wrong. Any assertion about how often a repeating group is drawn has to come from the fragment tree, so the new test counts the group's words per fragmentainer. I have added that as an invariant note.

## Tests

`RelocatedTableHeaderInstanceTests`:

- six relocation cases — three filler heights × `<thead>`/`<tfoot>` — asserting nothing on the page the table left and exactly one where it landed. **All six fail against `main`** (0 expected / 1 actual on the abandoned page, 1 / 2 on the landing page).
- one control: the same table allowed to split across the boundary still repeats its header on both pages per [css-tables-3 §6.2](https://www.w3.org/TR/css-tables-3/#repeated-headers). It passes before and after, so a fix that suppressed repetition outright rather than dropping stale records would fail it.

Full suite green on net8.0 (10,394 passed), 0 new build warnings, diff coverage 100%.

Docs describe the intended behaviour already and needed no change; `.claude/` carries a recent-fixes entry, a migration note and the invariant.

### DIFF
--- a/.claude/invariants/proxy-undoing-a-runs-proxies-means-dropping-the-instances-they-recorded.md
+++ b/.claude/invariants/proxy-undoing-a-runs-proxies-means-dropping-the-instances-they-recorded.md
@@ -1,0 +1,20 @@
+# Undoing a run's proxies means dropping the instances they recorded
+
+A `CssProxyBox` is a layout-time placeholder, not what paint reads. Every proxy that lays out also
+calls `HtmlContainerInt.RecordRepeatingGroupInstance`, and it is that record — a `CapturedInstance`
+held by `FragmentEmitter`, keyed `(tableBox, slot)` — that becomes a fragment and gets painted.
+
+So a pass that abandons a run's output has **two** things to undo, and removing the proxies from
+`CssBox.Boxes` is only the first. The second is `ClearCapturedInstances(root)`, which
+`CssLayoutEngineColumns` has always called and `CssLayoutEngineTable` did not.
+
+**The measured symptom.** A table carrying `break-inside: avoid` that does not fit in what is left
+of a page is laid out again on the next one. `RestoreStructureFromAnyPreviousRun` dropped the
+abandoned run's proxies, and a test counting `DisplayMode.TableHeaderGroup` boxes in the tree
+therefore passed — while the PDF drew the `<thead>` **three** times: once stranded on the page the
+table had left, at the position the abandoned run gave it, and twice on top of itself at the top of
+the page it moved to, where the abandoned run's page break had been.
+
+The census that sees it counts words per fragmentainer in `FragmentTree`, not boxes in the tree.
+Any assertion about how often a repeating group is drawn has to be taken from the fragment tree for
+the same reason.
--- a/.claude/migration-notes/2026-09-08-a-relocated-table-no-longer-repeats-its-abandoned-header.md
+++ b/.claude/migration-notes/2026-09-08-a-relocated-table-no-longer-repeats-its-abandoned-header.md
@@ -1,0 +1,21 @@
+# A relocated table no longer repeats its abandoned header
+
+A table that carries `break-inside: avoid` (or `page-break-inside: avoid`) and does not fit in what
+is left of a page moves whole to the next one. When such a table also repeats a `<thead>`, the
+headers belonging to the abandoned attempt were drawn as well.
+
+**Before.** Three copies of the header row: one stranded near the bottom of the page the table had
+left, with no table beneath it, and two drawn exactly on top of each other at the top of the page
+the table moved to. Overprinted text looks bolder or slightly smudged rather than obviously
+doubled, so the usual first sign was the stray header on the previous page.
+
+**Now.** One header, at the top of the page the table lands on — which is what a browser draws.
+
+**When a document author would notice.** Only where all three hold: a table with a `<thead>`, an
+avoiding `break-inside` on it (a common house style for keeping a small table together, and note
+that the UA print stylesheet supplies it for the header group itself, not for the table), and a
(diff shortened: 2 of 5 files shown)


jhaygood86:
Reviewed the diff, the invariant it cross-references, and reproduced the fix locally (Windows, net8.0). The core change is sound: the new `ClearCapturedInstances(_tableBox)` call sits inside the *same* `if (!_continuesAPreviousPass)` block as `RestoreStructureFromAnyPreviousRun()`, so the resumed-vs-fresh distinction can't diverge between the two calls — that was my main worry going in, and it isn't one. The six relocation fixtures reproduce the described 3-header defect against `main` exactly as described (0/1 on the abandoned page, 1/2 on the landing page) and all pass on this branch. Build is clean (0 warnings).

However, `RelocatedTableHeaderInstanceTests.TheSameTableAllowedToSplit_StillRepeatsItsHeaderOnBothPages` — the control the PR relies on to prove the fix doesn't suppress repetition outright — fails for me both on `main` and on this branch, identically:

```
Assert.Equal() Failure: Values differ
Expected: 1
Actual:   0
```
at `perPage[0]` (line 88).

Dumping the fragment tree for that exact fixture (`PushedTable(130)` without `break-inside: avoid`) shows why: nothing lands on page 0 at all — the whole table (header + all three body rows) is deferred wholesale to page 1, same as the `break-inside: avoid` case at the same filler height:

```
Fragmentainer count: 2
Slot 0: words=[filler]
Slot 1: words=[HEADWORD,One,Two,Three]
```

So at `fillerHeight: 130` the ~30pt left on the page isn't enough room for even the header row, and the table moves as a unit regardless of `break-inside`. The control doesn't currently reach the split code path it's meant to exercise, so it can't distinguish "correctly still repeating" from "fix accidentally suppresses repetition" — it fails the same way either way. Since `.github/workflows/test.yml`'s matrix includes `windows-latest`, this looks likely to fail CI there too, which would contradict the "full suite green" claim in the description.

Worth re-tuning the control fixture's filler height (or asserting on headroom rather than a fixed constant shared with the avoid fixtures) so it reliably splits — with the header landing on page 0 — across platforms, then re-verifying it fails before the fix and passes after. Everything else here looks ready; this is the one thing I'd want resolved before merge.

dchupp:
Force-pushed: the Windows job failed on my own **control** test, not on the fix.

`TheSameTableAllowedToSplit_StillRepeatsItsHeaderOnBothPages` named page 1 and page 2 outright. On Windows the same fixture puts nothing at all of the table on the first page — the metrics there make the filler plus the first row overflow it — so the control expected 1 header on page 1 and found 0. The fix's own six cases passed on all three platforms.

Every assertion is now stated against the pages the table's own **rows** reach rather than against page numbers: exactly one copy of the group wherever a row of the table is, and none anywhere else. That is css-tables-3 §6.2 restated as a per-page equality, it says the same thing whatever the metrics do to where the table lands, and it is strictly stronger than what it replaces — a stranded copy fails the "none anywhere else" half, a doubled one fails the "exactly one" half. The relocation itself is asserted separately, as no rows on the page the table was pushed off.

The control is also no longer metric-dependent in the other direction: it is a thirty-row table on a 200pt page, which has to split whatever a row measures.

Still 6 of 7 failing against `main`, full suite green locally on net8.0.

jhaygood86 (approved):
Verified the updated test directly: `AssertOneGroupPerPageTheRowsReach` now asserts `headers == rows.Select(n => n > 0 ? 1 : 0)` — exactly one header copy wherever a row of the table actually lands, and none anywhere else — which is metric-independent (no hardcoded page numbers) and strictly stronger than what it replaces, matching css-tables-3 §6.2 directly. The production fix in `CssLayoutEngineTable.cs` is unchanged from my original review, since the bug was in the test's assumption about which page the table would land on, not in the fix itself.

Approving. Merging now.

PR #963, opened 2026-09-08
A forced break at the end of a block no longer opens a line

dchupp:
A forced line break ([css-text-3 §5.5](https://www.w3.org/TR/css-text-3/#forced-line-break)) **ends** the line it falls on. `CssLayoutEngine` instead closes that line and starts the next one with the break's own word, so a break was always *followed* by a line — including where the break is the last thing in the block and there is nothing for that line to hold.

Printing a finished block's line boxes shows it plainly:

| markup | line boxes | browser |
| --- | --- | --- |
| `<div>x<br></div>` | `['x'] \| [<BR>]` | one line |
| `<div><br></div>` | `[] \| [<BR>]` | one line |
| `<div><br><br></div>` | `[] \| [<BR>] \| [<BR>]` | two lines |

The break's word never shares a line with the content it follows, and the trailing one always has a line to itself — holding nothing. CSS 2.1 [§9.4.2](https://www.w3.org/TR/CSS21/visuren.html#inline-formatting) is explicit about such a line: one with "no text, no preserved white space, no inline elements with non-zero margins, padding, or borders, and no other in-flow content" must "be treated as not existing for any other purpose". A `<br>` is not the preserved newline that sentence carves out. Every such block came out one line taller than a browser makes it.

## The fix

`DropATrailingForcedBreaksOwnLine` takes that line back where the block finishes, moving the block's bottom up to the line the break actually terminated.

Only the last line, and only when it holds nothing but breaks — `x<br><br>` really is two lines and `x<br><br>c` three, so every break but the final one keeps the line it opened. A block with **no inline content at all** is excluded: its single wordless line box is not a break's leftover, and dropping it would collapse a zero-height block rather than shorten it.

## Measured against Chrome 152

Five shapes, all now matching to the point where before each was one line out:

| markup | before | now, and Chrome |
| --- | --- | --- |
| `x<br>` | 2 lines | 1 |
| `<br>` | 2 lines | 1 |
| `x<br><br>` | 3 lines | 2 |
| `x<br>` + a `display: block` sibling | 2 lines + the block | 1 + the block |
| `x<br><br>c` | 3 lines | 3 — unchanged |

The shape this hits in practice is a table cell holding a label above an image, `label<br>`: every row of such a table gained a line, and on a long one that is enough to push what follows onto another page.

## Tests

`TrailingForcedBreakLineTests` states each height as a multiple of the block's own line height, taken from a single-line control in the same document, so it asserts the line **count** and nothing about what a line measures. `line-height` is pinned in the fixture — a break-only line and a text line differ by a quarter of a point under `normal`, which is the font's own metrics and not this test's subject. **Five of the nine cases fail against `main`**; the four that do not are the controls (`x<br><br>c`, `x<br>c`, `x`, and the empty block).

Full suite green on net8.0 (10,396 passed), 0 new build warnings, diff coverage 100%. The `br` row in `docs/html-css-support.md` now states the rule, and `.claude/` carries a recent-fixes entry and a migration note.

### DIFF
--- a/.claude/migration-notes/2026-09-08-a-trailing-br-no-longer-adds-a-line.md
+++ b/.claude/migration-notes/2026-09-08-a-trailing-br-no-longer-adds-a-line.md
@@ -1,0 +1,20 @@
+# A trailing `<br>` no longer adds a line
+
+A `<br>` with nothing after it in its block used to leave an empty line behind it, making the block
+one line taller than a browser renders it. It now ends the line it falls on and opens none. CSS 2.1
+[§9.4.2](https://www.w3.org/TR/CSS21/visuren.html#inline-formatting) requires a line box holding no
+content to "be treated as not existing for any other purpose".
+
+| markup | before | now, and in a browser |
+| --- | --- | --- |
+| `x<br>` | 2 lines | 1 line |
+| `<br>` | 2 lines | 1 line |
+| `x<br><br>` | 3 lines | 2 lines |
+| `x<br>` then a `display: block` sibling | 2 lines + the block | 1 line + the block |
+| `x<br><br>c` | 3 lines | 3 lines — unchanged |
+
+**When a document author would notice.** Any block whose inline content ends in a `<br>` is now one
+line shorter. The shape that shows it most is a table cell holding a label above an image —
+`label<br><img style="display:block">` — where every row of the table loses a line's height, and a
+long table can gain back a page. Content that relied on a trailing `<br>` for spacing will close up;
+use a margin, or a second `<br>` (which still leaves one empty line, as it does in a browser).
--- a/.claude/recent-fixes/2026-09-08-a-forced-break-at-the-end-of-a-block-opened-a-line.md
+++ b/.claude/recent-fixes/2026-09-08-a-forced-break-at-the-end-of-a-block-opened-a-line.md
@@ -1,0 +1,45 @@
+# A forced break at the end of a block opened a line
+
+A forced line break ([css-text-3 §5.5](https://www.w3.org/TR/css-text-3/#forced-line-break)) *ends*
+the line it falls on. `CssLayoutEngine` instead closes that line and starts the next one with the
+break's own word, so a break was always followed by a line — including where the break is the last
+thing in the block and there is nothing for that line to hold. CSS 2.1
+[§9.4.2](https://www.w3.org/TR/CSS21/visuren.html#inline-formatting) requires such a line to "be
+treated as not existing for any other purpose", and a `<br>` is not the preserved newline that
+sentence carves out. Every such block came out one line taller than a browser makes it.
+
+`DropATrailingForcedBreaksOwnLine` takes that line back at the point the block finishes.
+
+## What measuring it turned up
+
+- **The line-box shapes say it plainly.** Printing every finished block's lines:
+  `<div>x<br></div>` is `['x'] | [<BR>]`, `<div><br></div>` is `[] | [<BR>]`, and
+  `<div><br><br></div>` is `[] | [<BR>] | [<BR>]`. The break's word never shares a line with the
+  content it follows, and the trailing one always has a line to itself.
+- **Only the last break, and only when its line holds nothing else.** `x<br><br>` really is two
+  lines in a browser and `x<br><br>c` three — measured against Chrome 152, along with `x<br>`,
+  a bare `<br>`, and `x<br>` in front of a `display: block` sibling. Every one of the five now
+  matches Chrome to the point.
+- **A block with no inline content is not this case** and is deliberately excluded: its single line
+  box has no words, and dropping it would collapse a block rather than shorten it.
+- **The common shape is a table cell holding a label and an image.**
+  `label<br><img style="display:block">` gained a whole line in every row of such a table. On a
+  seven-row routing table that is 70pt, enough to push the table after it onto a second page.
+
+## Evidence
+
+`TrailingForcedBreakLineTests`: seven content shapes asserted as a multiple of the block's own line
+height, taken from a single-line control in the same document, plus the block-sibling case and the
+empty-block exclusion. `line-height` is pinned in the fixture — a break-only line and a text line
+differ by a quarter of a point under `normal`, which is the font's metrics and not this test's
+subject. Five of the nine fail against `main`; the four that do not are the controls. Full suite
+green on net8.0 (10,387 before the new tests), 0 new build warnings.
+
+## Correction
+
+The first version of this cited CSS 2.1 §16.6.1 for "a forced break ends a line". That is wrong and
+the maintainer caught it: §16.6.1 is "The 'white-space' processing model", entirely about collapsing
+tabs, spaces and linefeeds, and it never mentions `<br>` or a forced break. Verified by fetching the
+section. The term is defined in css-text-3 §5.5, and the rule this fix actually implements — an empty
+line box "must be treated as not existing" — is CSS 2.1 §9.4.2. Both verified the same way before
+being written here.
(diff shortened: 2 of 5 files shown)


jhaygood86:
First-look review of this PR against CLAUDE.md's testing/spec-verification conventions. I built the branch (0 build warnings on `dotnet build PeachPDF.slnx -t:Rebuild`), hand-traced `DropATrailingForcedBreaksOwnLine` and its call site against `FlowBox`/`CreateLineBoxes`, and re-ran the tests. Two concrete problems, both worth fixing before merge:

**1. Two of this PR's own nine new fixtures fail on this branch.** Running `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0 --filter "FullyQualifiedName~TrailingForcedBreakLineTests"` as checked out:

```
Failed PeachPDF.Tests.Integration.TrailingForcedBreakLineTests.ABlockIsAsTallAsTheLinesABrowserGivesIt(content: "<br>", lines: 1)
  Expected: 13.301   Actual: 12
Failed PeachPDF.Tests.Integration.TrailingForcedBreakLineTests.ABlockIsAsTallAsTheLinesABrowserGivesIt(content: "<br><br>", lines: 2)
  Expected: 26.602   Actual: 25.301
```

This contradicts the PR description's "Full suite green on net8.0 (10,396 passed)". Root cause: for a shape whose leading content is *only* break(s) (no real text before the trailing break run gets dropped), the surviving line's height comes from the pre-existing `box.ActualLineHeight` fallback at the wrap-check just above the word loop (`CssLayoutEngine.cs` ~line 2789), not from a real word's measured height — and in this test fixture's font, those two disagree by about 1.3pt (a "text" line measures 13.3pt, an "abandoned seed line" bailout measures 12pt flat). So the fix correctly reduces the *line count* for `<br>` and `<br><br>`, but the resulting single/double line is the wrong *height* — short by roughly 10%. `x<br>`, `x<br><br>`, `x<br><br>c`, and the block-sibling case all pass correctly (verified by hand-tracing the word loop) because they have a real word establishing the first line's true height; only the "no text at all, just break(s)" shapes are affected.

I confirmed the pre-fix baseline claim is otherwise accurate: reverting just `CssLayoutEngine.cs` to `main` (keeping the new test file) reproduces exactly 5/9 failures as described.

**2. The CSS 2.1 §16.6.1 citation is wrong**, and it's repeated in four places: the PR body, `.claude/recent-fixes/2026-09-08-a-forced-break-at-the-end-of-a-block-opened-a-line.md`, `.claude/migration-notes/2026-09-08-a-trailing-br-no-longer-adds-a-line.md`, and the XML doc comments on `DropATrailingForcedBreaksOwnLine` and `TrailingForcedBreakLineTests`. I fetched the actual section (`https://www.w3.org/TR/CSS21/text.html#white-space-model`, mirrored verbatim at `drafts.csswg.org/css2/#white-space-model`) — it's entirely about white-space *collapsing* (tabs/spaces/linefeeds under the `white-space` property) and never mentions `<br>` or "forced line break" (a full-text search of the whole CSS2/2.1 spec for "forced line break" returns zero hits). The term "forced line break" is actually defined in the current, actively-maintained CSS Text Module Level 3, §5 ("Line Breaking and Word Boundaries") — `https://www.w3.org/TR/css-text-3/#forced-line-break` — which is the citation this change should use instead. (CSS 2.1 does touch the underlying idea, but only via an example in the `:first-line`/`:first-letter` section — "the first line of the P in this fragment: `<p><br>First...` does not contain any letters" — not at §16.6.1.)

Minor: a stray blank line was inserted between the XML doc comment and `private static void FinalizeLineBoxes(...)` (unrelated to this change's actual diff) — didn't trip a warning but is unclean and should be dropped.

On the positive side: the core mechanism is sound — I hand-traced all 5 of the PR's documented shapes plus the `<br><br>`-alone fixture against the actual `FlowBox` word loop and they check out; a broader regression pass (`--filter "FullyQualifiedName~Table|FullyQualifiedName~Flex|FullyQualifiedName~Acid2"`, 1319 tests) is fully green; and the fix only runs on the truly-final (non-fragmenting) completion pass, so it shouldn't interact with mid-document page breaks.

dchupp:
Force-pushed: the Windows job failed on the two `<br><br>` cases.

The fixture pinned `line-height: 12pt` at `font-size: 10pt` to take the font's own metrics out of the measurement. That was not clear enough of them: on Windows a **text** line at that size measures 13.3pt — the font's extent exceeding the declared line-height — while an empty, break-only line measures the declared 12pt exactly. The control line and the probe lines then differ by 1.3pt each, which is not what the test is about.

`line-height` is now 30pt, well clear of anything the font can ask for at 10pt, so every line box is exactly that height whatever it holds and the assertion is purely about how many lines there are.

No change to the fix itself. Still 5 of 9 failing against `main`, full suite green locally on net8.0.

jhaygood86:
Verified the test fix directly: `line-height: 30pt` at `font-size: 10pt` is well clear of the font's own metrics, so every line box is exactly 30pt regardless of content — the count-only assertion now measures what it says it measures. CI is green on all three platforms.

One thing from the original review wasn't addressed in this update, though: the CSS 2.1 §16.6.1 citation (`w3.org/TR/CSS21/text.html#white-space-model`) is still used in 5 places — the PR description, the recent-fixes note, the migration note, and two XML doc comments — and I independently re-verified it's wrong. I fetched that section directly: it's titled "The 'white-space' processing model" and is entirely about whitespace collapsing (tabs/spaces/linefeeds around preserved newlines); it never mentions `<br>`, "forced break", or line-ending behavior. The actual definition lives in CSS Text Module Level 3 §5 (`w3.org/TR/css-text-3/#forced-line-break`), which I also fetched and confirms: "When a line is broken due to explicit line-breaking controls... it is a forced line break."

This doesn't affect correctness — the fix itself and its tests are sound — but it's a citation five places will carry forward as fact, including into a future release note. Could you swap all five to the CSS Text 3 reference?

dchupp:
You are right, and thank you for pushing on it twice — I verified it myself rather than just taking the correction, and it is worse than a wrong section number.

I fetched `w3.org/TR/CSS21/text.html`: §16.6.1 is "The 'white-space' processing model", entirely about collapsing tabs, spaces and linefeeds. It never mentions `<br>` or a forced break. And I fetched css-text-3: §5.5 "Line Breaking Details" carries the `#forced-line-break` anchor and the definition — "When a line is broken due to explicit line-breaking controls... it is a forced line break."

But while checking, I found that css-text-3 §5.5 defines the *term* and says nothing about whether a trailing break generates a line box — so citing it alone for this change would have been a second wrong citation, just a subtler one. The rule this fix actually implements is **CSS 2.1 [§9.4.2](https://www.w3.org/TR/CSS21/visuren.html#inline-formatting)**:

> Line boxes that contain no text, no preserved white space, no inline elements with non-zero margins, padding, or borders, and no other in-flow content ... **and do not end with a preserved newline** must be treated as zero-height line boxes ... and **must be treated as not existing for any other purpose**.

The line a trailing `<br>` opens holds nothing, and a `<br>` is not the preserved newline that clause carves out — so §9.4.2 says outright it is not there. That is the load-bearing citation; css-text-3 §5.5 is now used only where the *term* "forced line break" appears.

All five places are updated: PR description, both `.claude/` notes, and the two XML doc comments. The recent-fixes note also records the wrong citation and why it was wrong, so it does not get re-derived. I dropped the stray blank line before `FinalizeLineBoxes` too.

Full suite green on net8.0 (10,396), 0 build warnings.

jhaygood86 (approved):
Independently re-verified this myself: fetched CSS 2.1 §9.4.2 directly and it quotes exactly — "Line boxes that contain no text, no preserved white space... and do not end with a preserved newline must be treated as zero-height line boxes... and must be treated as not existing for any other purpose." That's genuinely the load-bearing rule for this fix, more precise than the CSS Text 3 citation I'd originally suggested (which only defines the term, not the no-line-box consequence). Good catch going further rather than just swapping in what I proposed.

Approving. Merging now.

Issue #917, opened 2026-09-07
Pagination cost grows ~O(n^3) with page count: FragmentEmitter.CatchUpStaleSlotsBehind dominates on long documents


rgroenewoudt:
We are currently testing to see if PeachPDF can replace Playwright for us. We are running into an issue that it gets very slow with many pages.

## Summary

On a long, table-heavy paged document, render time per page grows with document length instead of staying
flat. The reproducer below emits N identical sections, so doubling N doubles the page count:

| Sections | Pages | Time | Per page | Factor per doubling |
|---------:|------:|-----:|---------:|--------------------:|
| 6  | 54  | 6.0 s   | 0.11 s | — |
| 12 | 108 | 33.3 s  | 0.31 s | **5.6×** |
| 24 | 216 | 402.3 s | 1.86 s | **12.1×** |

4× the pages costs **67×** the time — an exponent around 3. A second run of the same three files on the
same machine gave 8.0 s / 50.8 s / 376.7 s, so the per-doubling factor is noisy (5.6–12×) but the overall
shape is not. Peak working set over the range is 153 MB → 299 MB, so this is not allocation or GC pressure.

CPU profiling puts the entire superlinear component in the fragmentation path, under
`FragmentEmitter.CatchUpStaleSlotsBehind` and the `ChildrenOf` enumeration beneath it. CSS parsing, the
cascade and table layout are all essentially linear over the same range.

This looks like the same *class* of problem as #573 (`PassRewind.RollBackTo` resetting an unbounded tail
of a container's children on every page/attempt, with `ResetForRefill()` walking each subtree), but in
ordinary paged block flow rather than multi-column. I have not read the fragmentation code, so that is a
guess from the shape of the profile rather than a diagnosis.

## Environment

- PeachPDF 0.9.17
- .NET 10.0 (`net10.0`), Debug build
- Windows 11 x64
- `PageSize.A4`, `DownscaleImages = false`, `FileUriNetworkLoader`; nothing else configured

## Profile

`dotnet-trace collect --providers Microsoft-DotNETCore-SampleProfiler`, converted to speedscope and
aggregated over the render thread with idle thread-pool threads excluded. Inclusive times, from traced
runs, so the absolute numbers are higher than the untraced ones above — the point is the growth column.

These traces are from our real report rather than the generated reproducer — the same document at 43 and
117 pages (2.7× the pages). I can share those two HTML files too (text scrubbed, self-contained) if they
are more useful than the generator.

[...]

PR #965, opened 2026-09-08
A page re-emitted behind the frontier no longer skips its own content

dchupp:
`ChildrenOf` starts a container's child walk past a prefix of children each already confirmed `EmittedNothingAtOrBefore`, caching how many in `CssBox._liveChildStartIndex`. That answer is **"this subtree's fragments are all behind slot S"** — a claim about S and every slot *after* it, and about no slot before it. The cache stored the layout generation but not S.

`CatchUpStaleSlotsBehind` re-emits a frozen slot *behind* the frontier whenever a late relocation reopens one. It read the cursor there and skipped children whose content belonged to exactly that slot, so their words were in no fragmentainer at all — content drawn on no page, with a gap of its own height where it had been laid out.

## Measured

On the document that found this, slot 0 was emitted three times:

| emission | words in the root draft |
| --- | --- |
| first | 169 |
| after an early reopen | 169 |
| after a relocation at slot 6 reopened it | **51** |

At that third emission the prose container's cursor read **4** of its 146 children. The 660 characters those four held were drawn on no page, above 420pt of blank at the top of page 1.

## Why it survived

`BuildDraft`'s own `EmittedNothingAtOrBefore` check is already slot-aware and **never fired at slot 0** — instrumented and confirmed, not a single hit. Only the cache derived from it was wrong. The guarded check and the unguarded cache can only disagree on a path that runs backwards, and `CatchUpStaleSlotsBehind` is the only one.

Forcing `_forcingUnprunedReferenceWalk` on that path also fixes it, which is how the cursor was identified — but that is the expensive full walk the method exists to avoid (`Finish`'s own replay is ~145,000 `BuildDraft` calls per stale slot on a large document), so it was a diagnosis rather than a fix.

Worth noting separately: `PEACHPDF_VERIFY_FRAGMENT_PRUNING` could not have caught this. `VerifyAgainstTheFullWalk` runs only where `EmitSlot` is called with `mayVerify: true`, and the stale-slot catch-up is not one of those calls — precisely the one path that reads a pruning cache at a lower slot than it was written at. I have left that as an invariant note rather than widening the oracle here, since that is its own change with its own risk.

## The fix

`LiveChildStartFor(slot)` returns the cached index only where `slot` is at or after the slot it was derived at; `RecordLiveChildStart` stores it; `DiscardEmittedNothing` resets it alongside the index and generation it already reset. Forward emission is unaffected — an increasing slot always satisfies the guard — so the #917 pruning work this builds on keeps its benefit.

## Tests

`StaleSlotChildCursorTests` asserts the one thing that is true whatever the pagination: **every word the document laid out is claimed by some fragmentainer**. Nothing in it names a page number or a page count, so it says the same thing on every platform — the trap `UnreachedWordClaimTests` avoids for the same reason, and whose invariant this is a one-sided form of (a repeated running element genuinely is claimed more than once, so this asks only that nothing is claimed *less* than once).

The fixture is headings and paragraphs in a small page; the UA print stylesheet's `h1-h6 { break-after: avoid }` supplies the relocations. Four paragraph-length sets paginate differently; **three lose words against `main` — 74, 242 and 221** — and the fourth is the case where nothing is disturbed.

Full suite green on net8.0 (10,391 passed), 0 new build warnings, diff coverage 100%.

### DIFF
--- a/.claude/invariants/fragmentation-a-pruning-answer-is-about-one-slot-and-every-later-one.md
+++ b/.claude/invariants/fragmentation-a-pruning-answer-is-about-one-slot-and-every-later-one.md
@@ -1,0 +1,25 @@
+# A pruning answer is about one slot and every later one
+
+`EmittedNothingAtOrBefore(S)` says "this subtree's fragments are all behind slot S". That is a claim
+about **S and every slot after it** — never about a slot before it, where the same subtree may hold
+real content. Every cache derived from that answer has to carry the slot it was taken at, or it
+silently answers the wrong question.
+
+`_liveChildStartIndex` did not. It records how many of a box's leading children were each confirmed
+`EmittedNothingAtOrBefore`, so a later slot's walk can start past them — and it stored the
+generation but not the slot. Emitting an *earlier* slot again is not hypothetical:
+`CatchUpStaleSlotsBehind` exists to do exactly that whenever a late relocation reopens a frozen
+fragmentainer.
+
+**The measured symptom.** A fifteen-page document whose page 1 was emitted three times — 169 words,
+169 words, then **51**. The prose container's cursor stood at 4 when slot 0 was rebuilt, so its
+first four children were skipped and their text appeared on no page at all: 660 characters gone,
+and a 420pt hole at the top of page 1 where they had been laid out.
+
+Two things follow for the next such cache:
+
+- **Store the slot next to the index.** Trust the cache only where the slot now being built is at or
+  after the slot the cache was derived at. Dropping to a lower slot must re-derive.
+- **The parity oracle does not cover this.** `VerifyAgainstTheFullWalk` runs only where `EmitSlot` is
+  called with `mayVerify: true`, and the catch-up re-emission is not — which is precisely the one
+  path that reads a cache at a lower slot than it was written at.
--- a/.claude/migration-notes/2026-09-08-content-dropped-from-a-re-emitted-page-comes-back.md
+++ b/.claude/migration-notes/2026-09-08-content-dropped-from-a-re-emitted-page-comes-back.md
@@ -1,0 +1,16 @@
+# Content dropped from a re-emitted page comes back
+
+A page whose content a later relocation disturbed is emitted again once layout settles. That second
+emission could skip a run of leading children in a container, and whatever they held was then drawn
+on no page at all — leaving a visible blank where the content had been laid out.
+
+**Before.** On a long document, the first few blocks of a container could be missing from the
+output entirely, with a gap of exactly their height in their place. Measured on a fifteen-page
+document: 660 characters absent and 420pt of blank at the top of page 1.
+
+**Now.** They are drawn.
+
+**When a document author would notice.** Only in a document long enough for a page to be finalised
+and then reopened — several pages at least, with something later in the flow (a `break-inside:
+avoid` box, a widows/orphans push, a table moved off a boundary) moving content that had already
+been placed. Nothing needs changing in a document; output that was silently missing text now has it.
--- a/.claude/recent-fixes/2026-09-08-a-live-child-cursor-outlived-the-slot-it-was-taken-at.md
+++ b/.claude/recent-fixes/2026-09-08-a-live-child-cursor-outlived-the-slot-it-was-taken-at.md
@@ -1,0 +1,36 @@
+# A live-child cursor outlived the slot it was taken at
+
+`ChildrenOf` starts a container's child walk past a prefix of children each already confirmed
+`EmittedNothingAtOrBefore`, caching how many in `CssBox._liveChildStartIndex`. The cache carried the
+layout generation but not the **slot** the answer was taken at, and "nothing at or after slot 6" is
+not "nothing at or after slot 0". `CatchUpStaleSlotsBehind` re-emits a frozen slot behind the
+frontier, read the cursor there, and skipped children whose content belonged to exactly that slot.
+
+`LiveChildStartFor(slot)` now returns the cached index only where `slot` is at or after the slot the
+index was derived at, and `RecordLiveChildStart` stores it.
+
+## What measuring it turned up
+
+- **The counts name the defect on their own.** On the document that found this, slot 0 was emitted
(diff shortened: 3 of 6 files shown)


jhaygood86:
Reviewed against CLAUDE.md conventions (build, tests, CI, spec/architecture cross-check). Overall this looks correct and well-evidenced — verified independently rather than taking the PR body's claims on faith:

**Mechanism verified by tracing through the code.** `LiveChildStartFor(slot)`/`RecordLiveChildStart(index, slot)` in `CssBox.cs` and their call sites in `FragmentEmitter.ChildrenOf` (~line 2596-2630) do what the PR says: every `ChildrenOf` call re-stamps `_liveChildStartSlot` to the *query* slot, even when the cached index doesn't advance further — so as ordinary forward emission proceeds through higher slots, a box's cursor slot naturally climbs past `CatchUpStaleSlotsBehind`'s eventual (lower) reopen target, and the `slot >= _liveChildStartSlot` guard reliably falls back to a full re-walk from 0 in exactly the reopen case. Confirmed this is the crux mechanism, not just a plausible-looking guard direction.

**Bug reproduction confirmed independently**, not just trusted from the PR body: reverted only the `CssBox.cs`/`FragmentEmitter.cs` fix (keeping the new test) and re-ran `StaleSlotChildCursorTests` — 1 of 4 fixtures failed pre-fix on this machine (28 words dropped), consistent with the PR's own "on this machine" font-dependent caveat (it claims 3/4 on the author's machine). All 4 pass with the fix applied.

**Cross-checked against #931's pruning machinery for the same cache-invalidation bug class** (per CLAUDE.md's post-change review pass): the `ChildrenOf` geometric "ahead of slot" skip (`childBox.OwnGeometryTop() >= container.PageTopOf(slot.Index + 1)`) is explicitly *not* cached — recomputed on every visit, with a comment noting this was deliberately fixed from an earlier cached version that had the identical class of bug. `CommitGeometricallySettledObservations`/`_emptySincePass` → `RecordEmittedNothingAt` already carries its own per-box `sinceSlot` correctly. No other slot-scoped cache in either file lacks a stored validity slot — `_liveChildStartIndex` was the only gap.

**Build**: 0 warnings (`dotnet build -t:Rebuild`). **Tests**: full net8.0 suite green, 10,391 passed (matches the PR body exactly). Broader multicol/table/pagination/column filter: 1265 passed. CI green on all 3 platforms. Confirmed the PR branch is properly downstream of #931 and current main (`git merge-base --is-ancestor` checks out clean) — no bypass/revert risk.

**Minor, non-blocking**: two comments still say `CssBox.LiveChildStart` after the rename to `LiveChildStartFor` — `CssBox.cs` line ~2551 ("via LiveChildStart once that child...") and `FragmentEmitter.cs` line ~2581 ("see CssBox.LiveChildStart's own remarks"). Not XML-doc `<see cref>`s so they don't break the build, but worth a follow-up touch so they don't read as referring to a symbol that no longer exists.

**Worth taking seriously, flagged prominently rather than as a blocker**: the PR's claim that `PEACHPDF_VERIFY_FRAGMENT_PRUNING`'s oracle can't catch this class of bug is verified — `mayVerify: true` is only ever passed from `EmitPass`; `CatchUpStaleSlotsBehind` and `Finish`'s replay both call `EmitSlot` without it. Given the failure mode here is *silent, undetected data loss in a real PDF* and this is the second time a cache in this exact file has shipped this bug class (the geometric-skip comment references a prior instance caught only by two integration tests happening to fail, not by the oracle), I'd suggest filing a tracking issue for widening `VerifyAgainstTheFullWalk` to cover the catch-up/replay paths, even though it's reasonable to keep it out of *this* PR's blast radius. The current mitigation is "we now have an invariant note reminding the next author" — real, but weaker than an oracle that would have caught this automatically the way it already does for the forward-pass case.

jhaygood86 (approved):
Reviewed this one with extra scrutiny given the failure mode — silently dropped content in a real generated PDF is about as bad as a defect gets, and it's in the deepest, most fragile part of this codebase's fragmentation engine, built directly on top of #931's pruning work merged earlier today.

Traced the actual mechanism rather than trusting the description: every `ChildrenOf` call re-stamps `_liveChildStartSlot` to the query slot, so as ordinary forward emission climbs through higher slots, a box's cached "valid-from" slot naturally rises past whatever lower slot `CatchUpStaleSlotsBehind` later reopens — which is what makes the `slot >= _liveChildStartSlot` guard reliably catch the reopen case while leaving forward-pass pruning untouched. Independently reproduced the bug (reverted just the two production files, kept the new test, confirmed a fixture failed pre-fix and passes with the fix restored). Confirmed against #931's own `ChildrenOf` code that the geometric "ahead of slot" skip was already deliberately left uncached for this exact reason (there's a comment noting an earlier cached version had this bug class) — so `_liveChildStartIndex` was genuinely the one remaining gap, not a symptom of something broader left unfixed. Confirmed the branch is properly downstream of both #931 and current main.

Agree with the (non-blocking) suggestion that this is the second time this file has shipped this specific bug class (an answer cached without the slot it's valid from) — worth a tracking issue for widening `PEACHPDF_VERIFY_FRAGMENT_PRUNING`'s oracle to cover the `CatchUpStaleSlotsBehind`/`Finish`-replay paths, even though I agree it's reasonable to not fold that into this PR.

Build clean, full suite matches the claimed count exactly (10,391), CI green on all three platforms.

Approving. Merging now.

dchupp:
Both follow-ups from your review are done:

- The two stale `CssBox.LiveChildStart` comments are fixed in [#969](https://github.com/jhaygood86/PeachPDF/pull/969) (comments only).
- The oracle gap is filed as [#970](https://github.com/jhaygood86/PeachPDF/issues/970). I agree it is the weaker mitigation as it stands — an invariant note is a convention, not a check. The issue records what I found while confirming your point: it is not a one-line `mayVerify: true`, because the catch-up and replay paths deliberately run under `_pruningSuspended`/`_forcingUnprunedReferenceWalk` (a replay may not draw new pruning conclusions), so it needs a differential that compares without writing new marks. Forcing the unpruned walk on the catch-up path does make the #965 defect vanish, so an unpruned reference walk is a sufficient oracle for this class — just the expensive one. Happy to attempt it if you want it.

PR #967, opened 2026-09-09
The cascade no longer re-defaults every list item's anonymous box

dchupp:
The cascade's defaulting step walks `CssDefaults.InitialValues` and sets every property back to its initial value. It is already skipped for a box still on the shared `ComputedStyle.Default` — but an **anonymous** box that box generation gave a structural display to has *already* forked off `Default`, so it took the full walk to rebuild a state it was one assignment away from.

Re-pointing at `ComputedStyle.Default` and restoring the display gives that state directly. Every area is copy-on-write, so the display assignment forks it straight back. An element-backed box still takes the loop — it has author declarations to apply over the top.

**Why it is safe** — corrected from the first version of this PR, which said an anonymous box reaches the loop "for exactly one reason: box generation assigned its display structurally". You were right that that is wrong, and instrumenting it makes it stark: of the **2,031** boxes reaching the branch over the corpus, **2,005 still hold the *initial* display**. They are pseudo-elements forked by `CssData`'s synthesis calling `InheritStyle` on them, out of band, before their own cascade pass runs.

One refinement to your finding: on the real corpus the dominant case is not `::marker` but **`::before`/`::after` — 1,971 of them, against 34 markers**. Same mechanism you identified (`InheritStyle` at synthesis, lines 906/918/946), different pseudo-element; `::marker` dominates the *list* fixture this PR's test uses (100 of 101), which is presumably what made it look dominant.

The real invariant is the one you stated: this is a **full state overwrite, equivalent to the old loop for any prior divergence**, not for one known cause. The loop discarded every non-display property unconditionally too, and `box.InheritStyle()` re-runs immediately after this regardless of branch. That framing is now in the code comment, the invariant file and the recent-fixes note.

The premise is already yours: [`CascadeDefaultingLoop_OnAFreshBox_OnlyKnownExceptionsAreNotNoOps`](../blob/main/src/PeachPDF.Tests/Html/Core/Dom/ComputedStyleTests.cs) establishes that setting a property to its own initial value leaves a box on `Default`.

## The shape that reaches it is a list

I counted the branch across eight markup shapes. Every one of these takes it **once**:

anonymous blocks from mixed inline/block content · a well-formed table · a table missing its `<tr>`s · loose text inside a table · `display: table-cell` outside a row · a float between two text runs · a plain `<div>` wall

A hundred `<li>` take it **101 times**. It is one forked anonymous box per list item, and `list-style: none` does not avoid it.

On real documents that is most of the cascade's allocation. Over a 26-document corpus of ordinary business documents the branch is taken **2,031 times**, against 3 for element-backed boxes, and this change takes the corpus from **930 MB to 818 MB** allocated per pass — **12% of the total** — across four alternating measurements (929.85 / 929.92 against 818.40 / 818.35).

## A warning about measuring this one

The first three fixtures I wrote showed **no difference at all**, because none of them generates a *forked* anonymous box. There is also an older note in my own records concluding this path was "no longer reached", from a synthetic fixture of 1,681 anonymous boxes measuring 53.88 MB against 53.83 MB per render. Both were true of the fixture and false of the engine. If you want to check this yourself, put a counter behind the `if` and run real documents through it — an allocation delta of zero proves the fixture misses the path, not that the path is cold.

## Test

`AnonymousBoxDefaultingTests` compares 400 list items against the same text in plain `<div>`s, **both measured in the same run**, so the comparison is self-calibrating and says the same thing whatever font a machine resolves:

| | list vs plain |
| --- | --- |
| with this change | **1.71x** |
| without | **2.41x** |
| bound | 2.05x |

About 20% of margin either side. The item text is one character deliberately: the saving is a fixed cost per item, so longer text dilutes it — at 200 items of a full sentence the window narrows to 1.48 against 1.60, too tight to gate on.

**The flakiness you hit is fixed, and it was a real bug in the test.** It used `GC.GetTotalAllocatedBytes`, which is **process-wide**, in a suite that runs collections in parallel — so it was counting whatever else was allocating at that moment. Stable in isolation, meaningless with the suite running. I reproduced your numbers exactly:

| counter | three full-suite runs |
| --- | --- |
| `GetTotalAllocatedBytes` (process-wide) | 1.50x, 4.89x, 2.26x |
| `GetAllocatedBytesForCurrentThread` | 1.71x, 1.71x |

The absolutes give it away too — the process-wide runs reported 1062 MB, 647 MB and 132 MB for a fixture that actually costs 53 MB. The measured region is now synchronous (`GetAwaiter().GetResult()`) so it stays on the thread it counts.

Full suite green on net8.0 (10,388 passed), 0 new build warnings, diff coverage 100%.

### DIFF
--- a/.claude/invariants/cascade-a-synthetic-fixture-can-miss-a-path-real-documents-take.md
+++ b/.claude/invariants/cascade-a-synthetic-fixture-can-miss-a-path-real-documents-take.md
@@ -1,0 +1,31 @@
+# A synthetic fixture can miss a path real documents take
+
+Twice the same wrong conclusion has been reached about the cascade's defaulting loop, in opposite
+directions, and both times from a fixture written to exercise it.
+
+- It was once recorded as **no longer reached** on `main`, from a fixture of 1,681 anonymous boxes
+  where allocation measured 53.88 MB against 53.83 MB per render.
+- Writing the fix, three fresh fixtures — anonymous blocks from mixed inline/block content, tables,
+  flex/grid utility markup — again showed **no difference at all**.
+
+Both were true of the fixture and false of the engine. The real corpus takes that branch **2,031
+times** and the patch is worth **112 MB per pass**, 12% of everything `main` allocates.
+
+**So: count the branch before concluding it is cold.** A counter behind the `if` and one run over
+real documents is minutes of work and is the only thing that distinguishes "this path is not hot"
+from "my fixture does not reach this path". An allocation delta of zero proves the second, not the
+first.
+
+## And count it before explaining *why* it is reached
+
+The same counter, kept in place a little longer, would have caught a wrong explanation as well as a
+wrong measurement. The fix's first justification said an anonymous box reaches the loop "for exactly
+one reason: box generation assigned its display structurally". Instrumented over the corpus, **2,005
+of the 2,031 still hold the initial display** — they are pseudo-elements (1,971 `::before`/`::after`,
+34 `::marker`) that `CssData`'s synthesis forked by calling `InheritStyle` on them, out of band,
+before their own cascade pass ran.
+
+The fix was still correct, but for a broader reason than the one written down: it is a **full state
+overwrite, equivalent to the old loop for any prior divergence**, not for one known cause. A
+justification that names a single mechanism is a claim about every box that reaches the code, and
+it is as measurable as the hit count is.
--- a/.claude/recent-fixes/2026-09-08-the-cascade-re-defaulted-every-list-items-anonymous-box.md
+++ b/.claude/recent-fixes/2026-09-08-the-cascade-re-defaulted-every-list-items-anonymous-box.md
@@ -1,0 +1,52 @@
+# The cascade re-defaulted every list item's anonymous box
+
+The cascade's defaulting step walks `CssDefaults.InitialValues` and sets each property back to its
+initial value. It is skipped for a box still on the shared `ComputedStyle.Default`, which is most of
+them — but an **anonymous** box that anything has already forked off `Default` took the full walk to
+reconstruct a state it was one assignment away from.
+
+Re-pointing at `ComputedStyle.Default` and restoring the structural display is that same state,
+directly. Every area is copy-on-write, so the display assignment forks it straight back.
+
+## What measuring it turned up
+
+- **The reason they are forked is not what this first claimed.** The original justification said box
+  generation had assigned the display structurally. Instrumented over the corpus, 2,005 of the 2,031
+  hits still hold the *initial* display: they are pseudo-elements (1,971 `::before`/`::after`, 34
+  `::marker`) forked by `CssData`'s synthesis calling `InheritStyle` on them before their own cascade
+  ran. The fix is safe for a broader reason — it is a full state overwrite, equivalent to the old
+  loop for **any** prior divergence, and `InheritStyle` re-runs immediately after it either way.
+  Caught in review, and only because the reviewer instrumented the branch rather than reading the
+  comment.
+- **The shape that hits it is a list, and nothing else I tried.** Counting the branch over eight
+  markup shapes — anonymous blocks from mixed inline/block content, a well-formed table, a table
+  missing its rows, loose text inside a table, `display: table-cell` outside a row, floats — every
+  one took it **once**. A hundred `<li>` took it **101 times**. It is one forked anonymous box per
+  list item, and `list-style: none` does not avoid it.
+- **On real documents that is the bulk of the cascade's allocation.** Across a 26-document corpus
+  the branch is taken **2,031 times** against 3 for element-backed boxes, and the patch takes the
+  corpus from **930 MB to 818 MB** of allocation per pass — 12% of everything, reproduced across
+  four alternating measurements (929.85/929.92 against 818.40/818.35).
+- **A synthetic fixture nearly missed it entirely.** The first three I wrote — anonymous blocks,
+  tables, flex/grid utility markup — showed no difference at all, because none of them generates a
+  *forked* anonymous box. Had I stopped there I would have concluded the path was cold, which is
+  the opposite of what the corpus says.
+- **The premise is already tested.** `ComputedStyleTests.CascadeDefaultingLoop_OnAFreshBox_OnlyKnownExceptionsAreNotNoOps`
+  establishes that setting a property to its own initial value leaves a box on `Default`. That is
+  exactly why re-pointing at `Default` reconstructs what the loop reconstructs.
+
+## Evidence
+
+`AnonymousBoxDefaultingTests` compares 400 list items against the same text in plain `<div>`s,
+measured in the same run so the comparison is self-calibrating and says the same thing whatever font
+a machine resolves: **1.72x with the fast path, 2.41x without**, bound at 2.05x. The item text is one
+character deliberately — the saving is fixed per item, so longer text dilutes it (at 200 items of a
+sentence the window narrows to 1.48 against 1.60, too tight to gate across platforms).
+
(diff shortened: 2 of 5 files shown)


jhaygood86:
First-look review. Verified: clean rebuild (0 warnings), full net8.0 suite green apart from one test noted below, and I independently traced the fast-path condition by instrumenting the branch and running the actual `<ul><li>x</li>...</ul>` shape the PR targets.

## The core correctness claim needs a correction, though the code is still safe

The comment on `CascadeApplyStyles` and the PR body both say an anonymous box reaches the fast path "for exactly one reason: box generation assigned its display structurally." I instrumented the branch and ran it against real `<li>` markup — **the dominant real-world case (a `<li>`'s synthesized `::marker` box) actually forks off `Default` via a *different* mechanism**: `CssData.cs:946` calls `markerPseudoBox.InheritStyle(box)` during selector-match-time pseudo-element synthesis, *before* the marker's own `CascadeApplyStyles` call ever runs. At that point `Display` is untouched (still Default's own "inline"). So the box that dominates the measured 2,031-hit corpus is forking on **inherited-property divergence**, not a structural `Display` write.

This doesn't make the fix wrong — I worked through why: `box.InheritStyle()` (step 2) unconditionally re-runs immediately after the fast path regardless of which branch step 1 took, so whatever the premature `InheritStyle` call left behind gets correctly re-derived anyway; and the *old* full loop already discarded every non-`Display` property unconditionally too, regardless of what caused the fork. So "re-point at `Default`, restore `Display`" is equivalent to the old loop for *any* fork reason, not just the one named. But the stated justification is factually wrong about the mechanism it's built on, in code explicitly flagged as correctness-critical — and this repo has already been bitten once by trusting an imprecise mental model here (see `.claude/recent-fixes/2026-08-17-table-cellpadding-border-cascade-defeated-by-defaulting-loop.md`'s "load-bearing lesson" about premature out-of-band writes before a box's own cascade pass). Worth fixing the comment (and the new invariant file, which repeats the same "structurally pre-assigned Display" framing) to state the real, broader invariant: the reset is safe because it's a full state overwrite equivalent to the old loop for *any* prior divergence, not because the divergence has a single known cause.

## `AnonymousBoxDefaultingTests.AListCostsLittleMoreThanTheSameTextWithoutOne` is flaky here

Ran it 3 times locally (Windows): **3.70x, 2.05x (borderline), 4.59x** against the claimed stable ~1.72x/2.05x bound with "<0.2% run-to-run spread." `gh pr checks` shows green on all three CI platforms including windows-latest, so this may be specific to my machine, but the numbers directly contradict the "self-calibrating, says the same thing on every platform" claim in the PR body — the spread I saw is two orders of magnitude larger than claimed. Worth a second look at whether the bound has enough margin, or whether an allocation-ratio gate is the right mechanism here vs. a counter-based assertion on the fast path itself.

Everything else checked out: `HtmlTag is null` boxes created by post-cascade correction passes (`CorrectInlineBoxesParent`, `CorrectAnonymousTables`, `CorrectAbsolutelyPositionedInlineElements`) never re-enter `CascadeApplyStyles`, so they're unaffected by this change; `@container` re-cascade reparses the tree from scratch each pass (no stale-cache risk); and no truly-anonymous box gets author styles beyond the marker/first-letter/first-line cases already exercised.

dchupp:
Both points were right, and both are fixed. Thank you — the second one was a genuine bug in the test, not a machine difference.

## The justification was wrong

I instrumented the branch rather than take the correction on trust, and it is starker than "not always": of the **2,031** boxes reaching it over the corpus, **2,005 still hold the *initial* display**. So "box generation assigned its display structurally" was wrong for 98.7% of them, in a comment flagged as correctness-critical.

One refinement to your finding, from the same instrumentation: on the real corpus the dominant case is **`::before`/`::after` — 1,971 of them — against 34 `::marker`**. Same mechanism you identified (`InheritStyle` at synthesis; it is called at lines 906, 918 and 946 alike), different pseudo-element. `::marker` dominates the *list* fixture this PR's test uses, 100 of 101, which is presumably why it looked dominant from there.

The comment, the invariant file and the recent-fixes note now all state the invariant you articulated: a **full state overwrite, equivalent to the old loop for any prior divergence**, because the loop discarded every non-display property unconditionally too and `InheritStyle` re-runs immediately after either way. The invariant file also gained a second half — count the branch before explaining *why* it is reached, not just before claiming it is hot.

## The flakiness was mine

It was not your machine. The test read `GC.GetTotalAllocatedBytes`, which is **process-wide**, and this suite runs collections in parallel — so it was measuring whatever else happened to be allocating. I reproduced your spread:

| counter | three full-suite runs |
| --- | --- |
| `GetTotalAllocatedBytes` (process-wide) | **1.50x, 4.89x, 2.26x** |
| `GetAllocatedBytesForCurrentThread` | **1.71x, 1.71x** |

The absolutes make it obvious in hindsight — the process-wide runs reported 1062 MB, 647 MB and 132 MB for a fixture that costs 53 MB. The measured region is now synchronous (`GetAwaiter().GetResult()`) so it stays on the thread it counts, and it is stable to the second decimal both in isolation and with the whole suite running around it. Mutation-checked again after the change: 2.41x against `main`, 1.71x with the fix, bound 2.05.

(#968 already used the per-thread counter — I hit this same trap there first and did not come back to fix it here. Apologies for the wasted review cycle.)

Full suite green on net8.0 (10,388), 0 build warnings.

jhaygood86 (approved):
Verified both fixes directly. The DomParser.cs comment now says outright "and it is not 'because the only divergence is a structural display write' - that was the original justification here and it is wrong," with the real instrumented numbers and the mechanism-independent safety argument (full overwrite, equivalent to the old loop for any prior divergence, since `InheritStyle` re-runs immediately after either branch). That's the correction I was hoping for — genuinely more useful than fixing the framing to match one dominant cause, since the *next* pseudo-element or synthesis path that forks a box for some other reason still won't need re-litigating this.

One thing worth a heads-up rather than a block: `AListCostsLittleMoreThanTheSameTextWithoutOne` passes CI on all three platforms, but on my machine it's a stable, repeatable 2.06x against the 2.05x bound (not wild swings like the process-wide-counter bug — same result across 4 runs). This machine has documented hardware quirks elsewhere in this project's history, so I don't think it's the same class of problem you already found and fixed, but the margin is tight enough that it might be worth a little more headroom on the bound at some point. Not blocking on it.

Build clean, cascade/anonymous-box/computed-style filter passes (203 tests, this one aside), CI green on all three platforms.

Approving. Merging now.

dchupp:
On the 2.06x heads-up: I could not reproduce it here, so instead of widening the bound I went looking for why our two machines disagree at all — and the test was wrong about its own premise.

Comparing one list document against one plain document cancels what the two documents *share*, but the fixed cost of rendering anything at all — document setup, font loading, page machinery — sat in **both** numerator and denominator. A constant added to both does not cancel in a ratio; it drags it toward 1 by however much a given machine charges for it. So the number moved with the baseline, which is precisely the machine-dependent term it was meant to be immune to. Your 2.06 against my 1.71 looks like that, not a hardware quirk.

[#975](https://github.com/jhaygood86/PeachPDF/pull/975) measures each shape at two sizes and takes the ratio of the **slopes** instead — the baseline drops out of each slope, and dividing the list slope by a plain block's cancels the per-item text work too:

| | list | plain block | ratio |
| --- | --- | --- | --- |
| with the fast path | 102.3 KB/item | 54.4 KB/item | **1.88x** |
| without | 157.3 KB/item | 54.8 KB/item | **2.87x** |

Bound 2.35x, window 53% wide against the old form's 41%. No change to the code under test.

I can't verify it fixes it on your machine, only that it removes the term I think caused the split — if it still lands near the bound for you, that is worth knowing, because it would mean the per-item cost itself differs between us, which is a different and more interesting problem.

PR #968, opened 2026-09-09
The GPOS subtable walk no longer allocates an enumerator per glyph

dchupp:
Every lookup's `Subtables` is declared `IReadOnlyList<T>`, so a `foreach` over it boxes an enumerator through the interface. `GposPositioner` does that **once per glyph** — `ApplyMarkToBase` calls straight into the loop for every glyph in the run — so the cost lands on every glyph of every text run in a document.

It fires on **plain Latin text**, where there is nothing to attach at all, because the enumerator is allocated before any coverage check runs.

Eight such walks are now indexed. The ninth `foreach` in the file is deliberately left alone: it walks a `GposSequenceLookupRecord[]`, and an array's enumerator is a struct that allocates nothing.

## Measured

Over 26 real business documents, allocation per pass:

| | MB |
| --- | --- |
| `main` | 930.16 |
| with this change | **844.75** |

**−85 MB, 9% of the total.** No behaviour change — the loop visits the same subtables in the same order.

## Test

`GposPositionerAllocationTests` calls `ApplyMarkToBase` directly against a lookup whose single subtable covers no glyph, so the only thing that can allocate is the walk itself:

| | allocated over 10,000 glyph visits |
| --- | --- |
| `main` | **400,000 bytes** — exactly 40 per visit |
| with this change | 0 |

Two things about that test are worth flagging, because both bit me first:

- **It needs a non-empty subtable list.** `List<T>` hands back a *cached singleton* enumerator when it is empty, so a lookup with no subtables allocates nothing and the test passes against `main`. One subtable whose coverage matches nothing keeps the loop body cold while still walking.
- **It reads `GC.GetAllocatedBytesForCurrentThread`, not `GC.GetTotalAllocatedBytes`.** The latter is process-wide, and this suite runs collections in parallel — the first full-suite run reported 1.2 MB of other tests' allocation and failed a passing build.

It asserts at the positioner rather than through a rendered document on purpose: allocation over a document scales with whatever font a machine resolves, so an end-to-end threshold is loose on one platform and wrong on another.

Full suite green on net8.0 (10,388 passed), 0 new build warnings, diff coverage 100%.

### DIFF
--- a/.claude/recent-fixes/2026-09-08-the-gpos-subtable-walk-allocated-an-enumerator-per-glyph.md
+++ b/.claude/recent-fixes/2026-09-08-the-gpos-subtable-walk-allocated-an-enumerator-per-glyph.md
@@ -1,0 +1,33 @@
+# The GPOS subtable walk allocated an enumerator per glyph
+
+Every lookup's `Subtables` is declared `IReadOnlyList<T>`, so a `foreach` over it boxes an enumerator
+through the interface. `GposPositioner` does that once per glyph — `ApplyMarkToBase` calls straight
+into the loop for every glyph in the run — so the cost lands on every glyph of every text run,
+whether or not the font has anything to position. Eight such loops are now indexed instead.
+
+## What measuring it turned up
+
+- **Worth 85 MB per corpus pass**, 9% of everything upstream `main` allocates over 26 real
+  documents (930.16 MB down to 844.75 MB). Additive with the anonymous-box defaulting fix: the two
+  together take it to 733.25 MB.
+- **It fires on plain Latin text**, where there is nothing to attach at all. The enumerator is
+  allocated before any coverage check, so a document with no combining marks pays it in full.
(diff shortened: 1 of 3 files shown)


jhaygood86 (approved):
Verified: all 8 loop conversions checked in full context — each is a mechanical, identical-order translation (same start/bound/step/break semantics), no off-by-one. Confirmed exhaustive: grepped every `.Subtables` access, exactly 8 sites, all converted, zero remaining `foreach` over `Subtables`; the other array-backed loops in the file were correctly left as-is. New allocation test correctly isolates the walk (non-empty, non-matching subtable, avoiding the empty-list singleton-enumerator trap) and uses the thread-local allocation counter, not the process-wide one.

Build clean, Gpos/Font/Text-filtered suite (1782 tests, including existing GPOS characterization tests) passes, CI green on all three platforms.

Approving. Merging now.

PR #969, opened 2026-09-09
Point two comments at LiveChildStartFor's real name

dchupp:
Follow-up to the minor point in your [#965 review](https://github.com/jhaygood86/PeachPDF/pull/965#issuecomment-5593726449).

#965 renamed `CssBox.LiveChildStart` to `LiveChildStartFor(slot)`, and two prose comments kept the old name:

- `CssBox.cs` ~2594 — "advances past a child via LiveChildStart once that child itself earned..."
- `FragmentEmitter.cs` ~2588 — "see CssBox.LiveChildStart's own remarks"

Neither is a `<see cref>`, so neither broke the build; both now point at a symbol that no longer exists. Comments only — no behaviour, no test change.

The one pre-existing `CS1998` in `CssBox.cs` (line 5922) is on `main` already and is untouched by this.

### DIFF
--- a/src/PeachPDF/Html/Core/Dom/CssBox.cs
+++ b/src/PeachPDF/Html/Core/Dom/CssBox.cs
@@ -2591,7 +2591,7 @@ namespace PeachPDF.Html.Core.Dom
                 box._emittedNothingAtSlot = -1;
                 box._emittedNothingScopeOwner = null;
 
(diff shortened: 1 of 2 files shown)


jhaygood86 (approved):
Verified: both locations exactly match what I flagged on #965, comment-only, no behavior change.

Approving. Merging now.

PR #974, opened 2026-09-09
Fix SVG diametric arc sweep direction

### LINKED ISSUE #973: SVG diametric arcs can render on the wrong side
# SVG diametric arcs can render on the wrong side

## Problem

SVG arcs spanning exactly 180 degrees can ignore `sweep-flag` and select the wrong semicircle. Two clockwise semicircular arcs intended to form a circle may overlap on one side.

## Reproduction
`
<svg viewBox="0 0 100 100">
  <path d="M50 10
           A40 40 0 0 1 50 90
           A40 40 0 0 1 50 10"
        fill="none"
        stroke="black"/>
</svg>
`

## Expected behavior

The two arcs form a complete circle.

## Actual behavior

Both arcs can render on the same side, producing an overlapping semicircle.

ChrisVanDijk:
Fixes #973

Summary
•	Normalize elliptical arc angles using the SVG sweep-flag.
•	Fix exact 180-degree arcs selecting the wrong semicircle.
•	Add clockwise and counterclockwise regression coverage.

Testing
•	SVG arc regression tests: 2 passed, 0 failed

### DIFF
--- a/.claude/invariants/svg-diametric-arc-direction-comes-from-sweep-flag.md
+++ b/.claude/invariants/svg-diametric-arc-direction-comes-from-sweep-flag.md
@@ -1,0 +1,9 @@
+# Diametric SVG arc direction comes from `sweep-flag`
+
+When an elliptical arc's endpoints are diametrically opposite, its angular magnitude is exactly
+180 degrees. The `large-arc-flag` cannot distinguish the two candidate arcs because neither is
+larger than the other; only `sweep-flag` selects the side.
+
+After endpoint-to-center conversion, normalize the signed sweep directly from `sweep-flag`.
+Do not infer direction from a strict less-than-180 comparison or from endpoint ordering. The
+measured failure mode is a two-arc circle whose halves both render on the same side.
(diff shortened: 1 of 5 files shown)


jhaygood86 (approved):
Welcome, and thanks for a well-targeted first PR. Reviewed against this repo's spec-compliance and rendering-correctness conventions.

Verified the spec claim directly — fetched W3C SVG 1.1 §F.6.5 myself: the Δθ correction is exactly "if fS=0 and Δθ>0, subtract 360°; if fS=1 and Δθ<0, add 360°," with no dependence on `large-arc-flag`. The old `isLargeArc == (Math.Abs(β-α) < Math.PI)` check was order-dependent and gave no answer at exactly 180° (`<` is false at `== π`); the new code implements the spec's sign-based rule directly and is more robust generally, not just at the boundary.

Then actually looked at the picture, per this repo's own convention for anything touching path/paint correctness: rendered the issue's repro through the CLI and rasterized with both PDFium and MuPDF. Pre-fix genuinely produces a single overlapping semicircle (matches the reported bug exactly); post-fix produces a complete circle, agreeing across both renderers. Also checked the existing "elliptical arc" TestHarness showcase renders identically before/after, so nothing already-shipped was silently affected.

Tests exercise the real `SvgTreeBuilder`→`SvgRenderer`→`GraphicsPathAdapter` pipeline and assert actual geometry (Bezier point extrema), not just that parsing succeeded — 100% branch coverage on the new lines, confirmed via coverage report.

Build clean, targeted suite (711 tests) passes, CI green on all three platforms.

Approving. Merging now.

PR #975, opened 2026-09-09
Measure the anonymous-box saving as a slope, not a whole-document ratio

dchupp:
Follow-up to the heads-up on [#967](https://github.com/jhaygood86/PeachPDF/pull/967#pullrequestreview) — the 2.06x you measured repeatably against the 2.05x bound. I could not reproduce it here, so rather than just widen the bound I went after why the two machines disagree at all, and the test was wrong about its own premise.

## The old form was not as self-calibrating as it claimed

It compared one list document against one plain document. That cancels what the two documents *share* — but the fixed cost of rendering anything at all (document setup, font loading, the page machinery) sat in **both** numerator and denominator. A constant added to both does not cancel in a ratio; it drags it toward 1 by however much the machine charges for it. So the number moves with the baseline, which is exactly the machine-dependent part it was supposed to be immune to. Your 2.06 against my 1.71 is that, not a hardware quirk.

## What it does now

Each shape is measured at two sizes and the ratio is taken between the **slopes**. The fixed baseline drops out of each slope; dividing the list slope by a plain block's cancels the per-item text work too. What is left is the list machinery.

| | list | plain block | ratio |
| --- | --- | --- | --- |
| with the fast path | 102.3 KB/item | 54.4 KB/item | **1.88x** |
| without | 157.3 KB/item | 54.8 KB/item | **2.87x** |

Repeatable to three decimal places (1.880, 1.879), in isolation and with the whole suite running in parallel around it.

The bound moves to **2.35x** — 25% of margin below, 18% above. The window is 53% wide where the old form's was 41%, and most of that had already gone on your machine.

**Nothing about the code under test changes.** This is the instrument only. Still mutation-checked against the pre-#967 engine: 2.85x without, 1.88x with.

Worth saying plainly: I cannot verify this fixes it on *your* machine, only that it removes the term I believe caused the divergence. If it still lands near the bound for you, that would be genuinely useful to know — it would mean the remaining per-item cost differs between us, which is a different and more interesting problem than a baseline offset.

Full suite green on net8.0 (10,420), run twice, 0 new build warnings.

### DIFF
--- a/.claude/recent-fixes/2026-09-09-an-allocation-ratio-that-still-carried-the-machines-baseline.md
+++ b/.claude/recent-fixes/2026-09-09-an-allocation-ratio-that-still-carried-the-machines-baseline.md
@@ -1,0 +1,33 @@
+# An allocation ratio that still carried the machine's baseline
+
+`AnonymousBoxDefaultingTests` compared one list document against one plain document and asserted the
+ratio. That is self-calibrating only for what the two documents share, and they shared less than the
+comment claimed: the fixed cost of rendering anything at all — document setup, font loading, the page
+machinery — sat in both numerator and denominator, so the ratio moved with whatever that costs on a
+given machine.
+
+It measured **1.71x** here against a bound of 2.05, and **2.06x** on a reviewer's machine, repeatably,
+across four runs. Same build, same direction, right on the bound — a merged test that fails for one
+of the two people who run it.
+
+It now takes each shape at two sizes and asserts the ratio of the **slopes**. The fixed baseline drops
+out of each slope, and dividing the list's by a plain block's cancels the per-item text work as well,
+leaving the list machinery itself.
+
+| | list | plain | ratio |
+| --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


jhaygood86 (approved):
Verified this actually fixes the divergence I saw, not just theoretically — ran the new slope-based test 4 times in isolation on this exact machine (previously a stable, repeatable 2.06x failure): all 4 pass now. Also ran it alongside the broader cascade/anonymous-box/list suite (649 tests) to confirm it stays stable under parallel load, not just alone.

The root-cause diagnosis is correct and the fix follows from it cleanly: a ratio of two single measurements only cancels what both share, and the fixed per-render baseline (document setup, font loading, page machinery) was in both terms, so it doesn't cancel — it drags the ratio toward 1 by however much that baseline costs on a given machine, which is exactly the machine-dependent noise the ratio was supposed to be immune to. Taking the ratio of slopes (marginal cost between two document sizes) removes that term from each side before dividing. This is a well-generalized fix, not just "make the number work for both of us."

Approving. Merging now.

PR #977, opened 2026-09-09
Cache the lookup-cache factory delegate instead of rebuilding it per call

dchupp:
`GposTable` and `GsubTable` cache every parsed lookup:

```csharp
public GposMarkToBaseLookup? GetMarkToBaseLookup(int lookupListIndex)
    => _markToBaseCache.GetOrAdd(lookupListIndex, ReadMarkToBaseLookup);
```

The second argument is a **method group**, and converting one to a `Func<>` allocates a fresh delegate on **every call** — including the overwhelmingly common one where the cache already holds the value and the factory is never invoked. The positioner and the shaper ask these questions once per glyph per lookup, so the delegate was allocated per glyph while the thing it exists to build was allocated once.

Seventeen getters now hold the delegate in a field, assigned with `??=` on first use so no constructor changes are needed.

## Measured

Over 26 real business documents:

| | allocated per pass |
| --- | --- |
| `main` | 734.6 MB |
| with this change | **677.9 MB** |

**−57 MB, 7.7% of the total.** On a text-heavy document the per-word cost falls from **12.26 KB to 9.76 KB** — about a fifth of the shaping allocation. No behaviour change; corpus fidelity is unchanged at 26/26.

## How it was found

`GC/AllocationTick` with call stacks put `System.Func<int,int>` and `System.Func<int,GposMarkToBaseLookup>` in the top ten allocated **types** of a whole document render. A `Func` ranking that high in an allocation profile of a layout engine is not a shape anyone writes on purpose, and the stacks led straight here.

## Test

`ALookupCacheHit_AllocatesNothing` primes the caches against the synthetic face the neighbouring tests already build, then asserts 6,000 subsequent hits allocate under 4 KB:

| | allocated over 6,000 cache hits |
| --- | --- |
| `main` | **384,000 bytes** — exactly 64 per call, one delegate |
| with this change | 0 |

Asserted at the table rather than through a rendered document because allocation over a document scales with whatever font a machine resolves; with the cache primed, the delegate is the only thing left that can allocate. It reads `GC.GetAllocatedBytesForCurrentThread` rather than `GC.GetTotalAllocatedBytes`, since this suite runs collections in parallel.

## Scope

Nineteen `GetOrAdd(key, MethodGroup)` sites exist in the tree. The two outside these tables — `HyphenationEngine.PatternSetCache` and `MimeTypeResolver.Cache` — are not on a per-glyph path and are deliberately left alone.

Full suite green on net8.0 (10,423 passed), 0 new build warnings.

### DIFF
--- a/.claude/recent-fixes/2026-09-09-a-method-group-passed-to-getoradd-allocates-per-call.md
+++ b/.claude/recent-fixes/2026-09-09-a-method-group-passed-to-getoradd-allocates-per-call.md
@@ -1,0 +1,36 @@
+# A method group passed to GetOrAdd allocates per call
+
+`GposTable` and `GsubTable` cache every parsed lookup —
+`_cache.GetOrAdd(lookupListIndex, ReadSomething)`. The second argument is a **method group**, and
+converting one to a `Func<>` allocates a fresh delegate on *every call*, including the
+overwhelmingly common one where the cache already holds the value and the factory is never
+invoked. The positioner and the shaper ask these questions once per glyph per lookup, so the
+delegate was allocated per glyph while the thing it existed to build was allocated once.
+
+Seventeen getters across the two tables now hold the delegate in a field (`??=` on first use, so no
+constructor changes).
+
+## What measuring it turned up
+
+- **Worth 57 MB per corpus pass, 7.7% of the total** — 734.6 MB down to 677.9 MB over 26 real
+  documents. Per-word allocation on a text-heavy document drops from **12.26 KB to 9.76 KB**, a
+  fifth of the shaping cost.
+- **It is exactly 64 bytes a call** — 6,000 cache hits allocated 384,000 bytes before, and zero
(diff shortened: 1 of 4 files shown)


jhaygood86 (approved):
Verified all 17 sites directly: each cached field's name matches its cache and its wrapped method exactly, no cross-wiring. Confirmed the "19 method-group GetOrAdd sites, 17 fixed + 2 deliberately left alone" scope claim myself via grep — matches exactly (`MimeTypeResolver.Cache`/`HyphenationEngine.PatternSetCache` are the two, as stated). Also confirms why this needed fixing at all rather than relying on compiler caching: these are all *instance* method groups, which always capture `this` on delegate conversion, so the compiler's automatic static-delegate-caching (which does apply to non-capturing lambdas/static method groups) never kicks in here — a fresh delegate really was allocated every call.

Ran the new allocation test 3 times in isolation (stable, 0 bytes each run) and the broader Gpos/Gsub/Font suite (875 tests, all pass).

Issue #29, opened 2025-06-12
page-break-inside: avoid does not seem to be working


DanteDeRuwe:
Hi,

page-break-inside: avoid does not seem to be working, I still have page breaks inside divs where I put this property.

I've seen in the code this property is recognized, but cannot seem to render properly.

Thanks in advance

PR #979, opened 2026-09-09
Implement overflow-wrap: break-word and anywhere

### LINKED ISSUE #696: overflow-wrap / word-wrap parsed but never applied (Tailwind's break-words)
## What's unsupported

`overflow-wrap` and its legacy alias `word-wrap` are recognized CSS-OM property names (`PropertyNames.OverflowWrap`/`PropertyNames.WordWrap`) — they parse and store on the style declaration — but neither has an entry in `css-properties.json`, so neither is wired to a `CssBox` and neither affects line breaking. An overlong unbreakable token (a long URL, a long unhyphenated word) always overflows its container rather than breaking mid-word when `overflow-wrap: break-word`/`anywhere` (or `word-wrap: break-word`) is set.

## Spec

[CSS Text Module Level 3 — the `overflow-wrap` property](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap).

## Impact

Tailwind's `break-words` / `wrap-break-word` utility (and any hand-written `overflow-wrap`/`word-wrap` CSS) has no effect — long unbroken tokens keep overflowing their box instead of wrapping. Documented as unsupported in [docs/html-css-support.md](../blob/main/docs/html-css-support.md).

ChrisVanDijk:
Fixes #696

`overflow-wrap` and its legacy `word-wrap` alias were accepted by the CSS parser but had no effect,
so an unbreakable token overflowed its box even when `break-word` was asked for, and `anywhere` was
rejected outright. Both names now support `normal`, `break-word`, and `anywhere`, and both inherit
(CSS Text 3 §5.4).

## Behaviour

Emergency wrapping is genuinely last-resort, not a second `word-break: break-all`:

1. An ordinary soft wrap opportunity before the token wins, so the whole word moves to a fresh line.
2. Hyphenation is tried next, under its existing `hyphenate-limit-*` gates.
3. Only if the token still does not fit is it split, at the last extended grapheme cluster boundary
   that fits. No hyphen is inserted.

The split is grapheme-safe, so a combining sequence, surrogate pair, or emoji sequence is never cut
in half, and the fragments carry the original run's bidi level, script tag, Arabic joining forms, USE
categories, per-codepoint font fallback, first-line style, and hyphenation candidates. It works in
vertical writing modes along the inline axis, and has no effect where `white-space` forbids wrapping.

Per spec, `anywhere` contributes its emergency opportunities to min-content sizing (so it shrinks a
grid track, flex item, shrink-to-fit box, or auto table column) while `break-word` deliberately does
not.

## Notes for review

**Split rollback.** A split is chosen against one line's available measure, so it must not survive
into a different one. The prefix/suffix pair links back to the original word, and two paths undo it:
a fragmentainer line discarded at a page break restores the word before the resumed pass re-places
it, and a fresh full reflow restores every prior split before choosing against the new measure. This
mirrors the existing hyphenation rollback. `RestoreOverflowWrapSplits` walks backward on purpose — a
suffix can itself have been split — and `TryRestoreOverflowWrapSplit` asserts the pair is adjacent
before merging, both covered by tests.

**Lazy intrinsic measurement.** `anywhere`'s min-content contribution is the widest indivisible
grapheme, which needs a graphics context that `GetMinMaxWidth` did not have. Rather than measure
every grapheme of every word during ordinary word measurement — which made an inherited
`body { overflow-wrap: anywhere }` about 3x slower to lay out even when nothing overflowed — the
graphics context is threaded into `GetMinMaxWidth`/`GetMinMaxSumWords` and the value is measured and
memoized on first intrinsic query, invalidated whenever word measurement re-runs.

**Table minimum widths.** `GetColumnMinWidths` already used `GetMinMaxWidth`, but the auto-layout
minimum pass still used `GetMinimumWidth()`, a separate and differently-behaved definition of
min-content (longest word plus one ancestor padding chain, no `<br>` handling, no `white-space`
clamp). Both now use `GetMinMaxWidth`, which also means a `white-space: nowrap` cell picks up the
min-content-is-max-content clamp there, as that clamp's own comment intends.

**Shared `white-space` predicate.** "Does `white-space` leave any soft wrap opportunity" had been
written out independently in the line breaker and the intrinsic walk, which let them disagree: a
nested `nowrap` run contributed one grapheme to its parent's min-content width while line layout
correctly refused to split it. Both, plus `GetMinMaxWidth`'s own pre-existing clamp, now read
`CssBox.WhiteSpacePermitsWrapping`.

**CSSOM alias.** A `word-wrap` declaration is constructed as an `OverflowWrapProperty`, whose name is
necessarily `overflow-wrap`, so reads, priority changes, and removals through the legacy spelling
addressed a slot nothing stored under. They now canonicalize to the same declaration.

## Docs and showcase

`docs/html-css-support.md` gains an `overflow-wrap` / `word-wrap` row and drops the stale "not
supported" bullet. A new `overflow_wrap` showcase covers all three values, the alias, and the
min-content distinction between `anywhere` and `break-word`.

## Testing

- Whole-solution rebuild: 0 warnings, 0 errors.
- Full net8.0 suite: 10,451 passed, 9 skipped.
- Changed production lines: 100% covered (278/278).
- Integration coverage for each value and the alias, inheritance, the ordinary-opportunity-first
  priority, `nowrap`/`pre` suppression, combining sequences, vertical writing mode, repeated reflow
  stability, min-content sizing in grid and auto tables, and split rollback.

Unrelated to this change: `FontSynthesisIntegrationTests.ObliqueWithExplicitAngle_…` fails on any
machine whose locale uses a comma decimal separator, because it formats its expected skew with
`.ToString("0.####")` and no `InvariantCulture` while the PDF is written invariant. It fails on `main`
too and is untouched here.


jhaygood86:
Reviewed against this repo's spec-compliance, rendering-correctness, and blast-radius conventions — this is a large, foundational change (`GetMinMaxWidth`, table auto-layout, bidi/shaping metadata, pagination rollback) so it got the highest scrutiny pass.

**Verified against the actual spec.** Fetched CSS Text 3 §5.4 directly: `anywhere`'s and `break-word`'s only behavioral difference is min-content participation, grapheme clusters must stay intact at the break, and no hyphen is inserted — all matches this implementation exactly, including inheritance (`Yes` in the spec's property table, `inherited: true` here).

**Rollback correctness (the highest-risk part).** Traced `RestoreOverflowWrapSplits`/`UndoAbandonedOverflowWrapSplits`/`TryRestoreOverflowWrapSplit` in full. Both invalidation paths are wired exactly where they need to be: `RestoreOverflowWrapSplits(blockBox)` runs before `FlowBox` makes any new split decision on a fresh (non-resumed) pass, and `UndoAbandonedOverflowWrapSplits` runs on a discarded fragmentainer line right alongside the existing `UndoAbandonedHyphenationSplits`, before the surviving words are marked `AwaitsTheNextFragmentainer`. The adjacency check in `TryRestoreOverflowWrapSplit` (index lookup + `ReferenceEquals` on the next slot) correctly refuses to merge a non-adjacent or already-modified pair. This mirrors the hyphenation rollback machinery closely enough that I'm confident in it.

**Grapheme-cluster safety, verified empirically, not just read.** `StringInfo.ParseCombiningCharacters` on net8.0 (ICU-backed) genuinely implements UAX #29 extended grapheme cluster boundaries — I ran it directly against a ZWJ family emoji, a regional-indicator flag pair, a skin-tone modifier sequence, and a surrogate-pair emoji, and all four came back as one indivisible cluster. So the "never split mid-cluster" claim holds for real. One suggestion: `OverflowWrapIntegrationTests.CombiningSequence_RemainsOneGraphemeAtEveryEmergencyBreak` only exercises `á` (a base + combining mark) — given the PR description explicitly calls out "a surrogate pair, or emoji sequence," it'd be worth a second case (a ZWJ sequence or flag pair) so that specific claim has direct test coverage rather than resting on `StringInfo`'s own correctness.

**Blast radius of the two unifications** (the other thing I was worried about, since both are described as fixing a prior *disagreement* between two call sites): `WhiteSpacePermitsWrapping` turned out narrower than it reads — it's used at exactly 3 sites, two of which are brand-new `overflow-wrap`-gated code, and the third (`GetMinMaxWidth`'s clamp) is a byte-for-byte-equivalent refactor of the existing `WhiteSpace.Value is NoWrap or Pre` check. Ordinary documents not using the property go through unchanged logic there. The table min-width change (`GetColumnMinWidths` switching from `GetMinimumWidth()` to `GetMinMaxWidth`) is a real behavior change for every auto-layout table, not just `overflow-wrap` users — but it's a one-method change in the direction of more spec-correct sizing (adds `<br>` handling and the `nowrap` clamp that were missing), and the full 726-test `Table` suite passes unchanged. I didn't find a dedicated before/after comparison against a real browser for an ordinary (non-`overflow-wrap`) table with `<br>` in a cell under auto layout — might be worth a quick manual check, though I wouldn't block on it given the existing suite's coverage.

**Small doc nit:** `.claude/recent-fixes/2026-09-09-overflow-wrap-breaks-overlong-words.md`'s "Table minimum widths" paragraph (and the matching PR description paragraph) has the two method names swapped — it says `GetColumnMinWidths` already used `GetMinMaxWidth` and the auto-layout pass used `GetMinimumWidth()`, but reading the diff, `GetColumnMinWidths` *is* the auto-layout pass that changed from `GetMinimumWidth()` to `GetMinMaxWidth`; `GetColumnsMinMaxWidthByContent` is the one that already used `GetMinMaxWidth`. Worth a one-line fix so the note reads correctly for the next person who greps it.

Otherwise this is exactly as advertised: whole-solution rebuild is 0 warnings/0 errors, the targeted suite (35 tests) and a broad regression pass across Table/Hyphenation/Bidi/Grid/Flex (1,777 tests total) all pass, and I rendered the new `overflow_wrap` showcase and rasterized it with both PDFium and MuPDF — they agree, the emergency break is clean (not mid-glyph), and the `anywhere`/`break-word` min-content distinction is visibly real (the `anywhere` chip wraps down to one grapheme per line, `break-word`'s stays the full word wide).

Nice work — welcome back.

jhaygood86 (approved):
Thoroughly reviewed given the scope of this one — thanks for a genuinely well-built feature. A few things I checked myself rather than took on trust:

Confirmed the spec claims against CSS Text 3 §5.4 directly: `anywhere`/`break-word` differ only in min-content participation, grapheme clusters must stay intact at the break, no hyphen inserted, inherits — all match.

Traced the rollback logic (`RestoreOverflowWrapSplits`/`UndoAbandonedOverflowWrapSplits`/`TryRestoreOverflowWrapSplit`) end to end since this was the highest-risk part — both invalidation paths (fresh reflow, discarded fragmentainer line at a page break) run before any new split decision, mirroring the proven hyphenation-rollback pattern exactly.

Wrote a standalone probe against `StringInfo.ParseCombiningCharacters` to verify grapheme-safety empirically rather than just reading the claim: confirmed a ZWJ family emoji, a regional-indicator flag pair, a skin-tone modifier sequence, and a surrogate-pair emoji each stay one indivisible cluster.

On the two shared-code unifications: `WhiteSpacePermitsWrapping` only reaches new, `overflow-wrap`-gated code plus one behavior-identical refactor, so ordinary documents are unaffected there. The table min-width unification (`GetMinimumWidth()` → `GetMinMaxWidth` in `GetColumnMinWidths`) *is* a real behavior change for every auto-layout table, not just this feature's users — but it's a move toward more spec-correct sizing (adds `<br>` handling and the `nowrap` clamp), and the full table suite (726 tests) passes unchanged.

Actually rendered the new showcase and rasterized with both PDFium and MuPDF — they agree, the min-content distinction between `anywhere` and `break-word` is visibly real, and nothing breaks mid-glyph.

PR #981, opened 2026-09-09
Look a rule's selector and declarations up by index, not with LINQ

dchupp:
`StyleRule.Selector` and `StyleRule.Style` were both:

```csharp
get => Children.OfType<T>().FirstOrDefault();
```

`StylesheetNode.Children` hands the child list back as `IEnumerable<IStylesheetNode>`, so each read builds an `OfType` iterator and boxes a list enumerator — **two allocations, every time a property is read**. A property getter is not somewhere anyone looks for that, which is presumably how it survived; it came fourth in the ranked allocating call sites of a whole document render, above the fragment emitter.

A new `protected IReadOnlyList<IStylesheetNode> ChildList` lets both getters walk by index. Indexing an `IReadOnlyList` allocates nothing.

## Measured

Over 26 real business documents, against `main` **as it stands after #977 and #978**:

| | allocated per pass |
| --- | --- |
| `main` | 619.3 MB |
| with this change | **579.3 MB** |

**−40 MB, 6.5% of the total.** Worth noting explicitly: it measured 39 MB *before* those two landed, so #978's rewrite of selector matching did not absorb it — the getter is read from more places than that one path. Corpus fidelity unchanged at 26/26.

## Test

`StyleRuleLookupAllocationTests` asserts 4,000 reads allocate under 4 KB:

| | allocated over 4,000 property reads |
| --- | --- |
| `main` | **352,000 bytes** — exactly 88 a read, two allocations |
| with this change | 0 |

A second test pins that the indexed walk finds the same children the LINQ form did, **including after `SelectorText` has replaced the selector** — the one case where the child is not the one the constructor added.

Per-thread rather than `GC.GetTotalAllocatedBytes`, since the suite runs collections in parallel.

## Why a walk and not a cached field

The setter is `ReplaceSingle(Selector, value)` — it reads the getter — so the write path is unchanged and there is no cached state that could go stale. A memoised field would need invalidating from `AppendChild`/`ReplaceSingle`/`RemoveChild` on the base class, which is a lot more surface for 88 bytes a read.

Full suite green on net8.0 (10,455 passed), 0 new build warnings.

### DIFF
--- a/.claude/recent-fixes/2026-09-09-a-property-getter-that-ran-a-linq-query-per-read.md
+++ b/.claude/recent-fixes/2026-09-09-a-property-getter-that-ran-a-linq-query-per-read.md
@@ -1,0 +1,33 @@
+# A property getter that ran a LINQ query per read
+
+`StyleRule.Selector` and `StyleRule.Style` were both
+`Children.OfType<T>().FirstOrDefault()`. `StylesheetNode.Children` hands the child list back as
+`IEnumerable<IStylesheetNode>`, so each read built an `OfType` iterator and boxed a list
+enumerator — two allocations, every time a property was read.
+
+Selector matching reads `Selector` once per box per candidate rule, so a document with many rules
+and many boxes pays it per pair. `ChildList`, a new `protected IReadOnlyList<IStylesheetNode>` on
+the node, lets both getters walk the children by index instead.
+
+## What measuring it turned up
+
+- **40 MB per corpus pass, 6.5% of the total** — 619.3 MB down to 579.3 MB over 26 real documents,
+  measured against `main` *after* #977 and #978 landed. It was 39 MB before those, so #978's rewrite
+  of selector matching did not absorb it: the getter is read from more places than that path.
+- **88 bytes per read, exactly**: 4,000 property reads allocated 352,000 bytes before and zero
+  after. Two allocations at 44 bytes apiece, which is what makes this testable to the byte.
+- **A property getter in an allocation profile is the tell.** `StyleRule.get_Selector` came fourth
+  in the ranked call sites of a whole document render, above the fragment emitter. Nobody expects a
+  property read to be a top-five allocator, which is exactly why it survived.
+- **The setter still reads the getter** (`ReplaceSingle(Selector, value)`), so the write path is
+  unchanged and there is no cached state to invalidate — the reason this is an indexed walk rather
+  than a memoised field.
+
+## Evidence
+
+`StyleRuleLookupAllocationTests` asserts 4,000 reads allocate under 4 KB, and separately that the
+indexed walk still finds the same children the LINQ form did — including after `SelectorText` has
+replaced the selector, which is the one case where the child is not the one the constructor added.
+Per-thread, because the suite runs collections in parallel.
+
+Full suite green on net8.0 (10,424), 0 new build warnings, corpus fidelity unchanged at 26/26.
--- a/src/PeachPDF.Tests/CSS/StyleRuleLookupAllocationTests.cs
+++ b/src/PeachPDF.Tests/CSS/StyleRuleLookupAllocationTests.cs
@@ -1,0 +1,111 @@
+using System;
+using System.Linq;
+using PeachPDF.CSS;
+using Xunit;
+
+namespace PeachPDF.Tests.CSS
+{
+    /// <summary>
+    /// Reading a rule's selector or declaration block allocates nothing.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Both were <c>Children.OfType&lt;T&gt;().FirstOrDefault()</c>. <c>Children</c> hands the child
+    /// list back as <see cref="System.Collections.Generic.IEnumerable{T}"/>, so that query allocates
+    /// an <c>OfType</c> iterator and boxes a list enumerator — on <b>every read</b>, of a property.
+    /// </para>
+    /// <para>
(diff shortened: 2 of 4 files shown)


jhaygood86 (approved):
Verified independently with the `--benchmark` mode (a real, reproducible advantage of having that tool in the tree now) against the full 111-showcase corpus rather than your private 26-document one — different corpus, same conclusion: 2 runs each side, Release, net10.0, 10 iterations:

| | allocation (mean of 2 runs) |
| --- | --- |
| `main` | ~2051.3 MB |
| this change | ~1985.4 MB |

**−65.9 MB, −3.2%**, consistent within <0.2% between runs on each side (wall time is noisier, as expected, and trended lower too but I'd weight allocation here). Different percentage than your 6.5% since it's a different, more selector-light corpus, but same direction and same order of magnitude — good independent confirmation.

Code itself is a clean, minimal translation — both getters are semantically identical indexed scans (first `ChildList[i]` assignable to `T`, same as `OfType<T>().FirstOrDefault()`), and `TheLookupsStillFindTheRightChildren` correctly pins the one case where the found child isn't the constructor's original (after `SelectorText` replaces the selector). Agreed that a walk over a cached field isn't worth it here — the setter already reads the getter, so caching would need invalidating from three call sites for 88 bytes.

One tiny nit, not blocking: `Style`'s `/// <inheritdoc cref="Selector"/>` has nothing to inherit from, since `Selector`'s explanation is a `//` line comment, not an XML doc comment — no build warning either way (confirmed), and the reasoning is obvious from the adjacent code, so feel free to leave it or swap it for a real one-liner whenever convenient.

Build clean, StyleRule/Selector-filtered suite (328 tests) passes.

jhaygood86:
CI failed on all three platforms — not a test failure (the full suite genuinely passes: 10,455/10,464 on every platform, matching what the PR description claims), but the diff-coverage gate:

```
src/PeachPDF/CSS/Rules/StyleRule.cs (83.3%): Missing lines 68
Total:   7 lines
Missing: 1 line
Coverage: 85%
```

Line 68 is the `return null;` fallback in `Style`'s getter — the "no `StyleDeclaration` child found" branch. Neither of the two new tests constructs a `StyleRule` without a declaration block (both go through `StylesheetParser().Parse(...)` with `color: red` present), so that branch is never exercised. `Selector`'s equivalent fallback is covered (presumably by an existing test elsewhere that parses a rule with no selector, or a malformed one), but `Style`'s isn't.

Could you add a case — parsing something that produces a `StyleRule` with no declaration block, or constructing one directly, and asserting `Style` is `null` — the same way `TheLookupsStillFindTheRightChildren` already checks the found-child case?

Everything else here still looks good — I ran this independently against the full showcase-corpus benchmark from #978 (different corpus than your 26 documents) and got a consistent, real allocation reduction (−3.2%) confirming the direction and rough magnitude of your own numbers.

jhaygood86 (changes requested):
Correcting my earlier approval — CI failed on all three platforms after I approved, on the diff-coverage gate (see the comment above: 85% overall, StyleRule.cs's Style getter's null-fallback branch at line 68 is uncovered). Requesting changes until that's addressed; will re-review once updated.

dchupp:
Fixed — force-pushed. Diff coverage is now 100% (7 of 7 coverable added lines), reproduced locally with the same instrument CI uses (`diff-cover` over `coverage.cobertura.xml` against the merge-base with `main`).

You were right that neither existing test could reach line 68: both parse real CSS, and the `StyleRule` constructor unconditionally appends both an `AllSelector` and a `StyleDeclaration`, so a parsed rule always has both children. The only way to a missing child is removing one.

New test `EachLookupIsNullOnceItsOwnChildIsGone` constructs a rule directly and removes each child in turn:

```csharp
var rule = new StyleRule(new StylesheetParser());

Assert.NotNull(rule.Selector);
Assert.NotNull(rule.Style);

rule.RemoveChild(rule.Style);

Assert.Null(rule.Style);
Assert.NotNull(rule.Selector);

rule.RemoveChild(rule.Selector);

Assert.Null(rule.Selector);
Assert.Equal(string.Empty, rule.SelectorText);
```

It covers a bit more than just the branch you flagged. Removing one child and asserting the *other* is still found pins the thing an indexed walk could plausibly get wrong that the LINQ form could not: `OfType<T>()` can only ever yield a `T`, whereas a hand-written loop that matched on the wrong type would happily return the sibling. Both directions are now asserted, and the last two lines pin `SelectorText`'s documented null-safety on the same object rather than by inspection.

Also took the doc nit — `Style`'s `<inheritdoc cref="Selector"/>` is gone, replaced by a real two-line comment saying why it is indexed and what `null` means, matching the style of `Selector`'s own comment.

Full suite green on this branch: **10,456 passed, 0 failed, 9 skipped** (net8.0, Release). Build clean apart from the pre-existing `CssBox.cs(5922)` CS1998.

And thank you for the independent −3.2% on the 111-showcase corpus — a second corpus agreeing on direction and order of magnitude is worth more than another decimal place on mine.

jhaygood86 (approved):
Verified: diff coverage now passes on CI (confirmed green on all three platforms, including the gate that failed last time). `EachLookupIsNullOnceItsOwnChildIsGone` covers more than the missing branch alone — removing one child and asserting the *other* is still found catches a failure mode the LINQ form structurally couldn't have (a hand-written loop matching the wrong type), which `OfType<T>()` guaranteed against for free. Good instinct testing both directions rather than just closing the coverage gap mechanically.

Build clean, StyleRule/Selector-filtered suite (329 tests) passes.

Approving. Merging now.

PR #984, opened 2026-09-09
Apply ccmp/locl to all text, and never draw a default-ignorable

### LINKED ISSUE #983: Emoji sequences don't compose, and variation selectors render as missing-glyph boxes
Rendering emoji with a real COLR v1 build of [Noto Color Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji) produces three distinct wrong results. They share a root area (GSUB feature selection and Unicode default-ignorable handling), so filing them together.

### Repro

```html
<style>
@font-face { font-family: 'NotoCOLR'; src: url('NotoColorEmoji-Regular.ttf') format('truetype'); }
body { font-family: 'NotoCOLR'; font-size: 48px; }
</style>
<p>&#10084;&#65039;</p>                                  <!-- ❤️  heart + VS16 -->
<p>&#127987;&#65039;&#8205;&#127752;</p>                 <!-- 🏳️‍🌈 ZWJ sequence -->
<p>&#127482;&#127480; &#127471;&#127477;</p>             <!-- 🇺🇸 🇯🇵 regional indicators -->
<p>&#127988;&#917607;&#917602;&#917619;&#917603;&#917620;&#917631;</p>  <!-- 🏴󠁧󠁢󠁳󠁣󠁴󠁿 tag sequence -->
```

### What happens

| Input | Expected | Actual |
|---|---|---|
| `U+2764 U+FE0F` | ❤️ | heart **followed by a tofu box** |
| `U+1F3F3 U+FE0F U+200D U+1F308` | 🏳️‍🌈 | white flag **plus a separate rainbow** |
| `U+1F1FA U+1F1F8` | 🇺🇸 | the letters **"US"** |
| `U+1F3F4 U+E0067…U+E007F` | 🏴󠁧󠁢󠁳󠁣󠁴󠁿 | a **bare black flag** |

Confirmed identical in both PDFium and MuPDF, so it is the produced PDF, not a viewer quirk.

### Causes

**1. `Default_Ignorable_Code_Point` falls through to `.notdef`.** The font has no cmap entry for `U+FE0F` — which is correct, since a variation selector is default-ignorable and no font is expected to draw one. It resolves to glyph 0 and gets painted. Same applies to ZWJ, the bidi controls, and the tag characters.

**2. A hidden ignorable breaks ligature component matching.** The font's rainbow-flag rule reads `flag + ZWJ + rainbow` and never mentions `U+FE0F`, so the unmapped selector sits between components and breaks the run. Real shapers step over it (HarfBuzz's `SKIP_MAYBE`).

**3. `ccmp`/`locl` are never applied to ordinary text.** Both are default-on in the [OpenType feature registry](https://learn.microsoft.com/en-us/typography/opentype/spec/featuretags), but they're currently reachable only from the Arabic joining pre-stage and the USE pre-stage. **Noto Color Emoji's entire GSUB `FeatureList` is a single `ccmp` record** — it declares no `liga`/`rlig`/`clig` at all — so every sequence ligature is unreachable.

### Why this wasn't noticed sooner

ZWJ (`U+200D`) has `Joining_Type=Join_Causing`, so an emoji ZWJ sequence produces a non-empty joining-forms array and trips the **Arabic** pre-stage, picking up `ccmp` as a side effect. That makes 🏳️‍🌈 *nearly* work (bug 2 aside) while every non-ZWJ sequence fails, which reads as "emoji are supported."

### Not the fix

cmap **format-14** variation sequences are also unapplied, but implementing that would not fix any of the above: all 371 `U+FE0F` records in the real font are *default* UVS records (glyph `None` = "use the base cmap glyph"), and `U+1F600 U+FE0F` is absent from its UVS table entirely. Sequence composition comes from GSUB, not the cmap.

ChrisVanDijk:
Fixes #983

Three defects in text shaping, found by rendering a real COLR v1 Noto Color Emoji face and looking at the output rather than at the code. They compose — only the third is what most emoji actually needed, and it was invisible behind the first two.

## 1. A default-ignorable no longer draws `.notdef`

New `Text/UnicodeDefaultIgnorables.cs` covers the 17 UCD ranges of Unicode's `Default_Ignorable_Code_Point` property — an inline comparison chain, far too small to justify the Brotli generated-table pattern `VerticalOrientationTable`/`BidiClassTable` use.

`GsubShaper.MapToGlyphs` flags such a glyph as `ShapedGlyph.IsHiddenIgnorable` at map time — the one place the codepoint and its resolved glyph are both in hand — and `OpenTypeDescriptor.DropHiddenIgnorables` deletes them at the very end of `Shape`.

**The ordering is load-bearing: flag early, delete last.** An ignorable is load-bearing *during* shaping (ZWJ is exactly what makes an emoji ZWJ sequence ligate; a bidi control can be the context a contextual rule matches), so it has to reach every lookup and only then be hidden — HarfBuzz's `hide_default_ignorables` sequences it the same way. It also can't be decided *later*, because a ligature merge rewrites cluster spans, so re-deriving "was this `.notdef` standing in for one ignorable codepoint?" afterwards is guesswork.

Only glyph index 0 is ever flagged, so a font that ships a real blank glyph for an ignorable is honored as authored — which is also what keeps a soft hyphen (`U+00AD`, itself default-ignorable, drawn as a visible hyphen by most fonts under `hyphens: none`) behaving exactly as before. Deletion remaps `AttachedToIndex`, which is a *glyph-list* index; `LigatureComponentClusterStarts` holds text offsets and needs no fixup.

## 2. A hidden ignorable no longer blocks a ligature

`TryMatchLigature` steps over an `IsHiddenIgnorable` glyph instead of matching against it (HarfBuzz's `SKIP_MAYBE`), carrying it in the existing `skipped` list so it is re-inserted after the merge and then deleted with the rest. Safe by construction: a component glyph id is a real glyph, never `.notdef`, so this can only turn a previous *failure* into a match.

## 3. `ccmp`/`locl` are applied to all text

They were reachable only from the Arabic and USE pre-stages, so anything else got none. Both are default-on in the [OpenType feature registry](https://learn.microsoft.com/en-us/typography/opentype/spec/featuretags), not stylistic opt-ins. `GetActiveLookupIndices` now adds both tags, **gated on neither pre-stage having run**. That gate is not tidiness — a second application is not idempotent, since a Type 2 (Multiple Substitution) decomposition will happily decompose its own output again.

This is what actually broke emoji. **Noto Color Emoji's entire GSUB `FeatureList` is a single `ccmp` record** — it declares no `liga`/`rlig`/`clig` at all — so every sequence ligature was unreachable:

| Input | Before | After |
|---|---|---|
| `U+2764 U+FE0F` | heart + **tofu box** | ❤️ |
| `U+1F3F3 U+FE0F U+200D U+1F308` | white flag + separate rainbow | 🏳️‍🌈 |
| `U+1F1FA U+1F1F8` | the letters **"US"** | 🇺🇸 |
| `U+1F3F4 U+E0067…U+E007F` | bare black flag | 🏴󠁧󠁢󠁳󠁣󠁴󠁿 |
| `U+1F44D U+1F3FD` | thumbs-up + **brown square** | 👍🏽 |

**Skin tone modifiers are the case most likely to be hit in an ordinary document, and they degrade differently** — worth knowing, because the difference makes them easy to misdiagnose. `U+1F3FB`–`U+1F3FF` are *not* default-ignorable: each has a real colour-swatch glyph, so an unapplied `ccmp` doesn't hide the modifier, it **draws** it. Hence a visible square rather than a missing-glyph box. Fixed by (3) alone; (1) and (2) contribute nothing there.

## The trap worth recording

**One case worked, and it worked by accident.** ZWJ is `Joining_Type=Join_Causing`, so an emoji ZWJ sequence produced a non-empty joining-forms array, tripped the **Arabic** pre-stage, and picked up `ccmp` as a side effect. That single working case is why the gap survived — it made emoji look supported.

It also cost real debugging time: reasoning from the code said the rainbow-flag ligature could not form, while the rendered PDF plainly showed that it did. An instrumented run (`join=4` on a six-UTF-16-unit emoji run) is what settled it. If you're ever checking whether a GSUB feature reaches a run, use a sequence with **no** ZWJ in it.

## Verification

- Every sequence family correct in **both PDFium and MuPDF** — ZWJ, regional-indicator, tag, and skin tone.
- Arabic, Devanagari and Latin render **pixel-identical** before/after (`ImageChops.difference` bbox `None`). This is what proves the gate holds and bounds the blast radius of (3) to text that previously got no `ccmp` at all.
- Suite: **10526 passing**. The one failure (`FontSynthesisIntegrationTests.ObliqueWithExplicitAngle_…`) was verified to fail identically on a clean `main` and is unrelated to this change.
- The two new suites together fail **38 of 47** cases against the pre-fix sources, so they are not no-op guards.
- **100% diff coverage** on all four changed source files; `dotnet build PeachPDF.slnx -t:Rebuild` is clean at **0 warnings**.

## Tests and assets

- `DefaultIgnorableShapingTests` — drives `CcmpLigatureTest.ttf` (hand-authored, CC0): its only GSUB feature is `ccmp` and it has no `U+FE0F` glyph, so a regression in (1) or (3) fails immediately. Includes a representative codepoint from every ignorable range plus its non-ignorable neighbour, so a mistyped range boundary is a concrete failure.
- `EmojiSequenceCompositionTests` — drives the **real** font via `NotoColorEmojiSequences-Subset.ttf` (OFL): ZWJ sequences, regional-indicator pairs, a tag sequence, all five skin tones (each asserted to compose to a *distinct* glyph, so a collapsed or dropped modifier fails), and a lone variation selector.

The existing `NotoColorEmoji-Subset.ttf` could not be used for either: it has **no GSUB features at all**. Both new fonts have generator scripts and licence files beside them; `docs/license.md` is updated.

## Deliberately out of scope

- **cmap format-14 variation sequences.** Considered first and measured: it would have fixed none of this. All 371 `U+FE0F` records in the real font are *default* UVS records (glyph `None` = "use the base cmap glyph"), and `U+1F600 U+FE0F` is absent from its UVS table entirely, so the tofu would have survived. Sequence composition comes from GSUB, not the cmap. Remains a recorded gap, now annotated with that measurement.
- **`SKIP_MAYBE` for contextual (GSUB 5/6) and GPOS matching.** Only ligature component matching needed it for these fonts, and the contextual matchers reach into Arabic/Indic paths that have needed repeated repair — a separate change with its own verification burden.

## Note for review

`ArabicJoiningCharacterizationTests.NoJoiningFormsRequested_SkipsTheCcmpPreStage` asserted the pre-fix behaviour (`ccmp` deliberately *not* applied without joining forms) and was pinning the bug in place. It is replaced by `NoJoiningFormsRequested_StillAppliesCcmp_ButOnlyOnce`, whose same-glyph-count-either-way assertion is what actually proves the new gate holds.


jhaygood86:
Reviewed this closely given the surface area (GSUB glyph deletion + index remapping, ligature skip/reinsert, the ccmp/locl gate) — really solid work, on par with #974/#979. The core logic all checks out:

- **`DropHiddenIgnorables`**: the `oldIndex -> newIndex` remap with `-1` for deleted glyphs is correct, including the "attached to a deleted glyph → `null`" case, the unsigned-cast bounds check, and the all-glyphs-kept / early-return fast path. Traced the boundary cases (first/last glyph deleted) by hand; no off-by-one.
- **`TryMatchLigature`'s skip-and-reinsert**: `skippedOffsets` are captured against the original (pre-`RemoveRange`) list before the mutation, so the re-insert at `index + 1` is safe and lands every skipped ignorable immediately after the merged glyph in original relative order. `componentClusterStarts` is built only from `matched` (never `skipped`), so a hidden ignorable never contaminates the ligature's own cluster bookkeeping.
- **The `ccmp`/`locl` gate**: confirmed it's a real NOR (`JoiningForms` empty AND `UseCategories` empty), not an OR-with-inverted-sense mistake. Reverted just `GsubShaper.cs` + `OpenTypeDescriptor.cs` to `main` and reran `DefaultIgnorableShapingTests`: 23 of 32 cases fail (see note below on the count), confirming the gate/guard isn't a no-op.
- Rendered the five sequences from `NotoColorEmojiSequences-Subset.ttf` (ZWJ+VS16 rainbow flag, ZWJ woman-technologist, both regional-indicator pairs, the Scotland tag sequence, and lone heart+VS16) through the actual CLI and rasterized with both PDFium and MuPDF — all five compose correctly in both, no tofu. Also rendered an Arabic joining sample against `main` vs this branch and diffed the rasterizations: `ImageChops.difference` bbox is `None`, confirming pixel-identity holds.

Two small things worth fixing before merge, both easy:

1. **Stale method reference in a comment.** `CssBox.cs`'s `NeedsPerCodepointFont`:
   ```
   // Shaping hides it instead (OpenTypeDescriptor.DropUnrenderableDefaultIgnorables).
   ```
   That method doesn't exist — it's `OpenTypeDescriptor.DropHiddenIgnorables`. Looks like a leftover from an earlier name for the method.

2. **The mutation-testing count is off.** I reverted `GsubShaper.cs`/`OpenTypeDescriptor.cs` to `main` and reran `DefaultIgnorableShapingTests` directly: it's 32 cases total (not 30, per the recent-fixes note) and **23** fail (not 21, per both the PR description and the recent-fixes note). Doesn't change the conclusion — it's still clearly not a no-op guard — but since `.claude/recent-fixes/2026-09-10-ccmp-is-default-on-and-default-ignorables-are-never-drawn.md` is meant to be durable engineering history, worth correcting the numbers there and in the invariants note before it lands.

Minor, non-blocking suggestion: `DropHiddenIgnorables`'s `kept.Count == 0` path (an entirely-default-ignorable input, e.g. shaping a lone ZWJ with nothing else) isn't exercised by any test in `DefaultIgnorableShapingTests` — every case pairs the ignorable with a real glyph (`"S" + codepoint`). The code is visibly correct for it (traced it by hand), but a one-line theory case would make that visible rather than implicit.

Everything else — the `UnicodeDefaultIgnorables` ranges (spot-checked against the UCD), the doc/license updates, the two new font fixtures and their `.LICENSE.txt` files, the accepted-gap update — looks accurate. `dotnet build -t:Rebuild` is 0 warnings/0 errors, and the targeted + broad regression suites (Font/Text/Svg/Gsub/Ligature/ArabicJoining/DefaultIgnorable) all pass.

jhaygood86:
One more thing worth fixing while this PR is already touching license documentation: `docs/license.md`'s font table is updated correctly (both new entries land there, matching the existing pattern for `ColorTestV0/V1.ttf`), but `THIRD-PARTY-LICENSES.md`'s "Bundled font assets" section isn't touched — its summary line still reads:

> **License:** varies per font (SIL OFL 1.1 or 3-Clause BSD)

That line was already missing CC0 before this PR (`ColorTestV0.ttf`/`ColorTestV1.ttf` are CC0), and `CcmpLigatureTest.ttf` adds a second CC0 font on top of it. Since this PR is the one touching that exact corner of the license docs, could you fold in a one-line fix there too — e.g. "SIL OFL 1.1, 3-Clause BSD, or CC0" — so the summary doesn't undersell what's actually bundled? `docs/license.md`'s own table stays the source of truth for the specifics either way.

jhaygood86 (approved):
All four points verified directly:

1. `CssBox.NeedsPerCodepointFont`'s comment now correctly names `DropHiddenIgnorables`.
2. `DroppingAnIgnorable_RemapsAMarkGlyphsAttachmentIndex` — this goes beyond what was asked: a real Arabic mark-attachment scenario (BEH decomposing into base+dot mark via `ccmp`, then GPOS mark-to-base recording the base's glyph-list index) that would directly catch a broken remap, not just a synthetic case. Plus the three empty-run fixtures (lone VS16, lone ZWJ, nothing-but-ignorables) covering `kept.Count == 0`.
3. Changing the recent-fixes note to say "re-measure rather than quote" is the right fix for the actual problem — the number drifted twice because it was transcribed instead of derived, and that's a documentation-process fix, not just a corrected digit.
4. `THIRD-PARTY-LICENSES.md`'s summary line now says "SIL OFL 1.1, 3-Clause BSD, or CC0 for the hand-authored test fixtures" — exactly matches what's actually bundled, and correctly leaves `docs/license.md` as the per-font source of truth rather than duplicating it.

Build clean, shaping-filtered suite (91 tests) passes.

Approving. Will merge once CI actually completes (not before, this time).

PR #985, opened 2026-09-10
Match ligatures without allocating per glyph position

dchupp:
Pure performance. No output byte changes.

## What was wrong

`ApplyLigatureLookup` tries a match at every position in the glyph run, and every attempt allocated before it had looked at anything:

- `TryMatchLigature`'s **first statement** was `skippedOffsets = []`, ahead of the coverage test that decides whether there is any work at all. Most positions match nothing, so most of those lists were built and dropped without ever being written to.
- Each candidate ligature then allocated its own `matched` and `skipped` lists, again before knowing whether it would match.
- `lookup.Subtables` is typed `IReadOnlyList<T>`, so the `foreach` over it boxed `List<T>`'s struct enumerator once more per position. Same defect, same shape, as the GPOS subtable walks in #968 — GSUB simply wasn't looked at then. Five sibling walks in this file (single, multiple, alternate, reverse-chain, sequence context) have it too and are converted the same way.

## Why it started mattering now

None of this is new, but **#984 made it hot**. `ccmp` and `locl` are Type 4 (ligature) lookups in real fonts, and that change enables them for *every* run rather than only for Arabic and USE text. Ordinary Latin text that previously ran no ligature lookup at all now runs one over every glyph position, so a per-position allocation that used to be paid by a minority of runs is now paid by all of them.

Measured on a 26-document corpus, `main` before and after #984: 2,457 MB → 2,638 MB, **+181 MB**, bisected to that single commit and reproducible to within 0.1%.

## The change

`skippedOffsets` is `null` on the failure path rather than an empty list, declared `[NotNullWhen(true)]` so the compiler knows the caller may only read it after a `true`. The two match lists become a `LigatureMatchScratch` owned by `ApplyLigatureLookup`, created lazily on the first candidate actually tried — so a run whose glyphs no subtable covers allocates nothing at all — and cleared per candidate rather than reallocated.

Clearing **per candidate, not per position**, is the load-bearing part: a candidate that fails partway leaves entries the next one must not see. `skipped` is handed to `ApplyLigatureAt` on a match and read there before the next call, so nothing ever observes a stale buffer.

`ApplyLigatureAt` also skips building its `skippedGlyphs` list when nothing was skipped, which is the ordinary case even on a match.

## Measurements

26-document corpus, medians of 3 interleaved reps, Release, net8.0, server GC, `FONTCONFIG_FILE` pinned so font resolution cannot drift between runs:

| | allocated | vs `main` |
| --- | ---: | ---: |
| `main` | 2,638.5 MB | — |
| this change | 2,444.4 MB | **−194.1 MB, −7.4%** |

Run-to-run spread under 0.1% on each side. That is more than the whole of #984's own cost — the correctness fix now runs at a **net saving** rather than a net cost, because the per-position allocation it exposed had been there all along on the `liga`/`clig` path.

Wall time trended lower too but I would not quote a figure from it: the c4 spread on this machine was wide enough that allocation is the only number here I would stand behind.

## Proof

Three tests, and each part of the change was reverted in turn to confirm the suite catches it:

| reverted | result |
| --- | --- |
| eager `skippedOffsets = []` | allocation test fails |
| `foreach` over `Subtables` | allocation test fails |
| `skipped.Clear()` | reuse test fails |
| `matched.Clear()` | reuse test fails |

- `WalkingARunForLigatures_AllocatesNothingPerPosition` — 10,000 match attempts against a coverage that matches nothing. Deliberately **not** an empty subtable list: `List<T>` hands back a cached singleton enumerator when empty, so the boxing under test disappears and the test passes against unfixed code. (That trap cost me a first attempt at #968's test.)
- `TwoLigaturesInOneRun_DoNotCarryTheFirstMatchsSkippedGlyph` — two ligatures form in one run, the first over a skipped mark. An uncleared skipped-offset list re-inserts that mark after the second ligature as well.
- `ALigatureInvokedFromAContextualRule_StillForms` — the other route into `ApplyLigatureAt`, a Type 5 contextual rule invoking a ligature lookup as a nested lookup at a single position.

- **Full net8.0 suite: 10,534 passed, 9 skipped, 0 failed** — including #984's own `DefaultIgnorableShapingTests` and `EmojiSequenceCompositionTests`, so emoji sequence composition and default-ignorable hiding are unchanged.
- Rendered PDFs are **byte-identical to `main`** across all 26 corpus documents after normalising `/CreationDate`, `/ID` and the random font subset tags.
- Diff coverage on the changed source: **100%** (28 of 28 coverable added lines), measured with `diff-cover` against the merge-base.

### DIFF
--- a/.claude/recent-fixes/2026-09-10-a-ligature-lookup-allocates-nothing-per-position.md
+++ b/.claude/recent-fixes/2026-09-10-a-ligature-lookup-allocates-nothing-per-position.md
@@ -1,0 +1,81 @@
+# A ligature lookup allocates nothing per glyph position
+
+Pure performance. No output byte changes.
+
+## What was wrong
+
+`GsubShaper.ApplyLigatureLookup` tries a match at every position in the glyph run, and every attempt
+allocated before it had looked at anything:
+
+- `TryMatchLigature` set `skippedOffsets = []` as its first statement, ahead of the coverage test
+  that decides whether there is any work at all. Most positions match nothing, so most of those
+  lists were built and dropped without ever being written to.
+- Each candidate ligature allocated its own `matched` and `skipped` lists, again before knowing
+  whether the candidate would match.
+- `lookup.Subtables` is typed `IReadOnlyList<T>`, so the `foreach` over it boxed `List<T>`'s struct
+  enumerator once per position. The same defect, in the same shape, as the GPOS subtable walks fixed
+  in #968 - GSUB was simply not looked at then.
+
+Five sibling subtable walks in the same file (single, multiple, alternate, reverse-chain, sequence
+context) had the identical boxing and are converted the same way.
+
+## Why it started mattering now
+
+None of this is new, but #984 made it hot. `ccmp` and `locl` are Type 4 (ligature) lookups in real
+fonts, and that change enables them for every run rather than only for Arabic and USE text. Ordinary
+Latin text that previously ran no ligature lookup at all now runs one over every glyph position, so a
+per-position allocation that used to be paid by a minority of runs is now paid by all of them.
+
+## The load-bearing idea
+
+`skippedOffsets` is `null` on the failure path rather than an empty list, declared
+`[NotNullWhen(true)]` so the compiler knows the caller may only read it after a `true`. The two
+match lists become a `LigatureMatchScratch` owned by `ApplyLigatureLookup` and cleared per candidate
+rather than reallocated - they are created lazily, on the first candidate actually tried, so a run
+whose glyphs no subtable covers still allocates nothing.
+
+Clearing per candidate, not per position, is what makes reuse safe: a candidate that fails partway
+leaves entries behind that the next candidate must not see, and `skipped` is handed out to
+`ApplyLigatureAt` on a match and read there before the next call, so nothing ever observes a stale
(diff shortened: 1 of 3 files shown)


jhaygood86 (approved):
Traced the highest-risk part by hand against a concrete two-ligature scenario (a mark-containing pair followed by a mark-free pair): the scratch buffers are correctly cleared at the top of every candidate's inner loop, before that candidate builds anything, and the previously-matched glyph's skip list is copied out into a fresh list at `ApplyLigatureAt` before the next candidate can touch the shared scratch — so nothing ever reads stale data across candidates. Recomputed `TwoLigaturesInOneRun_DoNotCarryTheFirstMatchsSkippedGlyph`'s expected output independently and got the same result.

Also independently rendered the same ligature-heavy document (exercising both `ccmp` via the #984 test font and `liga`/`clig` via Liberation Sans) on this branch and on its exact merge-base, and diffed the raw PDF bytes myself — content streams are byte-identical, only `/CreationDate`/`/ID`/the random subset tag differ, confirming the "no output byte changes" claim directly rather than trusting it.

Build clean, full suite (10,534 tests) passes, matching the PR's claimed numbers exactly. CI green on all three platforms.

Approving. Merging now.

PR #986, opened 2026-09-10
Make COLR/CPAL color glyphs searchable and selectable

### LINKED ISSUE #987: Color emoji (COLR/CPAL) cannot be selected, searched, or copied
## Summary

COLR/CPAL color glyphs render correctly, but they are not text as far as any PDF
viewer is concerned. A document full of color emoji has holes in it: you cannot
select an emoji, Ctrl+F will not find it, and copying a line that contains one
silently drops it.

## Repro

Render any document using a COLR/CPAL font (e.g. the COLR v1 build of Noto Color
Emoji), then open the PDF and try to select or search for an emoji.

```html
<!doctype html>
<html><head><meta charset="utf-8"><style>
  @font-face { font-family: 'Emoji'; src: url('NotoColorEmoji.ttf') format('truetype'); }
  .emoji { font-family: 'Emoji'; font-size: 42pt; }
</style></head>
<body>
  <p>Rainbow flag: <span class="emoji">&#127987;&#65039;&#8205;&#127752;</span></p>
</body></html>
```

**Expected:** the emoji behaves like text — selectable, findable, and copied back
as its original codepoint sequence.

**Actual:** it is vector artwork only. Selecting the paragraph and copying yields
`Rainbow flag: ` with nothing after it.

## Why it happens

`ColorGlyphPainter` decodes the COLR paint graph and emits the glyph as PDF
vector fills, deliberately skipping the font program and the text-showing
operator. That is the right call for the *artwork* — it is how gradients,
transforms, and blend-mode compositing get expressed as real vector content —
but it means nothing in the content stream tells a viewer that those paths are
text, and there is no `/ToUnicode` entry for the glyph.

## What a fix has to handle

- **Ambiguous CIDs.** A single COLR glyph can represent more than one source
  sequence. In Noto Color Emoji, `❤` and `❤️` (with a trailing U+FE0F) compose to
  the same glyph ID, so a glyph-keyed `/ToUnicode` map cannot round-trip both —
  the mapping has to be per-occurrence.
- **Characters shaping consumes.** Variation selectors, ZWJ, and emoji tag
  characters are deleted during shaping and have no glyph of their own, but they
  must still survive a copy.
- **Multi-codepoint sequences generally** — ZWJ sequences, regional-indicator
  pairs, skin-tone modifiers, and tag sequences all need to come back out
  exactly as they went in.
- **No visible change.** Whatever carries the text must paint no ink; the vector
  artwork stays the only thing rendered.

## Verification bar

Any fix here has to be checked against **both** PDFium and MuPDF. MuPDF is
markedly more lenient about what it will accept into a text page, so a
MuPDF-only check can show a feature working perfectly while Chrome and Edge
extract nothing at all.

ChrisVanDijk:
Fixes #987 

COLR/CPAL color glyphs have always painted correctly as native PDF vectors, but
they were invisible to text extraction: no `/ToUnicode`, no selection geometry,
nothing to copy or search. This adds that layer without changing a single pixel
of the visible output.

## What changed

Each color-font run still paints its artwork as vector fills, then emits one
shared rendering-mode-3 (invisible) text object over those vectors. Every
source-bearing glyph is wrapped in its own `/Span <</ActualText ...>>`
marked-content sequence.

`/ToUnicode` provides the ordinary CID mapping. Per-occurrence `/ActualText`
carries the exact source sequence, which a glyph-keyed map fundamentally cannot
do: in Noto Color Emoji, `❤` and `❤️` compose to the *same* CID, so only a
per-occurrence replacement string can round-trip both spellings.

A small subset of the color font is embedded solely to back that invisible text.
The real colored layer outlines are never embedded — they remain PDF vector
paths emitted from the in-memory font.

## The empty-base-glyph problem

A COLR base glyph usually has no `glyf` contour of its own; its artwork lives in
separate layer glyphs. PDFium omits a shown CID from its text page entirely when
that glyph has zero contours — so the first working version of this extracted
perfectly in MuPDF and produced **nothing at all** in Chrome and Edge.

`OpenTypeFontface.CreateFontSubSet` therefore substitutes a stand-in rectangle
for each selected empty base glyph. Sizing it to the real glyph box — its own
advance width by the font's ascent/descent — rather than to a token shape is
load-bearing: a 1×1-unit contour still extracts and still answers Ctrl+F, but
reports a 0.04pt character box against a 42pt emoji, so the search highlight is
invisible and a mouse drag has no hit target. Measured on a seven-emoji page,
1×1 units gave 0.042 × 0.042 pt boxes; a glyph-sized contour gives 52.3 × 49.8
pt. Text rendering mode 3 suppresses the contour at either size.

## Rejected alternative

Embedding the COLR layer closure was tried and removed. No text-showing operator
ever references a layer CID — only the base CID — so it does not stop PDFium
dropping the shown glyph, and it grew the seven-emoji subset from 1.6 KB to
9.5 KB. Recorded as an invariant so it does not get reintroduced.

## Verification

Checked against **both** PDFium and MuPDF, per this repo's two-renderer rule —
which is exactly what caught the original defect, since MuPDF happily accepts
the empty-glyph form that Chrome and Edge silently drop.

- Both engines extract all seven test sequences identically: plain heart, heart
  + VS16, skin-tone modifier, ZWJ sequence, regional-indicator pair, rainbow
  flag, and tag sequence.
- Rasters are byte-identical before and after, in both engines — the invisible
  text adds no ink.
- The emitted stand-in glyphs parse in fontTools as one contour with four
  on-curve points at the expected extents.
- Full suite green; solution builds with zero warnings.

## Also included

- Docs updated with the feature: `docs/html-css-support.md` and
  `docs/architecture.md` no longer claim color fonts are unembedded and
  unselectable, and the Color Fonts showcase copy is corrected. The stale
  "no text-shaping stage" and "color emoji is not supported" sections in
  `docs/architecture.md` are rewritten to match what the engine actually does.
- A migration note (behavior change, file-size increase, and color-font
  `.notdef` now reachable by the PDF/A conformance check) and an invariant note.
- A drive-by fix for a culture-dependent assertion in
  `FontSynthesisIntegrationTests`, which formatted its expected skew with the
  current culture and so failed on any comma-decimal locale.


jhaygood86 (approved):
Reviewed thoroughly given the scope, verifying the load-bearing claims directly rather than trusting the description:

**Pixel-identical**: rendered the same document (heart, heart+VS16, rainbow-flag ZWJ, US flag pair) on `main` and this branch, rasterized with both PDFium and MuPDF, diffed — extrema `(0,0)` in both engines. Genuinely no visible change.

**Text extraction**: confirmed the bug on `main` first (both engines return empty text for the emoji paragraphs), then confirmed this branch extracts the exact correct sequences in both engines.

**Per-occurrence `/ActualText`**: decompressed the actual content stream — heart and heart+VS16 both show `<0042> Tj` (same CID, confirming the shared-CID premise this whole design is built on) but carry different `/ActualText` values (`<FEFF2764>` vs `<FEFF2764FE0F>`). Genuine per-occurrence disambiguation.

**Tagged-PDF interaction**: rendered with `--tagged-pdf` — the new `/Span<</ActualText ...>>` BDC nests entirely inside the existing structure span without its own MCID, MCIDs stay sequential across paragraphs, and MuPDF extracts correctly from the tagged output too. No collision with the existing marked-content/structure-tree machinery.

**Stand-in glyph geometry**: confirmed in code as genuinely derived from the font's own `hmtx`/`hhea` tables, not a hardcoded size.

Build clean, targeted suite (27 tests) and broad Font/Svg/Pdf regression (10,535 tests) both pass — ordinary text rendering is unaffected.

Approving. Merging now.

PR #988, opened 2026-09-10
Optimize searchable COLR/CPAL glyph rendering

ChrisVanDijk:
Reduces CPU time and allocations across searchable/selectable COLR/CPAL glyph rendering, font subsetting, vector-path construction, and PDF serialization.

## Changes

- Reuse shaped glyph runs instead of shaping twice.
- Avoid sorting and per-glyph `StringBuilder` allocations for disjoint clusters.
- Write UTF-16 hex values using stack-allocated buffers.
- Return glyph data as `ReadOnlySpan<byte>` and copy directly into font subsets.
- Avoid unnecessary subset bookkeeping and temporary arrays.
- Format PDF path coordinates with stack-buffered `double.TryFormat`, avoiding boxing and composite formatting.
- Pre-size color-glyph path storage from decoded outlines.
- Pass path points and types to the renderer as read-only spans, eliminating per-path `ToArray()` copies.
- Add allocation, formatting, path-syntax, and ownership regression coverage.

## Performance

Initial searchable/selectable glyph optimizations:

- End-to-end color-glyph rendering: **9.9% faster**
- Ownership mapping allocations: **76% lower**
- Full-font subsetting allocations: **19% lower**
- UTF-16 hex serialization microbenchmark: **31% faster**

Follow-up vector-path optimizations, measured over 20 documents containing 100 Noto Color Emoji glyphs each:

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| Render allocations per document | 17.2 MB | 12.6 MB | **26.7% lower** |
| Total allocations per document | 21.0 MB | 16.4 MB | **21.8% lower** |
| Render time per document | ~21.0 ms | ~20.2 ms | **~4% faster** |

The generated PDF remained 756,801 bytes and was byte-identical after normalizing timestamps, random font-subset tags, and document IDs.

## Verification

- 10,548 net8.0 tests passed; 9 platform-specific tests skipped
- 97% diff coverage
- Full solution rebuild with zero warnings and errors
- Independent final code review found no issues

### DIFF
--- a/.claude/recent-fixes/2026-09-10-color-glyph-selection-avoids-redundant-work.md
+++ b/.claude/recent-fixes/2026-09-10-color-glyph-selection-avoids-redundant-work.md
@@ -1,0 +1,29 @@
+# Color-glyph selection avoids redundant shaping and transient buffers
+
+_Landed 2026-09-10._
+
+The invisible text added for searchable/selectable COLR glyphs originally repeated shaping after
+`DrawString` had already shaped the run for measurement. It also built cluster ownership through a
+sorted glyph-index list and one `StringBuilder` per glyph in the ordinary disjoint-cluster case, then
+created several temporary strings and byte arrays to serialize each `/ActualText` span and CID.
+
+The color path now shapes once and reuses that result for measurement, vector painting, source
+ownership, and invisible text. Disjoint clusters take a linear no-sort path and materialize their
+source ranges directly; overlapping clusters retain the widest-span precedence fallback. UTF-16BE
+hex strings and CIDs append directly to the content stream. Subsetting likewise reads contour counts
+without copying glyph data, allocates selection bookkeeping only for fonts that need it, and writes
+synthetic selection contours directly into the destination `glyf` table.
+
+A release-mode benchmark rendering and saving twenty 100-emoji documents measured median elapsed
+time falling from 888.2 ms to 800.0 ms (9.9%) and allocations from 424,484,096 to 420,372,992 bytes
+(1.0%). The isolated ownership mapper fell from 3,061,600 to 727,200 bytes over one hundred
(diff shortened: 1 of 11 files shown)


jhaygood86 (approved):
Reviewed with extra scrutiny given this PR's description reads noticeably differently from your usual detailed style — verified everything myself rather than taking the summary at face value, and it holds up:

Traced the shaping-reuse logic precisely: pre-PR, a color-font `DrawString` shaped twice with identical arguments (once for measurement, once for painting). This PR shapes once and reuses it for painting always, and for width measurement whenever `CanMeasureAsSingleShapedRun` holds (no control characters), correctly falling back to the untouched per-line path otherwise. Confirmed `MeasureShapedRunWidth`'s formula is bit-for-bit identical to the old measurement path's single-line case, so nothing legitimately divergent got collapsed.

Checked every `stackalloc` individually — both are correctly sized for their actual worst case (`AppendPdfNumber`'s 64-char buffer even falls back to `ToString` for values `TryFormat` can't fit, which is tested against `double.MaxValue`), and neither escapes its synchronous call frame (no `async` methods in the touched renderer file, so no suspension-across-stackalloc risk). The `ReadOnlySpan<byte>` glyph data spans into the font's own immutable backing array and is consumed synchronously within one loop, never stored.

Independently reproduced the byte-identical claim with a standalone harness rather than trusting it: rendered the same emoji document against #986's merge commit and against this branch, normalized `/CreationDate`/`/ModDate`/`/ID`/subset tags myself, and got byte-for-byte identical output — then rasterized with both PDFium and MuPDF to confirm it still renders correctly.

Build clean, targeted suite (52 tests) and full suite (10,548 tests) both pass, matching the PR's claimed count exactly.

Approving. Merging now.

PR #990, opened 2026-09-10
Write PDF content streams directly as bytes

ChrisVanDijk:
Replaces the intermediate UTF-16 content-stream buffer with a chunked byte writer, reducing allocations and CPU usage in graphics-heavy PDFs.

### Changes

- Write PDF operators directly to adaptive byte buffers.
- Grow content chunks from 256 bytes to a 16 KiB cap.
- Preserve byte-exact numeric formatting and raw-encoding behavior.
- Avoid intermediate `StringBuilder`, `string`, and encoding allocations.
- Parse composite glyph components directly from their source byte spans.
- Prevent font subsetting from moving the process-wide shared font cursor.
- Adjust the allocation regression threshold for the reduced plain-block baseline.

### Test coverage

- Add dedicated tests for the complete `PdfContentWriter` public surface.
- Verify both constructors and invalid chunk sizes.
- Cover every `Append` and `AppendFormat` overload.
- Cover null, empty, exact-fit, and multi-chunk inputs.
- Verify all raw byte values from `0x00` through `0xFF`.
- Verify invariant formatting and format-buffer reuse.
- Exercise every adaptive growth boundary through repeated 16 KiB chunks.
- Verify `Clear()` resets the growth schedule and supports byte-exact reuse.
- Add regression coverage proving composite glyph closure leaves the shared font cursor unchanged.

### Additional fix

Ubuntu CI exposed a race between font subsetting and GSUB shaping. Both operations could access the same process-wide cached `OpenTypeFontface`, while composite glyph closure moved its shared `Position` cursor without coordinating with GSUB's lock.

Composite glyph closure now parses its local `ReadOnlySpan<byte>` directly and no longer touches the shared cursor.

### Performance

- Render allocations: **12.6 MB → 7.5 MB/document** (**40.3% lower**)
- Render time: approximately **10% faster**
- `char[]` allocations: **~90% lower**
- `string` allocations: **~82% lower**
- `StringBuilder` allocations: **~91% lower**
- Generated PDF size and normalized bytes unchanged

### Verification

- **10,563 tests passed**, 9 skipped, 0 failed
- **100% diff coverage** across 154 changed production lines
- Full solution rebuild with zero warnings
- CI verified the font-subsetting fix across Windows, Ubuntu, and macOS
- Final code review found no issues

### DIFF
--- a/.claude/invariants/fonts-composite-closure-does-not-move-the-shared-face-cursor.md
+++ b/.claude/invariants/fonts-composite-closure-does-not-move-the-shared-face-cursor.md
@@ -1,0 +1,10 @@
+# Composite glyph closure does not move the shared font-face cursor
+
+`OpenTypeFontface` instances are cached process-wide, and `Position` is one mutable cursor shared by all
+table readers. `GlyphDataTable.CompleteGlyphClosure` runs during PDF font subsetting while another render
+or xUnit class can shape text against the same face.
+
+Composite closure must parse the `ReadOnlySpan<byte>` returned by `GetGlyphData`; it must not seek or read
+through `OpenTypeFontface.Position`. Locking the glyph-table method does not coordinate with GSUB's
+font-face lock. The measured failure was simultaneous `IndexOutOfRangeException` crashes in full-font
+subsetting and SVG small-caps GSUB lookup on Ubuntu CI.
--- a/.claude/invariants/pdf-standard-fixed-format-is-not-custom-optional-decimal-format.md
+++ b/.claude/invariants/pdf-standard-fixed-format-is-not-custom-optional-decimal-format.md
@@ -1,0 +1,13 @@
+# Standard fixed-point formatting is not the PDF writer's custom format
+
+The PDF graphics writer historically formats coordinates with custom invariant patterns such as
+`0.####`. Do not replace that with `Utf8Formatter`'s `F4` output followed by trimming trailing zeroes.
+
+A randomized comparison found real finite coordinates where the last emitted decimal differed (for
+example, one path produced `-225163.6761824` with the custom format but `-225163.6761823` through the
+apparently equivalent fixed-point route). Both APIs ultimately perform decimal conversion, but their
+rounding paths are not interchangeable.
+
+Any numeric-writer optimization must compare against `double.ToString`/`TryFormat` using the exact
+existing custom format over boundary cases and a broad finite-value sample. PDF byte equivalence is the
+contract; visually indistinguishable coordinates are not sufficient.
--- a/.claude/recent-fixes/2026-09-10-font-subsetting-no-longer-races-shaping.md
+++ b/.claude/recent-fixes/2026-09-10-font-subsetting-no-longer-races-shaping.md
@@ -1,0 +1,18 @@
+# Font subsetting no longer races shaping
+
+`GlyphDataTable.CompleteGlyphClosure` used the process-wide cached `OpenTypeFontface.Position` cursor to
+read composite-glyph components. Its method-level synchronization locked the glyph table, not the font
+face, so it did not coordinate with GSUB readers that correctly lock the shared face.
+
+The full-font allocation test repeatedly subset Source Sans while an SVG small-caps test read that same
+cached face's GSUB table. On Ubuntu CI, their cursor writes interleaved: both tests failed with
+`IndexOutOfRangeException`, one in `CreateFontSubSet` and one in `GetActiveLookupIndices`. Adding bounds
+checks would only have changed the symptom.
+
+Composite closure now parses each glyph's existing `ReadOnlySpan<byte>` directly with big-endian span
+readers. It neither moves nor locks the shared cursor and removes the old synchronized-method overhead.
+A deterministic regression test verifies that closure leaves `OpenTypeFontface.Position` unchanged.
+
+The focused COLR/CPAL and SVG font-variant classes passed together (24 tests), followed by all 10,551
+runnable net8.0 tests. The complete branch retained 100% diff coverage, the solution rebuilt with zero
+warnings, and the independent diff review found no actionable issues.
--- a/.claude/recent-fixes/2026-09-10-pdf-content-streams-write-bytes-directly.md
+++ b/.claude/recent-fixes/2026-09-10-pdf-content-streams-write-bytes-directly.md
@@ -1,0 +1,36 @@
+# PDF content streams write raw bytes directly
+
+_Landed 2026-09-10._
+
+`XGraphicsPdfRenderer` formerly accumulated every page or form in a `StringBuilder`, materialized one
+large string at close, and raw-encoded that string into a new byte array. For a 100-emoji document this
+character pipeline remained the largest allocation source after path formatting stopped boxing its
+coordinates.
+
+The renderer now writes into `PdfContentWriter`, a chunked raw-byte accumulator. Small streams begin
+with a 256-byte chunk while larger streams grow to fixed 16 KiB chunks, avoiding both a large minimum
+allocation and large-object-heap growth copies. Existing composite-format call sites use one reusable
+scratch `StringBuilder`; path coordinates still use the exact invariant custom numeric format, then copy
+their stack-formatted characters directly into the byte chunks. Closing a renderer performs one
+exact-sized byte copy into the `PdfStream`, with no whole-content `char[]`, string, or encoding pass.
+
+A tempting direct UTF-8 fixed-point formatter was rejected after randomized equivalence testing found a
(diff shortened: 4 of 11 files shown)


jhaygood86 (changes requested):
CI is now green on all three platforms — confirmed the Ubuntu-only `IndexOutOfRangeException` failures were caused by `GlyphDataTable.AddCompositeGlyphs` reading through the process-wide shared `OpenTypeFontface.Position` cursor while GSUB shaping read the same cached face on another xUnit test-class thread (method-level `[Synchronized]` locked the glyph table, not the shared face, so it didn't coordinate with GSUB's own lock). The latest commit's fix — parsing composite glyphs entirely from the local `ReadOnlySpan<byte>` instead of the shared cursor, with a regression test asserting `Position` is unchanged — is correct and well-evidenced. I rebuilt (0 warnings) and ran the full net8.0 suite locally against the updated branch (10,551 passed, 0 failed), matching CI.

One gap remains in `PdfContentWriter`'s own test coverage. `EnsureCurrentChunk()`'s geometric growth (256 → 1024 → 4096 → capped at 16384 bytes) is the path every real page/form render exercises, since `XGraphicsPdfRenderer` always constructs `PdfContentWriter` with the parameterless constructor. None of the three new/updated unit tests in `XGraphicsTests.cs` actually exercise that growth logic:

- `AppendPdfNumber_UsesAllocationFreeFastPathAndPreservesFallbackFormatting` uses `new PdfContentWriter(100_000)` for its main assertion (fixed chunk size, growth disabled), and its one use of `new PdfContentWriter()` (line 238) only appends a single formatted `double.MaxValue` value — well under the 256-byte initial chunk, so it never even reaches the first growth transition.
- `AppendPdfNumber_MatchesInvariantCustomFormatting` uses `new PdfContentWriter(128)` (fixed, growth disabled).
- `PdfContentWriter_PreservesRawEncodingAcrossChunksAndFormatting` uses `new PdfContentWriter(4)` and `new PdfContentWriter(0)` (fixed sizes, growth disabled).

So the only code path production actually uses (`_usePool`-eligible geometric growth via the parameterless constructor) has zero dedicated verification that content written across a 256→1024→4096→16384 boundary crossing round-trips correctly through `ToArray()`/`ToString()`. The three chunk-boundary tests that do exist all use small, fixed chunk sizes instead, which is a different code path (`_maximumChunkSize == _initialChunkSize`, no growth branch ever taken). A test using `new PdfContentWriter()` that writes enough content to cross multiple growth boundaries (e.g. thousands of small appends) and asserts the concatenated output is byte-exact would close this gap.

Everything else about this change looks solid — the byte-writer approach, the growth arithmetic (verified by inspection to never exceed `_maximumChunkSize`), and `ToArray()`'s reliance on each completed chunk's own `.Length` all check out.

jhaygood86:
Following up on the growth-path coverage gap from my review: I'd suggest pulling this into its own dedicated `PdfContentWriterTests.cs` under `PeachPDF.Tests/PdfSharpCore/Drawing.Pdf/` rather than adding more cases into `XGraphicsTests.cs` — the writer is a standalone class with its own contract, and testing it in isolation (no `XGraphics`/`PdfPage` scaffolding needed) makes the growth behavior easy to verify directly. Concretely, that file should cover:

1. **The production growth path** — using `new PdfContentWriter()` (the parameterless constructor every real render uses), write enough varied content to cross every documented boundary (256 → 1024 → 4096 → 16384, then confirm the 16384 cap repeats rather than growing further or misbehaving — i.e. write past at least two 16384-byte chunks). Assert the reconstructed `ToArray()`/`ToString()` is byte-exact against an independently-built expected value, and that `Length` matches at each cumulative boundary (255/256, 1279/1280, 5375/5376, 21759/21760, 38143/38144) to catch an off-by-one right at a chunk seam.
2. **`Clear()` followed by reuse** — the existing `PdfContentWriter_PreservesRawEncodingAcrossChunksAndFormatting` test only checks that `Clear()` leaves the writer empty; nothing currently verifies that `_nextChunkSize` correctly resets to the initial size and that a writer instance produces correct output when reused for a second document/page after `Clear()`, which is exactly the pattern `XGraphicsPdfRenderer` relies on.

I'm flagging this rather than writing it myself so you can fold it in alongside whatever shape you'd prefer for the fixture data — happy to take another look once it's up.

jhaygood86:
Expanding on the dedicated test-file ask — here's the full checklist I'd want covered for `PdfContentWriter` to call it exhaustive, organized by member:

**Constructors**
- `PdfContentWriter()` — starts empty (`Length == 0`, `ToArray()` empty, `ToString() == ""`).
- `PdfContentWriter(chunkSize)` for a representative positive size — starts empty, same as above.
- `PdfContentWriter(0)` and a negative size both throw `ArgumentOutOfRangeException` (only `0` is currently covered).

**`Append(char)`**
- Single and repeated appends build the expected byte sequence.
- A char value `> 0xFF` (e.g. `'ā'`, already used elsewhere) truncates via the raw `(byte)value` cast rather than throwing — worth asserting explicitly here since it's the one `Append` overload that can silently lose data, and this class is the sole boundary where content becomes "raw PDF bytes."

**`Append(string?)`**
- `null` is a no-op (`Length` unchanged) — currently untested.
- `""` is a no-op.
- Ordinary content appends correctly.

**`Append(ReadOnlySpan<char>)` and `Append(ReadOnlySpan<byte>)`**
- Empty span is a no-op for both.
- A span whose length exactly fills the remaining capacity of the current chunk (the `count == _currentChunk.Length - _currentCount` edge in the `while` loop).
- **A single `Append` call whose span is larger than one chunk**, spanning multiple chunks *within that one call* — distinct from many small appends accumulating across chunks, since it's the only case that exercises the `while (!value.IsEmpty)` loop crossing more than one `EnsureCurrentChunk()` transition per call. This applies to both overloads.
- `Append(ReadOnlySpan<byte>)` preserves raw byte values across the full `0x00`-`0xFF` range, not just ASCII.

**The four `AppendFormat` overloads** (`object?[]`, `arg0`, `arg0+arg1`, `arg0+arg1+arg2`)
- Each overload individually produces correct, invariant-culture output (decimal separator etc. unaffected by `CultureInfo.CurrentCulture`).
- Two consecutive `AppendFormat` calls don't leak content between them — proves the internal format buffer (`PrepareFormatBuffer`/`_formatBuffer.Clear()`) is actually being reset, not just happening to look right for one call.
- A format call whose output is long enough to itself cross a growth boundary.

**The production growth path** (as I mentioned before) — `new PdfContentWriter()`, content crossing every documented boundary (256 → 1024 → 4096 → 16384, cap repeating past two 16384 chunks), byte-exact `ToArray()`/`ToString()`, and `Length` checked at each cumulative boundary (255/256, 1279/1280, 5375/5376, 21759/21760, 38143/38144).

**`ToArray()`**
- Empty writer returns `[]`, not null.
- Multi-chunk content concatenates in the correct order with no gaps/overlaps.

**`ToString()`**
- Round-trips through `PdfEncoders.RawEncoding` correctly, including high-byte values (`0x80`-`0xFF`) as the file's own header comment describes ("byte-to-char identity mapping") — not just ASCII content.

**`Clear()`**
- Resets `Length` to 0 and `ToArray()` to empty (existing coverage, fixed-chunk-size only).
- Calling it on an already-empty writer is a safe no-op.
- **Growth schedule resets correctly** — after `Clear()` on a parameterless-constructor writer that had already grown past 256 bytes, the next write starts a fresh 256-byte chunk rather than continuing from wherever `_nextChunkSize` left off; verify by writing enough after `Clear()` to cross the first boundary again and checking it lands correctly.
- **Reuse after `Clear()`** produces byte-exact output with no leftover data from the prior content — this is exactly the pattern `XGraphicsPdfRenderer` relies on when reusing a writer across pages/forms.

**Fluent chaining**
- `Append`/`AppendFormat` return `this`; a chained sequence (`writer.Append(...).AppendFormat(...).Append(...)`) produces the same result as the equivalent sequence of separate statements.

That's the full public surface. Happy to review once it's up.

jhaygood86 (approved):
The dedicated PdfContentWriterTests.cs is exhaustive — it precisely covers every item from my checklist: both constructors plus negative-size rejection, char/string/span-of-char/span-of-byte Append behavior (including the >0xFF truncation case and full 0x00-0xFF byte range), all four AppendFormat overloads with culture-invariance and buffer-reset verification, the production growth path checked at every exact cumulative boundary (255/256/1279/1280/5375/5376/21759/21760/38143/38144/38145), high-byte ToString() preservation, Clear()'s growth-schedule reset (verified via allocation-size proxy) and reuse correctness, and fluent chaining. Rebuilt clean (0 warnings) and ran the full net8.0 suite locally against the latest commit (10,563 passed, 0 failed). CI is green on all three platforms. Approving.

PR #992, opened 2026-09-10
Match CSS keywords ordinally, not through the invariant culture

dchupp:
`css-properties.json` marked 33 properties `"keywordComparison": "invariant-ignore-case"`, and `RegistryEmitter` turned that into `StringComparison.InvariantCultureIgnoreCase` in both the generated validator and the generated canonicaliser. Invariant comparison is *linguistic* — it runs through `CompareInfo.Compare`, roughly an order of magnitude more expensive than an ordinal scan.

## How it was found

A sampled CPU profile of the engine rendering a 26-document corpus — warmed 15 sweeps past the plateau, pinned to 4 cores, waiting threads excluded — put **6.28% of all work in `CompareInfo.Compare`, 89.9% of it under a single generated setter** (`Set_OutlineColor`). Nothing else near the top of that profile was a string comparison at all.

This came out of looking for what allocation profiling *cannot* see: a 13.4% allocation reduction across two earlier changes bought only ~3% throughput, which said fairly clearly that allocation had stopped being the constraint.

## Why ordinal is right

css-values-4 §4.1 "Pre-defined Keywords" defines keyword matching as **ASCII case-insensitive** ("Keywords are identifiers and are interpreted ASCII case-insensitively (i.e., [a-z] and [A-Z] are equivalent)."), and every keyword in `supportedValues` is ASCII — checked across all 215 properties, none has a non-ASCII declared value. The invariant culture bought nothing the spec asks for.

Both emit sites now use `OrdinalIgnoreCase`, and `KeywordComparison` loses its `InvariantIgnoreCase` member so nothing can ask for a culture-aware keyword match again.

The JSON spelling is still **accepted** and maps to ordinal rather than being rejected. Falling through to the `Ordinal` default would make a keyword match only its exact declared casing — turning `INVERT` from a match into a non-match, a far worse regression than the culture-awareness being removed.

## This is not a correctness fix

Worth stating plainly, because it looks like one and I initially wrote it up as one.

The comparisons genuinely differ in isolation — linguistic comparison gives format characters zero weight, so `"in­vert".Equals("invert", InvariantCultureIgnoreCase)` is `true` where the ordinal form is `false`. That invites the conclusion that `outline-color: in­vert` used to canonicalise into `invert`.

**It did not.** Such a declaration is rejected *before* the generated setter's keyword comparison ever runs, so the property falls back to its initial value either way. I established that by reverting the emitter to `InvariantCultureIgnoreCase`, rebuilding from clean, and getting **identical** results for a soft hyphen, a zero-width space, a leading ZWSP and a trailing ZWJ — across both a `<color> | keyword` property and a keyword-only one.

So the culture-aware comparison was unreachable cost: it changed no output and consumed 6% of the engine's CPU.

## Evidence

Sampled thread-time profiles before and after, same corpus, same 15-sweep warm-up, same 60-second window, 4 cores, waiting threads excluded:

| frame | before | after |
| --- | ---: | ---: |
| **`CompareInfo.Compare`** | **6.28%** | **0.00%** |
| `List<CSS.Token>..ctor(IEnumerable)` | 7.32% | 13.41% |
| `Enumerable.Select().ToArray()` | 4.99% | 5.18% |
| `FcConfigSubstitute` | 2.69% | 2.65% |
| `GraphicsAdapter.DrawString` | 4.43% | 4.50% |

The comparison is gone outright. Every other frame is unchanged; the ones whose share rose did so because the same work is a larger fraction of a smaller total, not because they got slower.

**End-to-end throughput is deliberately not quoted.** This is worth ~6% of CPU in a pipeline that is not purely CPU-bound, and a bare-process run-to-run spread of 10–15% cannot resolve that from four reps. The profile measures the thing that changed; wall clock would only measure the noise.

## Tests

`OutlineColorKeyword_MatchesAsciiCaseInsensitivelyAndNotLinguistically` is kept as a **guard** on the contract and is documented as passing either way, so nobody later mistakes it for evidence of a behaviour change.

The proof for this change is the golden-file tests — reverting either emit site fails three of them (`..._For_OrdinalIgnoreCase`, `..._For_The_Legacy_InvariantIgnoreCase_Spelling`, `..._For_A_Union_Length_Or_Keyword_Property`).

- Full net8.0 suite: **10,555 passed**, 9 skipped, 0 failed.
- Source-generator suite: **119 passed**, 0 failed.

### DIFF
--- a/.claude/recent-fixes/2026-09-10-css-keywords-match-ordinally.md
+++ b/.claude/recent-fixes/2026-09-10-css-keywords-match-ordinally.md
@@ -1,0 +1,78 @@
+# CSS keywords match ordinally, not through the invariant culture
+
+Pure performance. No behaviour change — see "What this is not", which is the important half.
+
+## What was wrong
+
+`css-properties.json` marked 33 properties `"keywordComparison": "invariant-ignore-case"`, and
+`RegistryEmitter` turned that into `StringComparison.InvariantCultureIgnoreCase` in both the
+generated validator and the generated canonicaliser. Invariant comparison is *linguistic*: it goes
+through `CompareInfo.Compare`, which is roughly an order of magnitude more expensive than an
+ordinal scan.
+
+A sampled CPU profile of the engine rendering a 26-document corpus, warmed 15 sweeps past the
+plateau and pinned to 4 cores, put **6.28% of all non-waiting work in `CompareInfo.Compare`, and
+89.9% of that under one generated setter** (`Set_OutlineColor`). Nothing else in the top of that
+profile was a string comparison at all.
+
+## Why ordinal is right
+
+css-values-4 §4.1 "Pre-defined Keywords" defines keyword matching as **ASCII
+case-insensitive**: "Keywords are identifiers and are interpreted ASCII case-insensitively
+(i.e., [a-z] and [A-Z] are equivalent)." Every keyword in `supportedValues` is ASCII —
+checked, all 215 properties, none has a non-ASCII declared value. So
+the invariant culture bought nothing the spec asks for, and `OrdinalIgnoreCase` is both correct
+and cheap.
+
+`invariant-ignore-case` is still accepted in the JSON and now maps to ordinal, rather than being
+rejected. Rejecting it would fall through to the `Ordinal` default, which is *not* case-insensitive
+at all — turning `INVERT` from a match into a non-match. That would be a far worse regression than
+the culture-awareness being removed.
+
+## What this is not
+
+**It is not a correctness fix, and the first draft of this note wrongly said it was.**
+
+The two comparisons genuinely differ in isolation: linguistic comparison gives format characters
+zero weight, so `"in­vert".Equals("invert", InvariantCultureIgnoreCase)` is `true` where the
+ordinal form is `false`. It is tempting to conclude the engine used to canonicalise
+`outline-color: in­vert` into `invert`.
+
+It did not. Driven end to end, such a declaration is rejected *before* the generated setter's
+keyword comparison runs, so the property falls back to its initial value either way. That was
+established the only way it could be — by reverting the emitter to `InvariantCultureIgnoreCase`,
+rebuilding from clean, and getting **identical** results for a soft hyphen, a zero-width space, a
+leading ZWSP and a trailing ZWJ, across both a `<color> | keyword` property and a keyword-only one.
+
+So the culture-aware comparison was unreachable cost: it changed no output and consumed 6% of the
+engine's CPU. `OutlineColorKeyword_MatchesAsciiCaseInsensitivelyAndNotLinguistically` is kept as a
+guard on the contract and is documented as passing either way, so nobody later mistakes it for
+evidence of a behaviour change.
+
+## Evidence
+
+Sampled thread-time profiles before and after, same corpus, same 15-sweep warm-up, same 60-second
+window, 4 cores, waiting threads excluded:
+
+| frame | before | after |
+| --- | ---: | ---: |
+| `CompareInfo.Compare` | **6.28%** | **0.00%** |
+| `List<CSS.Token>..ctor(IEnumerable)` | 7.32% | 13.41% |
+| `Enumerable.Select().ToArray()` | 4.99% | 5.18% |
+| `FcConfigSubstitute` | 2.69% | 2.65% |
+| `GraphicsAdapter.DrawString` | 4.43% | 4.50% |
+
+The comparison is gone outright. Every other frame is unchanged; the ones whose percentages rose
+did so because the same work is now a larger share of a smaller total, not because they got slower.
+
+End-to-end throughput is **not** quoted here. The change is worth ~6% of CPU in a pipeline that is
+not purely CPU-bound, and a bare-process run-to-run spread of 10-15% cannot resolve that from four
+reps. The profile measures the thing that changed; wall clock would only measure the noise.
+
+- Full net8.0 suite: 10,555 passed, 9 skipped, 0 failed.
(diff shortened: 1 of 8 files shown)


jhaygood86 (changes requested):
This is a rigorously-verified change and I independently confirmed both of its central claims rather than taking the write-up at face value:

- **"Not a correctness fix"**: I temporarily reverted both `RegistryEmitter.BuildKeywordCanonicalizationExpression` and `ValidatorExpressionBuilder.BuildKeywordClause` back to `InvariantCultureIgnoreCase`, rebuilt from clean, and ran `OutlineColorKeyword_MatchesAsciiCaseInsensitivelyAndNotLinguistically` — all 7 cases (including the four format-character variants) produced identical results to the ordinal version. Confirmed: the format-character leniency was genuinely unreachable.
- **Golden-file mutation coverage**: with that same revert in place, the source-generator suite failed exactly 3 tests (116 passed, 3 failed) — matching the PR's claimed `..._For_OrdinalIgnoreCase`/`..._For_The_Legacy_InvariantIgnoreCase_Spelling`/`..._For_A_Union_Length_Or_Keyword_Property` trio precisely.
- I also independently re-checked the "every keyword is ASCII" claim by parsing `css-properties.json` myself (215 properties, `supportedValues`/`supportsSupportedValues`) — zero non-ASCII values found, matching the PR's own count.
- Rebuilt clean (0 warnings) and ran the full net8.0 suite (one unrelated allocation test, `CompoundSelectorAll_PlainBranch_AllocatesNothingPerCall`, failed under full-suite load but passed cleanly in isolation — pre-existing JIT-tiering flakiness in CSS selector matching, untouched by this diff, not a regression here) and the generator suite (119/119 passed).

One real issue: **the spec citation is wrong.** `css-values-4 §3.1.5` doesn't exist in the current CSS Values and Units Module Level 4 — I fetched the spec directly and confirmed §3 is "Combining Values: Interpolation, Addition, and Accumulation" (with only a §3.1 "Range Checking" subsection). The actual "Keywords are identifiers and are interpreted ASCII case-insensitively" text you're quoting correctly is in **§4.1, "Pre-defined Keywords"**. This citation appears in three places — the PR description, `.claude/recent-fixes/2026-09-10-css-keywords-match-ordinally.md`, and the code comment in `RegistryEmitter.cs` (`BuildKeywordCanonicalizationExpression`) — and should be corrected to §4.1 in all three before merge, since a future reader following "§3.1.5" would land on an unrelated section.

The quoted principle itself is accurate, and everything else about this change — the emitter/validator agreement, the legacy JSON spelling being remapped rather than rejected, the golden-file and CssUtilsTests coverage — is solid. Requesting changes only for the citation fix.

dchupp:
You are right about the citation, and I checked it rather than taking the correction on trust — the skill I follow here says to fetch every section before citing it, *including one a maintainer hands you*, because verifying a correction has previously turned up a better citation than the one suggested.

Fetching https://www.w3.org/TR/css-values-4/ and listing its headings:

```
3.   Combining Values: Interpolation, Addition, and Accumulation
3.1.   Range Checking
4.   Textual Data Types
4.1.   Pre-defined Keywords
4.1.1.   CSS-wide keywords: initial, inherit and unset
```

There is no §3.1.5 — §3.1 has no subsections at all. §4.1 carries the sentence verbatim:

> Keywords are identifiers and are interpreted ASCII case-insensitively (i.e., [a-z] and [A-Z] are equivalent).

That is the rule this change implements, so §4.1 is both the correct section and the right one to cite. (For completeness, the same document states the ASCII rule separately for unit identifiers in §5.4 and for function names in §9 — neither is what this diff touches, so §4.1 is the whole of it.)

Corrected in all three places, and I quoted the spec sentence in the two prose copies so a future reader does not have to re-fetch:

- PR description (above)
- `.claude/recent-fixes/2026-09-10-css-keywords-match-ordinally.md`
- the code comment in `RegistryEmitter.BuildKeywordCanonicalizationExpression`

`ValidatorExpressionBuilder.BuildKeywordClause` carried no citation, so there was no fourth copy to fix; a tree-wide grep for the wrong section number now returns only `xunit.runner.visualstudio` version strings.

Also rebased onto current `main` (`b23abbf6`) — the branch was two commits behind and no file overlaps. Re-ran there: source-generator suite 119/119, main suite 10580 passed / 9 skipped, and the build adds no warnings (6 CS1998s, all pre-existing in files this diff does not touch).

One caveat on that suite number, since it is not a clean 0-failed. `AnonymousBoxDefaultingTests.AListItemCostsLittleMoreThanAPlainBlock` fails on my box at **2.68x against the 2.55 bound** (143.1 KB vs 53.4 KB per item). It is not this branch: checked out `origin/main` with no commit of mine on top, the same test reports the identical 2.68x / 143.1 KB / 53.4 KB, and it passes at `4eb48ea7`, the commit before #990 raised that bound from 2.35 to 2.55. Your CI passes it on both net8.0 and net10.0, so I am treating it as an artefact of my local setup — no .NET 10 SDK here, so I run the net8.0 target rolled forward onto the 9.0 runtime, and an allocation ratchet is exactly the kind of test that measures differently under a different runtime. Flagging it only because the new bound apparently has less headroom on a runtime CI does not exercise; nothing here asks you to act on it.

jhaygood86 (approved):
Citation fixed — confirmed both the recent-fixes doc and the RegistryEmitter.cs comment now correctly cite css-values-4 §4.1 "Pre-defined Keywords", and diffed the updated commit against the merge base to confirm only the comment/doc text changed (the emitter/validator logic is byte-identical to what I already independently verified). Rebuilt clean (0 warnings), full net8.0 suite passes (10,581 passed, 0 failed), generator suite passes (119/119). CI is green on all three platforms. Approving.

PR #993, opened 2026-09-10
Presize TokenValue's list from the count it already has

dchupp:
`TokenValue`'s only sequence constructor took `IEnumerable<Token>` and did `new List<Token>(tokens)`. That overload presizes **only** when the source is an `ICollection<T>`. `IReadOnlyList<T>` is not one — and neither is `TokenValue` itself, which implements `IReadOnlyList<Token>` and stops there — so it enumerated the source and grew the backing array 4, 8, 16, 32…, reallocating and copying at every step, for a sequence whose length was known before the first element was read. It allocated an interface enumerator to do it, too.

Converters build one of these per declared value — `Original = new TokenValue(tokens)` appears at a dozen sites — and every one of them has `IReadOnlyList<Token>` as its static parameter type. So the slow path wasn't an edge case, it was the only path.

## The change

A second constructor taking `IReadOnlyList<Token>`, presizing from `Count` and copying by index. Overload resolution picks it at every existing call site without touching them.

## Evidence

Sampled thread-time profiles of the engine rendering a 26-document corpus — 15 warm-up sweeps past the plateau, 4 cores, 60-second window, waiting threads excluded:

| frame | before | after |
| --- | ---: | ---: |
| **`List<CSS.Token>..ctor(IEnumerable)`** | **7.29%** | **1.67%** |
| `Enumerable.Select().ToArray()` | 5.12% | 5.08% |
| `FcConfigSubstitute` | 2.67% | 2.65% |
| `GraphicsAdapter.DrawString` | 4.39% | 4.40% |

About **5.6 percentage points of engine CPU**, and 77% of that frame's own cost. Everything else is unchanged; frames whose share rose did so because the same work is a larger fraction of a smaller total.

## Allocation barely moves, and that's the interesting part

**2,260 MB → 2,254 MB over the corpus, −0.3%.**

I'd rather state that than bury it: the win here is enumeration and copying, not bytes. A token list is short, so growing it wastes little memory and a lot of time. An allocation profile ranks this near the bottom — it was the largest single frame in a CPU profile.

That's also why I went looking with a CPU profile at all: a 13.4% allocation reduction across two earlier changes bought only ~3% throughput, which said allocation had stopped being the constraint.

## Tests

`BuildingFromAKnownLengthList_GrowsItsArrayOnlyOnce` is mutation-verified — with the overload removed, the same 64-token build allocates **5,256 bytes against a 2,560-byte one-array floor**, and the test fails naming that number.

It deliberately wraps its source so only `IReadOnlyList<Token>` is visible. Handing it the `List<Token>` directly would let the general overload presize from `ICollection<T>` and the test would pass against unfixed code — the same trap that cost a first attempt at the GPOS enumerator test in #968.

- Full net8.0 suite: **10,576 passed**, 9 skipped, 0 failed.
- Diff coverage: **100%** (5 of 5 coverable added lines).
- A 26-document corpus renders **byte-identically** after normalising `/CreationDate`, `/ID` and the random font subset tags.

### DIFF
--- a/.claude/recent-fixes/2026-09-10-tokenvalue-presizes-from-a-known-count.md
+++ b/.claude/recent-fixes/2026-09-10-tokenvalue-presizes-from-a-known-count.md
@@ -1,0 +1,55 @@
+# TokenValue presizes from a count it already has
+
+Pure performance. No behaviour change.
+
+## What was wrong
+
+`TokenValue`'s only sequence constructor took `IEnumerable<Token>` and did
+`new List<Token>(tokens)`. That overload presizes **only** when the source is an
+`ICollection<T>`. `IReadOnlyList<T>` is not one — and neither is `TokenValue` itself, which
+implements `IReadOnlyList<Token>` and stops there — so it enumerated the source and grew the
+backing array 4, 8, 16, 32…, reallocating and copying at every step, for a sequence whose length
+was known before the first element was read. It also allocated an interface enumerator to do it.
+
+Converters build one of these per declared value (`Original = new TokenValue(tokens)` appears at a
+dozen sites), and every one of them has `IReadOnlyList<Token>` as its static parameter type. So the
+slow path was not an edge case, it was the only path.
+
+## The change
+
+A second constructor taking `IReadOnlyList<Token>`, which presizes from `Count` and copies by
+index. Overload resolution picks it at every one of those call sites without touching them.
+
+## Evidence
+
+Sampled thread-time profiles of the engine rendering a 26-document corpus — 15 warm-up sweeps past
+the plateau, 4 cores, 60-second window, waiting threads excluded:
+
+| frame | before | after |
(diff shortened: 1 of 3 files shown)


jhaygood86 (approved):
Small, well-scoped change with a solid mutation test. I independently reproduced it rather than trusting the claimed numbers: temporarily removed the new IReadOnlyList<Token> overload, rebuilt clean, and reran BuildingFromAKnownLengthList_GrowsItsArrayOnlyOnce — it failed with "allocated 5,256 bytes, more than 1.5x the 2,560 bytes" exactly matching the PR's claimed figures, then restored the overload and confirmed it passes again. Overload resolution correctly binds every existing IReadOnlyList<Token>-typed call site to the new constructor with zero call-site changes. Rebuilt clean (0 warnings), full net8.0 suite passes (10,576 passed, matching the claimed count exactly). CI is green on all three platforms. Approving.

PR #995, opened 2026-09-10
Ask fontconfig once per generic family, not once per document

dchupp:
`LinuxSystemFontResolver.ResolveGenericFamily` did a full fontconfig round trip on every call — `FcPatternCreate`, `FcPatternAddString`, `FcConfigSubstitute`, `FcDefaultSubstitute`, `FcFontMatch`, then marshalling the family name back out of native memory.

It is called far more often than it looks. `PdfGenerator` holds its `PdfSharpAdapter` as an **instance** field, and that constructor resolves every generic family (`serif`/`sans-serif`/`monospace`/`cursive`/`fantasy`) plus `system-ui`. Callers construct a `PdfGenerator` per document, so rendering N documents made **7N identical native round trips**, every one returning the same answer.

A sampled CPU profile put **2.67% of all non-waiting engine work in `FcConfigSubstitute` alone**, with `FcFontMatch` on top, all under this one method.

## Why caching is safe

The answer cannot change while the process runs. `fcConfig` is a `Lazy<IntPtr>` over `FcInitLoadConfigAndFonts`, so fontconfig's configuration is loaded exactly once and every later query is asked of that same immutable config. A font installed mid-process would not be seen without a reload either way — so this changes no result, only how often the round trip is paid.

The delegate handed to `GetOrAdd` is cached in a static field rather than written inline. A method group converted at the call site allocates a fresh delegate on every call, which would put back a per-call allocation while removing a per-call native round trip — the same trap #977 fixed for the GSUB/GPOS lookup caches.

## Evidence

Sampled thread-time profiles, 26-document corpus, 15 warm-up sweeps past the plateau, 4 cores, 60-second window, waiting threads excluded:

| frame | before | after |
| --- | ---: | ---: |
| **`FcConfigSubstitute`** | **2.67%** | **0.00%** |
| `List<CSS.Token>..ctor(IEnumerable)` | 13.20% | 13.77% |
| `Enumerable.Select().ToArray()` | 5.23% | 5.32% |
| `GraphicsAdapter.DrawString` | 4.65% | 4.65% |

Total work samples in the same window fell **140,082 → 136,329** — the first change in this series where the sampler saw measurably *less* work, rather than the same work redistributed.

**Throughput — sustained, not burst.** 45-second holds at concurrency 4 after a 15-sweep warm-up, interleaved, three reps each:

| | docs/s |
| --- | --- |
| before | 92.97, 93.01, 92.30 |
| after | **95.77, 95.62, 95.18** |

**+2.85%**, winning every rep, with under 1% spread on each side.

Worth saying that the instrument mattered as much as the change here: a burst probe has a 10–15% run-to-run spread and could not have resolved this at all. A warmed, sustained hold can, which is why this PR quotes a throughput number where an earlier one of mine deliberately refused to.

## Tests

`ResolveGenericFamily_OnLinux_AsksFontconfigOncePerFamily` asserts by **reference**, not equality — every uncached resolution marshals a fresh string out of native memory, so two uncached calls can be equal but never the same instance. `Assert.Equal` would pass against unfixed code and prove nothing. It also asserts the first result is non-null before comparing, so it cannot pass vacuously on a host where fontconfig returns null for everything.

Mutation-verified: bypassing the cache fails it.

- Full net8.0 suite: **10,582 passed**, 9 skipped, 0 failed.
- 26-document corpus: **26 pass / 0 review / 0 fail**, byte-identical after normalisation.

(The file is `[ExcludeFromCodeCoverage]` — it is Linux-only P/Invoke, so its lines never execute on the Windows and macOS CI legs.)

### DIFF
--- a/.claude/recent-fixes/2026-09-10-fontconfig-is-asked-once-per-family.md
+++ b/.claude/recent-fixes/2026-09-10-fontconfig-is-asked-once-per-family.md
@@ -1,0 +1,66 @@
+# fontconfig is asked once per generic family, not once per document
+
+Pure performance. No behaviour change.
+
+## What was wrong
+
+`LinuxSystemFontResolver.ResolveGenericFamily` did a full fontconfig round trip on every call —
+`FcPatternCreate`, `FcPatternAddString`, `FcConfigSubstitute`, `FcDefaultSubstitute`, `FcFontMatch`,
+then marshalling the family name back out of native memory.
+
+It was called far more often than it looks. `PdfGenerator` holds its `PdfSharpAdapter` as an
+**instance** field, and that adapter's constructor resolves every generic family
+(`serif`/`sans-serif`/`monospace`/`cursive`/`fantasy`) plus `system-ui`. Callers construct a
+`PdfGenerator` per document, so rendering N documents made **7N identical native round trips**,
+every one of them returning the same answer.
+
+A sampled CPU profile put 2.67% of all non-waiting engine work in `FcConfigSubstitute` alone, with
+`FcFontMatch` on top of that, all under this one method.
+
+## Why caching is safe
+
+The answer cannot change while the process runs. `fcConfig` is a `Lazy<IntPtr>` over
+`FcInitLoadConfigAndFonts`, so fontconfig's configuration is loaded exactly once and every later
+query is asked of that same immutable config. A font installed mid-process would not be seen
+without a reload either way, so the cache changes no result — only how often the round trip is paid.
+
+The delegate handed to `GetOrAdd` is cached in a static field rather than written inline. A method
+group converted at the call site allocates a fresh delegate every call, which would have put back a
+per-call allocation while removing a per-call native round trip — the same trap #977 fixed for the
+GSUB/GPOS lookup caches.
+
+## Evidence
+
+Sampled thread-time profiles, 26-document corpus, 15 warm-up sweeps past the plateau, 4 cores,
+60-second window, waiting threads excluded:
+
+| frame | before | after |
+| --- | ---: | ---: |
+| **`FcConfigSubstitute`** | **2.67%** | **0.00%** |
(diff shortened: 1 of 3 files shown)


jhaygood86 (approved):
Verified this on real Linux, not just by reading the diff. Checked out the branch in WSL (Ubuntu, with a working libfontconfig.so.1), built clean (0 warnings), and ran the new ResolveGenericFamily_OnLinux_AsksFontconfigOncePerFamily test against real fontconfig — passed. Then mutation-tested it myself: bypassed GenericFamilyCache.GetOrAdd to call the uncached resolver directly, rebuilt, and reran — it failed exactly as claimed (Assert.Same failure, two distinct "DejaVu Sans" string instances from two live fontconfig round trips), confirming the reference-equality test is real and would catch a regression. Restored the file and confirmed it passes again.

Full net8.0 suite passed in WSL except for AListItemCostsLittleMoreThanAPlainBlock (an allocation-margin test in AnonymousBoxDefaultingTests, unrelated to this PR's fontconfig caching) — I confirmed this same failure reproduces identically on current main in the same WSL environment, so it's pre-existing environment-specific flakiness (WSL's JIT-tiering/CPU characteristics vs. the GitHub Actions runner), not a regression from this change. The real CI run confirms this: Ubuntu passed cleanly there.

Also rebuilt and ran the full suite on Windows directly (0 warnings, 10,582 passed matching the claimed count exactly) to confirm the cross-platform C# compiles correctly even though the P/Invoke path is Linux-only at runtime. CI is green on all three platforms. The ConcurrentDictionary.GetOrAdd pattern is safe for PdfGenerator's documented per-thread usage model — a concurrent first-access race across threads could invoke the uncached resolver more than once for the same key, but the final stored/returned value is still consistent, and the underlying fontconfig config is immutable for the process lifetime, so no correctness issue results. Approving.

PR #996, opened 2026-09-10
Fix browser-compatible wrapping of inline emoji runs

ChrisVanDijk:
## Summary

- make inline markup transparent to line breaking
- wrap eligible emoji at complete Unicode grapheme boundaries
- preserve modifiers, variation selectors, flags, ZWJ sequences, and punctuation adjacency
- prevent mapped variation selectors from adding an independent glyph advance
- align horizontal, vertical, and min-content wrapping behavior
- update documentation and the emoji showcase

## Validation

- Full net8.0 suite: 10,642 passed, 9 platform-specific skips
- Focused wrapping/shaping suite: 201 passed
- Diff coverage: 99%
- Workspace build successful
- Exact Segoe UI Emoji example matches Edge line distribution


jhaygood86 (changes requested):
This is a well-executed change overall, and I verified the substantive claims independently rather than taking the write-up at face value:

- **Spec compliance is correct**: CSS Text 3 §5.5 states inline box boundaries do not themselves introduce a wrap opportunity, and `HasOrdinaryWrapOpportunityBefore` (`CssLayoutEngine.cs:3596-3627`) implements exactly that.
- **Grapheme/emoji handling checks out against real UAX #29 rules and live Unicode 17.0.0 UCD data** — cross-checked the hand-rolled `_emojiLineBreakRanges` and `ProhibitsLineBreakAfter`/`ProhibitsLineBreakBefore` tables against `LineBreak.txt`/`emoji-data.txt` directly; no mismatches found (including the PR/PO numeric-prefix/postfix tables, all 26+26 entries).
- **Rasterized the Segoe UI Emoji test scenario myself with both PDFium and MuPDF** — pixel-consistent between renderers, line distribution matches the test's own assertions, and VS16 sequences (✌️, 🕊️) render as single glyphs, confirming the "mapped variation selector adds a false glyph advance" bug is genuinely fixed.
- **CJK IVS is not regressed** — confirmed cmap format-14 parsing (the actual IVS mechanism) isn't implemented anywhere in this codebase, so hiding variation-selector advance unconditionally can't touch it.
- Build/CI: 0 warnings, all three platforms green, diff coverage 99% (2 uncovered lines, both trivial defensive guards).

**One real bug, unfixed, needs to block merge**: `HasInterElementWhitespaceBefore` (`src/PeachPDF/Html/Core/Dom/CssLayoutEngine.cs:3737-3739`) is not recursive:

```csharp
private static bool HasInterElementWhitespaceBefore(CssRectWord word) =>
    DomUtils.PrecedingBoxAcrossFirstChildChain(word.OwnerBox)?.Text is { } text
    && HtmlUtils.IsNullOrCollapsibleWhitespace(text);
```

It reads `.Text` only off the *immediate* preceding sibling box. `CssBox.Text` (`CssBox.cs:1018`) is only ever set on a leaf text box — a structural container box's `.Text` is `null`, since its content lives in its children. This is exactly the trap its RI-parity counterpart `CountTrailingRegionalIndicatorsIn` (`CssBox.cs:1475-1492`) correctly avoids by explicitly recursing through `box.Boxes` before falling back to `box.Text` on a leaf.

**Concrete failure**: `<span>A</span><b><i> </i></b><span>B</span>`. The whitespace lives inside `<i>`, nested two levels below the preceding sibling `<b>`. `PrecedingBoxAcrossFirstChildChain` returns `<b>` (whose own `.Text` is `null`), so the whitespace is invisible to this check and "AB" is now treated as one unbreakable token — a real regression versus pre-PR behavior, where every inline boundary was unconditionally a wrap opportunity. I don't see a test in the diff that exercises whitespace nested more than one level inside a non-leaf preceding inline sibling.

Please make `HasInterElementWhitespaceBefore` recurse the same way `CountTrailingRegionalIndicatorsIn` does, and add a regression test for the nested-whitespace pattern above. Everything else here is solid — will take another look once that's in.

(Minor, non-blocking: `.claude/recent-fixes/2026-09-10-inline-boundaries-do-not-create-wrap-opportunities.md` cites a stale "10,621 passing" figure versus the actual ~10,642-10,670; worth a quick fixup whenever convenient.)

jhaygood86 (approved):
The fix mirrors CountTrailingRegionalIndicatorsIn's recursion exactly as requested, correctly handling both arms I was concerned about (a real-content child ending the search, an empty inline being skipped past). I mutation-verified it myself: reverted HasInterElementWhitespaceBefore back to the shallow one-level check, rebuilt, and reran the new Normal_WhitespaceNestedInsideAPrecedingInlineIsAWrapOpportunity test — it failed with the exact predicted symptom (1 line instead of 2, "AB" welded into one unbreakable token), then restored the fix and confirmed it passes again. The new test's four cases (flat/nested/trailing/text) cleanly cover all three logical branches of the new EndsWithCollapsibleWhitespace method. The new invariant file is exactly the kind of record CLAUDE.md asks for — states the measured symptom and pins it to the specific test.

Rebuilt clean (0 warnings), full net8.0 suite passes (10,643 passed — the prior 10,642 plus the one new test, 0 failed). CI is green on all three platforms. Approving.

PR #997, opened 2026-09-10
Draw color-glyph artwork once into a Form XObject

ChrisVanDijk:
Every occurrence of a COLR/CPAL glyph re-emitted its whole decoded paint graph into the
page content stream. Measured with Segoe UI Emoji, 400 occurrences on one page:

| Content | PDF bytes per occurrence |
| --- | --- |
| Latin text | 818 B |
| 😀 plain color glyph | 16 KB |
| ✌️ variation-selector sequence | 42 KB |
| 👨‍👩‍👧 ZWJ sequence | 133 KB |

That is a **53.3 MB single-page PDF**, and it is slow to *display* as well as to write,
because a viewer re-parses those paths once per occurrence too. The artwork is
byte-identical every time — only the placement differs.

Each distinct glyph's artwork now goes into one Form XObject, and each occurrence emits a
placement `cm` plus `Do`.

| Fixture | Before | After |
| --- | --- | --- |
| 400 ZWJ family emoji, one page | 53,295,953 B / 7.3 s | 236,905 B / 1.8 s |
| 400 variation-selector emoji | 16,786,530 B | 139,985 B |
| Two-page report (tables, cards, emoji in running text) | 1,081,412 B | 448,460 B |
| `emoji` showcase | 322,322 B | 212,797 B |
| `font_palette` showcase | 217,104 B | 161,729 B |

## Design

- **Size is deliberately not in the key.** The artwork is drawn at a canonical 100-unit em
  and the placement `cm` scales it by `fontSize / EmSize`, so one stored copy serves every
  size the glyph is used at. This is the one place the color-glyph cache differs from
  `PdfImageTable`, whose embedded pixels genuinely do depend on display size.
- **The key is everything that changes the artwork, and nothing else**: descriptor identity,
  glyph id, CPAL palette index, `font-palette` entry overrides, and the *text colour* — a
  COLR layer or paint can name it through the `0xFFFF` "use foreground" sentinel, and a
  glyph with no colour record is filled in it outright. Overrides are compared by value: a
  resolved palette reaches the backend as a freshly allocated dictionary per `DrawString`,
  so identity comparison would have deduped nothing.
- **Form size comes from a measure pass**, not a guessed box: the same paint is re-run
  accumulating the bounds of every fill it would make. That is exact rather than
  conservative because every COLR v1 leaf fill is bounded by its enclosing glyph clip, and
  v0 layers by their own outlines. A fixed generous box was considered and rejected — it
  needs an arbitrary constant, and a `PaintTransform` can put ink outside any constant.
- **No `/Group` on the form.** It is invoked by a bare `Do`; adding an isolated
  transparency group would *change* compositing, since COLR v1 `PaintComposite` blend modes
  blend against the page backdrop today.
- **The invisible text stays on the page**, one text object per occurrence, because its
  `/ActualText` describes that occurrence's own source sequence (❤ with and without VS16 is
  the canonical case). Recorded as an invariant, since breaking it is silent — the document
  renders pixel-identically and only text extraction is wrong.

## Bug this surfaced: COLR paint alpha was discarded

`ResolveColor` scaled a paint's alpha as though `XColor.A` were a 0–255 byte when it is a
0–1 double, so alpha 0.2 rounded to a fully transparent `0` (and anything above 0.5 to
1/255). Every non-opaque COLR paint was effectively invisible.

Noto Color Emoji's flags outline themselves with a `#1A1A1A` layer at alpha 0.2 — the light
grey border in Noto's own reference rendering — and those borders were missing. What hid
the bug is that its *symptom* depended on graphics-state history rather than on the glyph:
rendered alone, 🇯🇵 produced an opaque near-black border, while the same glyph on a busy
page emitted `/ca 0` and vanished. Moving the artwork into a form changed which way that
fell, which is what exposed it. Fixed here; the border now matches the reference.

## Cost

A distinct color glyph costs about **440 bytes** of form dictionary, object header and xref
entry that an inlined copy did not. The worst case by construction is the `color_emoji`
showcase — a per-paint-feature breakdown where nearly every glyph appears once — at
198,964 → 210,769 bytes (+5.9%). Promoting only on a second occurrence was considered and
rejected: it buys back a rounding error against 16–133 KB per repeat, in exchange for two
code paths.

## Verification

- **Rasterized through both PDFium and MuPDF**, per the repo's paint-verification
  convention. On the two-page report, PDFium's maximum per-channel difference against the
  pre-change output is **1/255** over 24 pixels of 1.17M; MuPDF's is 4–7/255 — antialiasing
  moving under the new coordinate quantization, not geometry. (Quantization actually
  improves: four decimal places at a 100-unit em resolves finer than four places at 12pt.)
- **Showcases: 110 of 113 byte-identical.** The three that move are the colour-font ones —
  two smaller, and `color_emoji` rendering the flag borders that were previously dropped.
- **Cross-page reuse confirmed structurally**: one form object, `/XObject << /Fm0 6 0 R >>`
  in two page resource dictionaries. One form is invoked five times across four different
  font sizes on one page.
- Emoji still come back from text extraction in both engines.
- Suite: 10,699 passed, 0 failed. Diff coverage 97.3%. `dotnet build -t:Rebuild`: 0 warnings.

## Also fixed: a `.gitignore` trap

A bare `*.pdf` pattern (present twice) also matched the **directory**
`src/PeachPDF/PdfSharpCore/Drawing.Pdf` on case-insensitive filesystems. Git refused to
recurse into it, so every new file added there was invisible to `git status` with no way to
re-include it from inside. `!*.pdf/` / `!*.html/` re-include the directories while leaving
the file ignores intact — both new source files in this PR landed in exactly that folder.

## Not done here

The same treatment for SVG (a logo in a running header re-emits its scene graph per page;
`CssImagePainter.PaintSvgLayer` even builds a *fresh* Form XObject per paint). Left for its
own change because it touches the paint path for all SVG content, not just repeated
content. Groundwork is recorded in the dev notes.


jhaygood86 (approved):
Reviewed this thoroughly given the blast radius (core color-glyph painting used by every emoji/color-font PDF). Verified independently rather than trusting the write-up:

- **Cache key correctness**: confirmed it captures descriptor identity, glyph id, palette index, text/foreground color, and font-palette overrides (compared by value — necessary since a resolved palette reaches the backend as a freshly allocated dict per DrawString), while deliberately excluding size (the placement `cm` scales a canonical 100-unit em). Content-stream inspection confirms this works in practice: one Form XObject invoked at two different scales for the same glyph at two font sizes, pixel-correct at both.
- **The "no /Group" compositing claim** checks out — adding an isolated transparency group would change how COLR v1 PaintComposite blend modes composite against the page backdrop, so a bare Do is correct here.
- **Form bounds from a measure pass**: confirmed sound — COLR v1's FillClip requires an enclosing glyph clip, so there's no unbounded-ink case a guessed box would need to cover.
- **The bundled alpha fix**: dual-renderer (PDFium + MuPDF) rasterization confirms the flag border is now correctly visible at alpha 0.2, matching Noto Color Emoji's own reference rendering. Independent investigation traced the exact mechanism (the old ResolveColor formula genuinely computed alpha 0.2→0) and additionally surfaced a related-but-distinct pre-existing bug in PdfGraphicsState.RealizeFillColor's alpha-change check (missing an IsEmpty guard after a q/Save reset) — out of scope for this PR, so I've filed it as a separate follow-up rather than blocking here.
- **Cross-page reuse**: confirmed structurally — one Form XObject defined once, referenced via indirect object from multiple pages' resource dictionaries.
- **The .gitignore fix** is a real, correctly-scoped fix (re-includes the `Drawing.Pdf/` directory that a bare `*.pdf` pattern was case-insensitively swallowing on Windows).
- Diff coverage independently re-verified at 97% (4 lines missing across 148, all narrow defensive branches), well above the 90% gate.

Rebuilt clean myself (0 warnings), ran the full net8.0 suite (10,699 passed, matching the claimed count exactly, 0 failed). CI is green on all three platforms. Approving.

PR #1001, opened 2026-09-11
Fix Charts.css labels: pseudo-elements, flex items, positioned margins

ChrisVanDijk:
The `charts_css` showcase drew every chart's bars correctly and **none** of its labels — no `Q1`–`Q4`
axis labels, no `$42M` bar values, no `JavaScript`/`Python` row labels, and the area/line data labels sat
on the chart baseline instead of at their data points. Then the column chart's primary axis drew twice,
as a black bar above a grey one. Six independent defects behind those symptoms, found one behind the
other.

## 1. A universal pseudo-element rule attached generated content to bare text

`CssData.DoesSelectorMatch`'s pseudo-element arm creates the `::before`/`::after` box as a side effect of
matching, and `DoesSelectorMatch(AllSelector)` is structural, so the universal selector reached every box
— including the anonymous box a raw text node becomes. Charts.css's own
`.charts-css *::before, .charts-css *::after { box-sizing: border-box }` therefore hung two empty pseudo
boxes off every text box in the document.

That left the text box holding **both** its own `Words` and child `Boxes`, a shape
`CssLayoutEngine.FlowBox` does not flow. The word was never positioned by an inline flow again, stayed
flagged `CssRect.AwaitsTheNextFragmentainer`, and `FragmentEmitter` dropped it — so the box laid out at
the right size and painted its background and border, and only the glyphs were missing.

A pseudo-element is generated on an **element**
([Selectors 4 §3.3](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements)).

## 2. `margin-block-start: auto` on a positioned box resolved to 0

Nothing implemented [CSS 2.1 §10.6.4](https://www.w3.org/TR/CSS21/visudet.html#abs-non-replaced-height)'s
"solve the equation for the `auto` margin", so Charts.css's axis labels — `position: absolute; inset: 0`
with a height, an auto start margin and a negative end margin — were drawn across the *top* of every
chart instead of under its primary axis.

Implemented for both absolute and fixed boxes. An auto-height containing block has no final height while
its children lay out, so the box's own epilogue can only compute a provisional answer; the containing
block revises it from its own epilogue, applying the delta.

## 3. CSS 2.1 §9.2.1.1's anonymous block was created inside flex containers

That is a **block container** rule. Applied to a flex container it swallowed an item: the auto-sized
wrapper became the item, so the child's own `width`/`height` sized nothing. Charts.css's area and line
cells are exactly that shape — an inline `td::after` spacer beside a `display: flex` `.data` label — so
every data label collapsed onto the chart baseline.

Skipped for flex/grid containers and replaced with what
[css-flexbox-1 §4](https://www.w3.org/TR/css-flexbox-1/#flex-items) actually asks for: a contiguous
sequence of **two or more** child text runs joined into one item. A lone run is already the anonymous
item, since the parser gives every text node its own tagless box — wrapping it anyway reintroduces the
same defect one level down and splits a label mid-word under `overflow-wrap: anywhere`.

## 4. An unresolvable percentage height collapsed a flex/grid container

Both engines read `GetBoxHeight`'s `null` through `?? 0`. That `null` means
[CSS Box Sizing 4 §5](https://www.w3.org/TR/css-sizing-3/#definite)'s "behaves as automatic", so
`height: 100%` against an auto-height parent became "definite, and it is 0" — and a `column` container
stacked every item at its content origin, drawn on top of each other.

## 5. An absolutely-positioned child inflated its parent's painted extent

With the labels back, the column chart's primary axis drew as two stacked bars. In a browser they are one
line: the black axis is `tr { border-block-end }`, the grey is the last secondary-axis grid line — a
repeating background on the *same* row, whose fifth tile lands in the bottom border strip because
`background-clip` defaults to `border-box`, and the border paints over it.

Instrumenting that row: `decoH=147.857`, `boundsH=131.357`, `marginBottom=16.500`. Layout was right; paint
received the **margin box**. `FragmentEmitter.ExtentOf` grows a box's extent to cover content that
overflows bounds pinned before that content was known — but it counted out-of-flow children too, and the
axis labels are positioned with an auto block-start margin and a negative block-end margin *precisely so
they hang below their row*. They dragged the row's own border down with them. Confirmed by hiding just the
label in the otherwise identical document: `decoH` dropped to 131.357, an exact match.

An out-of-flow box is taken out of the flow entirely (CSS 2.1 §9.3.1) and contributes nothing to its
ancestor's size (§10.6.3). In-flow overflow still extends the box, guarded by its own test.

This also moved the `Q1`–`Q4` labels below the axis, where a browser puts them — the same 16.5pt, and easy
to read as intentional.

## 6. A Form XObject tile's natural size was truncated to whole points

With (5) fixed the two lines were still ~2.6pt apart. `background-size` resolved correctly —
`tileH=32.6518`, `×4 = 130.6071`, exactly the padding box — so the error was in the repeat.
`ImageAdapter.Width`/`Height` returned `XImage.PixelWidth`/`PixelHeight`, which for an `XForm` are `(int)`
casts of its view box. A form has no pixels, and for a *repeated* tile that truncation does not stay
sub-point: 32.6518 → 32, accumulating once per tile until the fourth grid line sat clear of the axis.

Forms now report their exact point size; raster images still report pixels, which the caller converts. The
tiling call site already declared `intrinsicSizeInCssPixels: false` — it was asking for points and being
handed truncated ones. **This path is shared by every tiled Form XObject**, so SVG patterns, masks and
opacity groups were all placing non-integer-sized tiles slightly wrong too.

## Also included

- **A layout-internal `display` on a flex/grid item is blockified** (css-display-3 §2.7). It previously
  stayed layout-internal and no engine laid it out, so a `<tbody style="display: flex">` holding ordinary
  `<tr>`s rendered completely blank.
- **`float` is coerced to `none` on an item** (css-flexbox-1 §4). Collecting a floated child as an item is
  only half the rule: while the box still answered `IsFloated`, the float machinery displaced it as well,
  so a float between two ordinary items dropped onto its own row and left a hole.
  `Floating.Footnote` — genuinely out of flow — is untouched.

## Deliberately out of scope

Blockifying **inline-level** flex items at cascade time, which the spec also requires. It was tried and
regresses replaced elements: a replaced box takes its size from the phantom word carrying its content,
which only reaches the box through inline flow, so as a block-level box it fills its containing block
instead (CSS 2.1 §10.3.4's intrinsic width for block-level replaced content is not implemented here). The
flex/grid engines already lay every item out blockified, so an inline item's own `width`/`height` already
apply — the box tree, not the computed value, was what needed fixing.

## Verification

- `dotnet test --framework net8.0`: **10759 passed**, 9 skipped, 0 failed (plus 96 CLI and 119
  source-generator tests).
- `dotnet build PeachPDF.slnx -t:Rebuild`: **0 warnings**.
- Diff coverage: **117 executable changed lines, 0 uncovered**.
- All 113 showcases regenerated; `charts_css.pdf` rasterized through **both PDFium and MuPDF** (they
  agree) and compared against the browser rendering of the same HTML. Flexbox/grid showcases spot-checked
  for regressions.
- The axis fix was pixel-measured at 5× before and after: the grid now steps 32.65pt exactly and its fifth
  line lands inside the black border, which is what makes the two one line again.
- Every new test was confirmed to **fail** with its own fix reverted. The pseudo-element regression asserts
  on `BoxFragment.Words`, because the box tree looked correct while only the fragment tree lost the text.

## Docs

`docs/html-css-support.md` updated in the same change (flex/grid item generation, indefinite percentage
heights, §10.6.4 auto margins, and the pseudo-element/element rule). Migration notes and recent-fixes
entries are included.


jhaygood86 (changes requested):
This is an exceptionally well-documented PR bundling six real, independently-diagnosed defects plus two more. I reviewed it in depth and independently verified the majority of it:

- **Pseudo-elements attach to elements only** (`CssData.cs`) — correct per Selectors 4 §3.3, and I confirmed the fix by reverting it and observing `UniversalPseudoElementIntegrationTests` fail with the exact claimed symptom (pseudo boxes generated on an anonymous text box).
- **Anonymous-block skip inside flex/grid** (`DomParser.cs`) — correct per css-flexbox-1 §4's item-generation algorithm, traced the index arithmetic in `WrapFlexOrGridTextSequences` directly.
- **Percentage height → indefinite, not 0** — correct per css-sizing-3 §5.
- **Out-of-flow children excluded from extent growth** (`FragmentEmitter.cs`) — correct per CSS 2.1 §9.3.1/§10.6.3, confirmed by reverting the guard and observing `PositionedChildDecorationExtentTests` fail with the exact claimed inflated extent.
- **Form XObject tile natural size truncation** (`ImageAdapter.cs`) — correct; confirmed there are exactly two `ImageAdapter` construction sites in the codebase (raster load, always `XImage`; `CreateTile`, always `XForm`), so the fix's scope is safe, and a second tiled-pattern showcase (`svg.pdf`) rasterizes pixel-identically across PDFium/MuPDF after the change.
- Both "also included" fixes (layout-internal `display` blockification on flex/grid items; `float` coerced to `none` on items, with `Floating.Footnote` correctly untouched) check out against css-display-3 §2.7 and css-flexbox-1 §4 respectively.
- Rasterization confirms the actual reported symptom is fixed: `charts_css.pdf` renders labels correctly, the double-axis-bar defect is gone, and data labels sit at their data points, pixel-identical between PDFium and MuPDF.
- Build/test: 0 warnings, full net8.0 suite passes (10,759 matching the claim), a full Flexbox/Grid/Multicol/Columns regression sweep (919 tests) passes, diff coverage is effectively 100% on the changed lines.

**One real bug, independently confirmed against the actual CSS 2.1 spec text, needs to block merge**: `ResolvePositionedAutoBlockMargins`'s guard (`src/PeachPDF/Html/Core/Dom/CssBox.cs:6309-6311`) only checks that `Top` and `Bottom` are non-auto — it never checks that `Height` is also non-auto. Per CSS 2.1 §10.6.4 (I fetched the spec text directly): the "solve the equation for the auto margin" procedure applies only when `top`, `height`, *and* `bottom` are all specified (non-auto). When `height` is `auto` instead (with `top`/`bottom` set), the spec's rule is the opposite — auto margins must be treated as `0`, and `height` is solved from the equation, filling the containing block exactly. Right now the code applies the margin-solving path unconditionally whenever `top`/`bottom` are set, regardless of whether `height` is auto.

I reproduced this myself with a standalone repro (not part of the PR's test suite, which I confirmed doesn't cover this specific combination — every case in `AbsolutePositioningIntegrationTests` specifies an explicit `height`):

```html
<div id="cb" style="position:relative; height:100pt; margin:0; padding:0; border:0;">
  <div id="abs" style="position:absolute; top:0; bottom:0; margin-top:auto; margin-bottom:0;">x</div>
</div>
```

Expected: `abs` fills the containing block exactly (`Location.Y` = 20, matching `cb`'s top). Actual: `Location.Y` = 105 — 85pt past `cb`'s own bottom edge. This also directly contradicts the PR's own shipped documentation (`docs/html-css-support.md:622` and the new migration note), both of which describe the correct (spec-conformant) behavior that the code doesn't actually implement for this case.

Please add a "Height is non-auto" condition to the guard at `CssBox.cs:6309-6311`, and add a regression test for `height:auto` + `top`/`bottom` set + an auto margin — this exact combination isn't covered by the current `AbsolutePositioningIntegrationTests`. Doesn't affect the Charts.css showcase itself (which uses explicit heights throughout), but it's a real, spec-violating regression risk for anyone else hitting this combination.

Separately, non-blocking: the deliberately-out-of-scope inline-level flex/grid item blockification (mentioned in the PR body and recent-fixes note, with solid technical reasoning for deferring it) is a genuine CSS Display 3 §2.7 spec deviation. Per CLAUDE.md's convention, this needs a tracking GitHub issue and a corresponding `.claude/accepted-gaps/` file — right now it's only documented in prose. Worth adding before or shortly after merge, doesn't need to block this PR.

jhaygood86:
The auto-margin fix looks correct — I mutation-verified it myself (reverted the `HasDefiniteHeight` guard, rebuilt, and confirmed both `AbsoluteBothInsets_AutoHeight_FillsTheContainingBlock_RatherThanSolvingTheMargin` and `AbsoluteBothInsets_IndefinitePercentageHeight_IsTreatedAsAuto` fail with exactly the symptom my original repro found — 105 instead of 20 — then restored it and confirmed both pass again). Good catch extending `HasDefiniteHeight` for the fixed/page-area case too, rather than leaving that reachable-but-untested.

I filed the tracking issue for the inline-level flex/grid blockification gap: [#1002](https://github.com/jhaygood86/PeachPDF/issues/1002). Could you swap the `#TODO` placeholder in `.claude/accepted-gaps/inline-level-flex-and-grid-items-are-not-blockified.md` for `#1002` in your next push? Not blocking merge on that by itself — everything else here is solid and CI is running now.

jhaygood86 (approved):
The requested fix is correct — verified precisely per CSS 2.1 §10.6.4 and mutation-tested by reverting the HasDefiniteHeight guard myself: both new tests fail with the exact original symptom (105 instead of 20), and pass again once restored. Good design reusing the existing HasDefiniteHeight predicate rather than duplicating logic, and correctly extending it for the fixed/page-area case that became newly reachable.

Accepted-gap placeholder is now resolved — thanks for filing #1003 for the inline-level blockification deviation (I'd independently filed #1002 for the same thing before seeing your push; closed it as a duplicate now that #1003 is the one referenced from the accepted-gap file).

Rebuilt clean (0 warnings), full net8.0 suite passes (10,762, 0 failed). CI is green on all three platforms. Approving.

PR #1005, opened 2026-09-11
Update System.IO.Pipelines reference

mcraiha:
Also add comment that the package reference can be removed in future.


jhaygood86:
This PR is obsolete, we already changed it so that the System.IO.Pipelines reference is only on the .NET 8.0 build. The .NET 10 and 11 builds don't include it anymore

PR #1010, opened 2026-09-11
Fix interactive form field geometry, value replacement and attributes

ChrisVanDijk:
Three independent defects in interactive PDF forms, none of which broke a test in the existing ~10,800-test suite. All were found by rendering output and looking at it, and the second and third were confirmed against Adobe Reader on a real device.

## 1. Checkbox and radio controls were not square

An unstyled `<input type="checkbox">` was drawn 9.75pt wide by 16.75pt tall, and a radio button as an ellipse of the same proportions. Setting equal `width` and `height` did not help — the control stayed taller than it was wide by a fixed 7pt.

`CssRectFormField` was the only phantom-word type leaving `CssRect.IsImage` false. Despite the name, that flag is the engine's "atomic, non-text word carrying its owner box's whole replaced geometry" marker, not a statement about pixels — `CssRectShape`, a list-marker bullet, sets it and carries no image either. `CssLineBox.UpdateRectangle` splits the box-model arithmetic across the two axes by it: the horizontal border and padding are added there, the vertical is not, because `CssLayoutEngine.MeasureIntrinsicSize` has already folded it into the word's height. A form field therefore got the vertical inset **twice** and the horizontal **not at all**.

Text fields were wrong the same way, just less visibly: a `width: 160pt` field drew its chrome 160pt wide — the content box — instead of its 165.5pt border box, and the AcroForm widget's `/Rect` was that same wrong rectangle.

Setting the flag also opts a word into the extra trailing space `CssRect.ActualWordSpacing` reserves for replaced content, which a form field must not have — it has no source whitespace to stand in for, and a whole space would open up between a checkbox and the label written immediately after it. The two questions are now separate: `ReservesTrailingSpace` defaults to `IsImage`, and a form field overrides it false.

Squareness additionally required the UA stylesheet to stop applying the text-field `padding: 1pt 2pt` to checkbox and radio controls, which every browser's own UA sheet zeroes there. They take a browser-like margin instead, so a bare `<input type="checkbox"> Label` is no longer flush against its label.

## 2. Editing a field left the old value visible behind the new one

Reported from Adobe Reader with a screenshot: a text field showed the newly typed value, the generated one, and mojibake (`-DQH 'RH`), all overlapping. The mojibake is the tell — those are the generated string's bytes `<002D0044005100480003002700520048>` (glyph IDs into PeachPDF's embedded subset, spelling "Jane Doe") re-rendered through WinAnsi.

The baked `/AP /N` appearance stream drew the value with no `/Tx BMC` … `EMC` marked-content sequence around it. That sequence is the contract ISO 32000-1 §12.7.3.3 defines for variable text, and it is how a reader knows *which part of the appearance to replace*: keep everything before `BMC` and after `EMC` (the border and background), regenerate what lies between. Apache PDFBox's `AppearanceGeneratorHelper.setAppearanceContent` spells out the fallback exactly — with no `BMC` present it **appends**, which is precisely the doubling observed.

`RGraphics` gains a `BeginVariableText()`/`EndVariableText()` pair, emitting the sequence in graphic mode (§14.6 forbids a marked-content sequence straddling a `BT`/`ET` boundary). It is deliberately **not** a reuse of `EndMarkedContent`: that one is called from inside a text object by `ColorGlyphPainter` — the legal `BT BDC … EMC ET` nesting — and forcing graphic mode there would cut the invisible-text run short.

A comb field's cell dividers moved out of the replaceable region with it, or a reader regenerating the value would erase the cells. The sequence is emitted even for an empty value, and even for a field whose padding and border leave no drawable content box — both need a region to replace just as much as a filled one, and are the cases hardest to notice before someone types into them.

Verified on the reporting device: editing, clearing and re-selecting all behave.

## 3. HTML form-control attributes never reached the PDF

The only `/Ff` bits ever set were the two PeachPDF's own CSS extensions drive (`Comb`, `DoNotScroll`), so every HTML attribute was silently dropped. Most consequentially `type="password"`, whose value was drawn legibly onto the page and whose typing the reader was never told to mask.

| HTML | PDF |
|---|---|
| `readonly` | `/Ff` ReadOnly |
| `disabled` | `/Ff` ReadOnly + NoExport — HTML defines a disabled control as neither editable nor submitted |
| `required` | `/Ff` Required |
| `maxlength` | `/MaxLen` |
| `type="password"` | `/Ff` Password, plus asterisks in the drawn appearance |
| `placeholder` | `/TU` tooltip and accessible name |

A comb declaration beats a disagreeing `maxlength`, since a comb field's cell count *is* its `/MaxLen`. A radio group takes the shared flags from *any* member, per the HTML Standard, not only from whichever button is parsed first. Password masking is presentation only — `/V` keeps the real value, because the document asked for the field to be prefilled with it.

A new opt-in property, `-peachpdf-pdf-form-field-placeholder: auto` (initial `none`), additionally draws a field's `placeholder` as muted hint text. It is only safe because of §2 — the hint sits inside the `/Tx` region, so a reader wipes it on the first keystroke instead of leaving it behind the typed text — and it is mixed to an *opaque* colour rather than drawn at half alpha, which would trip `PdfATransparencyGuard` and make merely opting in throw under PDF/A.

## Not a defect

Adobe draws a red outline around every `required` field and keeps it there after it is filled in. This was reported as a bug and is not one: it is the reader's "Required Fields Highlight Color" preference marking the field as required, not a validation state about the current value. A diagnostic PDF with a *prefilled* required field settled it — still red. Documented so it is not "fixed" later.

## Deliberately out of scope

`type="number"`/`type="date"` formatting and `min`/`max`/`step` validation need Acrobat's `/AA` JavaScript convention (`AFNumber_Format`/`AFDate_Format`). That is executable script embedded in the document, unsupported by some readers and rejected by some PDF/A profiles, so it was left out rather than made an invisible default. Those input types continue to work as plain text fields, which is what they already did.

## Verification

- Full suite 10,806 passed / 0 failed (net8.0); source-generator suite 119; CLI suite 96.
- 100% diff coverage (151/151 changed library lines).
- The nine new geometry tests were confirmed to fail 9/9 against the unfixed code, and the trailing-space test to fail when its override is removed.
- The `interactive_pdf_forms` showcase gained an attributes section and a drawn-placeholder section, re-rendered through both PDFium and MuPDF.


jhaygood86 (changes requested):
Reviewed this in depth given the blast radius (interactive AcroForm generation). All three defect fixes check out, verified independently rather than trusting the write-up:

- **Checkbox/radio squareness + text field border-box sizing**: traced the exact mechanism (`MeasureIntrinsicSize` folding vertical inset into word height while `CssLineBox.UpdateRectangle` only added horizontal inset for non-`IsImage` words, so `CssRectFormField`'s `IsImage == false` caused vertical inset to apply twice and horizontal never). Mutation-verified by reverting `IsImage` — 7/10 geometry tests failed with the exact claimed symptoms (9.75pt vs 11.25pt square, 160pt vs 165.5pt border-box width), then confirmed passing again restored.
- **`/Tx BMC ... EMC` variable-text marked content (ISO 32000-1 §12.7.3.3)**: confirmed via Apache PDFBox's `AppearanceGeneratorHelper` source that the "replace from `BMC` to matching `EMC`" convention is real, not invented. Confirmed `BeginVariableText`/`EndVariableText` correctly force graphic mode (closing any open `BT`/`ET` first) per §14.6, and comb-field dividers stay outside the marked region while comb characters stay inside it — matching the PR's own claim about that detail. Mutation-verified by removing the Begin/End calls — 5/7 tests failed as expected (the 2 that still passed are the button-field case, which never emits the sequence).
- **HTML attributes → PDF fields**: byte-level PDF inspection (`pikepdf`) confirms the full `/Ff`/`/MaxLen`/`/TU`/`/V` mapping table exactly, including password masking being presentation-only (`/V` holds the real plaintext) and comb overriding a disagreeing `maxlength`. The "any group member's `required` sets the flag" claim is confirmed against the actual HTML Standard's radio-group semantics.
- Build/test: 0 warnings, full net8.0 suite passes (10,806, exact match), CLI (96) and generator (119) suites pass, diff coverage 100% (151/151), CI green on all three platforms, rasterization (PDFium + MuPDF) confirms genuinely square checkboxes, circular radios, and border-box-matched text field chrome.

**Two things need addressing before merge:**

1. **Documentation bug**: `docs/html-css-support.md:1646` says the placeholder hint is "drawn in the field's own `color` at half opacity." That's factually wrong — the implementation deliberately avoids alpha (it would trip `PdfATransparencyGuard` under PDF/A) and mixes to an opaque color instead, which the PR's own description gets right. `.claude/recent-fixes/2026-09-11-form-control-attributes-and-placeholder.md:44` has the same stale claim, inconsistent with its own lines 87-89. Low severity, but a reader would form the wrong mental model of how this stays PDF/A-safe, and a future contributor could "restore" alpha thinking that was the design. Quick fix, both locations.

2. **Design question on `-peachpdf-pdf-form-field-placeholder: auto`**: why does this need a new proprietary CSS property at all, on top of the standard HTML `placeholder` attribute? Every browser displays `placeholder` text automatically whenever a field is empty — no CSS opt-in required — and printing (the closest analogue to PDF generation) captures that same live-rendered state, placeholder included. Right now, setting `placeholder="Jane Doe"` on an input produces *zero* visible difference in the generated PDF by default (only an invisible `/TU` tooltip) unless the document author also separately opts in with this new property — a real divergence from how every other CSS-driven renderer treats that same unmodified HTML. If there's a deliberate reason for the opt-in (e.g., baking static hint text into an interactive form's default appearance is a materially different decision than a browser's ephemeral, always-live placeholder, or bundling "draw it automatically" together with the marked-content fix felt like too much default-behavior change at once), that reasoning should be stated explicitly in the PR/docs rather than left implicit — and it's worth reconsidering whether matching standard placeholder behavior by default (with this property available for turning it *off*, if anything) is the better fit for a library whose whole pitch is HTML/CSS-as-the-API. Not asserting the current design is wrong, but it needs a real answer before shipping a new permanent property.

Everything else here is excellent work. Happy to re-review once these are addressed.

jhaygood86:
Following up with concrete direction on the placeholder-property question, rather than just posing it as open-ended:

1. **Drop the proprietary `-peachpdf-pdf-form-field-placeholder` property entirely.** A new vendor-prefixed CSS property for something the platform already has a name for isn't the right shape for this library — PeachPDF's whole pitch is that HTML/CSS is the API, and a bespoke property here just adds something every user has to discover and remember, for behavior the web platform already defines.

2. **Drive it off the standard [`placeholder`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/placeholder) attribute instead**, matching how every browser and print pipeline already treats it: shown automatically whenever the field is empty, no opt-in required. The attribute is already parsed and mapped to `/TU` in this PR — this just means the same attribute also drives whether the hint gets drawn, rather than requiring a second, separate signal.

3. **Add `::placeholder` and `:placeholder-shown` selector support as part of this same body of work**, to the extent the PDF form-field model allows:
   - **`::placeholder`** should let a document author style the hint the normal CSS way — `color`, `font-style`, `opacity`, etc. — instead of `FormFieldAppearanceBuilder` hardcoding the muted-opaque-mix color internally. The existing `PdfATransparencyGuard` constraint still has to hold: if an author's `::placeholder` rule specifies an alpha-reduced color under a PDF/A profile, that should either get mixed to opaque the same way the current fallback does, or fail the same conformance check the rest of the pipeline already enforces — not silently ignored.
   - **`:placeholder-shown`** can only be evaluated at generation time, not as a live interactive state — PDF's AcroForm model has no native "currently showing placeholder" concept, which is exactly why this PR has to bake static hint text into the default appearance in the first place. So it resolves the same way the current "should I draw the hint" check already does: matches when the field's value is empty and a `placeholder` is present, at the moment the PDF is generated. That's a real, useful selector even under that constraint — it lets an author style the field itself differently while empty (e.g. a lighter border), not just the hint text.

This also means the marked-content work in this PR isn't wasted if the property goes away — the `/Tx BMC ... EMC` wrapping fix stands on its own regardless of how the hint's styling is driven.

ChrisVanDijk:
Implemented the requested design.
- Removed -peachpdf-pdf-form-field-placeholder from the parser, property registry, computed styles, docs, tests, and showcase.
- Empty placeholder-capable inputs now automatically render their standard HTML placeholder, consistent with the [HTML Standard](https://html.spec.whatwg.org/multipage/input.html#the-placeholder-attribute). /V remains empty and /TU remains populated.
- Added generation-time :placeholder-shown matching in [FormFieldMapper.cs (line 145)](P:/Projects/PeachPDF/src/PeachPDF/Html/Core/Handlers/FormFieldMapper.cs:145).
- Added normal-cascade ::placeholder styling for color, opacity, and font properties through a detached style box in [DomParser.cs (line 859)](P:/Projects/PeachPDF/src/PeachPDF/Html/Core/Parse/DomParser.cs:859).
- Placeholder appearances consume that style in [FormFieldAppearanceBuilder.cs (line 110)](P:/Projects/PeachPDF/src/PeachPDF/Html/Core/Handlers/FormFieldAppearanceBuilder.cs:110).
- The UA default is opaque #7f7f7f, remaining PDF/A-1-safe. Author-specified translucent colors or opacity use the ordinary transparency path and trigger PdfATransparencyGuard under PDF/A-1.
- Corrected both stale documentation claims and documented the static AcroForm limitation in [html-css-support.md (line 947)](P:/Projects/PeachPDF/docs/html-css-support.md:947).
- Kept the /Tx BMC … EMC behavior intact.

Validation:
- Full net8.0 suite: 10,822 passed, 9 skipped, 0 failed.
- Source-generator suite: 119 passed.
- Solution rebuild: 0 warnings, 0 errors.
- Changed production-line coverage: 61/61, 100%.
- git diff --check: clean.

jhaygood86 (changes requested):
This is an excellent redesign — exactly the direction requested. The proprietary property is gone, `placeholder` now drives the hint automatically the way every browser and print pipeline treats it, and `::placeholder`/`:placeholder-shown` are real, working selectors:

- **`::placeholder`** resolves through the genuine CSS cascade into a detached style box (`DomParser.ResolvePlaceholderStyle`) rather than a hardcoded internal color mix — confirmed `color`/`opacity`/font properties all flow through correctly, and the detached box is correctly kept out of the normal layout tree (removed from `Boxes` immediately after construction) since its text only ever gets painted into the AcroForm `/Tx` appearance region.
- **PDF/A safety is preserved the right way**: the new `ApplyOpacity` multiplies the resolved color's alpha by the cascaded `opacity`, so the *default* (no author `::placeholder` rule) case stays fully opaque exactly as before, and an author who explicitly sets a translucent `::placeholder` color now correctly hits `PdfATransparencyGuard` like any other translucent fill — consistent with the rest of the pipeline, not silently ignored, which is exactly the resolution I asked for.
- **`:placeholder-shown`** is correctly restricted to generation-time evaluation (`FormFieldMapper.IsPlaceholderShown`), shared between selector matching and the appearance builder so the two can't disagree, and the docs are honest about it not being live after generation.
- Moving `::placeholder` out of `UnmatchableSelectors` is a real behavior change for existing documents (e.g. Tailwind's preflight, which unconditionally emits a `::placeholder` rule) — confirmed this is intentional and desired, not an oversight, since real cascade-driven styling is the whole point.

**One real spec-compliance bug, needs to block merge**: `FormFieldMapper.IsPlaceholderShown` (`src/PeachPDF/Html/Core/Handlers/FormFieldMapper.cs`) puts `"number"` in the list of types that do **not** support `placeholder`. I checked the HTML Standard directly (https://html.spec.whatwg.org/multipage/input.html#the-placeholder-attribute) — `number` **is** one of the states the `placeholder` attribute applies to, alongside `text`/`search`/`tel`/`url`/`email`/`password`. This isn't just a stray line: `PlaceholderSelectorTests.cs` line 23 explicitly asserts `<input type='number' placeholder='hint'>` should **not** match `:placeholder-shown`, locking the wrong behavior into the test suite. `docs/html-css-support.md`'s own description ("drawn automatically... as HTML defines," linking that exact spec section) makes this a documented contradiction, not just an internal inconsistency.

Please move `"number"` to the `true` branch (alongside `Text`/`search`/`tel`/`url`/`email`/`password`) and flip the expected value on that test case. Everything else here is genuinely well done — happy to take another look once that's fixed.

jhaygood86 (approved):
The number-input fix is exactly right — moved to the true branch alongside the other placeholder-supporting types, matching the HTML Standard, and the test case now asserts the correct behavior. Ran PlaceholderSelectorTests directly (13/13 pass) and the full net8.0 suite (10,822 passed, 0 failed, exact match). Rebuilt clean (0 warnings). CI is green on all three platforms.

This PR ended up in a genuinely better place than where it started: three real defect fixes (checkbox/radio geometry, ISO 32000 marked-content compliance for editable field values, HTML-attribute-to-PDF mapping), a proprietary CSS property replaced with standard placeholder/::placeholder/:placeholder-shown support driven by the real cascade, and a spec-accuracy bug caught before merge. Approving.

Issue #1014, opened 2026-09-12
A block's final line counts its trailing white space in max-content width, widening shrink-to-fit boxes


ChrisVanDijk:
[css-text-3 §4.1.2](https://www.w3.org/TR/css-text-3/#white-space-phase-2) hangs a line's trailing
white space so it does not contribute to the line's width. `CssBox.GetMinMaxSumWords` applies this at
a `<br>` — the one point in its walk where a line is known to have ended — but not at the end of a
block, so a block whose last line ends in white space measures one space wider than it draws.

### Reproduction

```html
<div style="font:16px monospace">AB </div>
```

One character advances 6.5977pt at this font. `GetMinMaxWidth` on that `<div>`:

| | min-content | max-content | glyphs actually drawn |
|---|---|---|---|
| `AB ` (trailing space) | 13.1953 | **19.7930** | 13.1953 |
| `AB` (no trailing space) | 13.1953 | 13.1953 | 13.1953 |

Every shrink-to-fit consumer inherits the extra space. A float is the clearest one:

```html
<div style="font:16px monospace"><div style="float:left">AB </div></div>
```

| | float width |
|---|---|
| PeachPDF, `AB ` | **19.7930pt** |
| PeachPDF, `AB` | 13.1953pt |
| Chromium 148, `AB ` | 17.5938px = 13.1953pt |
| Chromium 148, `AB` | 17.5938px = 13.1953pt |

Chromium gives the same width either way — it hangs the space. PeachPDF is one space wider when the
source has one.

`white-space: pre` / `pre-wrap` are correct already (the space becomes its own `IsSpaces` word and
measures 19.7930 in both engines, which is the intended behaviour there), and `&nbsp;` is correctly
counted. It is only the collapsible trailing space on a block's own final line that is wrong.

### Why the obvious repair is wrong

There used to be a subtraction at the end of that branch:

```csharp
// removed in #1011
if (box.Words.Count > 0 && !box.Words[^1].HasSpaceAfter)
    maxSum -= box.Words[^1].ActualWordSpacing;
```

Its only job was cancelling the phantom word space `CssRect.ActualWordSpacing` gave every `IsImage`
word; once #1011 deleted that term the guard selected exactly the words whose spacing is provably
zero, so it was dead code and was removed.

**Inverting it to `HasSpaceAfter` would be a regression.** `GetMinMaxSumWords` carries *one* running
`maxSum` across a whole subtree, so a given box's last word is not the line's last word whenever a
[...]

Issue #1013, opened 2026-09-12
text-align: justify distributes expansion at every word boundary instead of at justification opportunities


ChrisVanDijk:
`CssLayoutEngine.ApplyJustifyAlignment` counts *every* word on a justified line and spreads the
leftover width after each of them:

```csharp
// src/PeachPDF/Html/Core/Dom/CssLayoutEngine.cs:4730-4746
foreach (var w in lineBox.Words)
{
    textSum += w.Width;
    words += 1d;
}
var spacing = (availWidth - textSum) / words;
...
currentX = word.Right + (overflowsLine ? Math.Max(spacing, word.ActualWordSpacing) : spacing);
```

Per [css-text-3 §7.3](https://www.w3.org/TR/css-text-3/#justification), expansion belongs at
*justification opportunities* — for `text-justify: auto` in a Latin script, the word separators — not
at every inline-box boundary. Two adjacent inline boxes with no white space between them are not a
justification opportunity, so justification must not separate them.

`HasSpaceAfter` is consulted only as a floor for an overflowing line, never as the predicate for
whether a boundary is an opportunity at all.

### Reproduction

```html
<div style="width:200pt; font:16px monospace; text-align:justify">
  A<span id="s">B</span> CD EF GH IJ KL MN OP QR ST UV WX YZ AB CD EF GH IJ
</div>
```

Measured on the first (justified, non-final) line, at `font: 16px monospace` where one character
advances 6.5977pt:

| | `A` right edge | `B` left edge | gap |
|---|---|---|---|
| PeachPDF, `text-align: justify` | 26.5977 | 32.7837 | **6.186pt** |
| PeachPDF, `text-align: left` | 26.5977 | 26.5977 | 0 |
| Chromium 148 | — | flush at `A`'s right edge (8.7969px = 6.5977pt) | 0 |

Chromium's number is from `page.Locator("#s").BoundingBoxAsync()` on the same markup: `#s.x == 8.7969`,
exactly one character advance, i.e. `AB` renders contiguous on the justified line.

The same gap opens on both sides of an `<img>` or inline `<svg>` that has no source white space
around it — so a justified paragraph containing an inline icon still shows the phantom separation
that #1011 removed everywhere else.

### Why it wasn't fixed with #1011

#1011 removed the unconditional trailing word space from `CssRect.ActualWordSpacing`, which is the
*natural* inter-word gap shared by every alignment mode and by intrinsic-width measurement. This
[...]

PR #1015, opened 2026-09-12
Stop replaced content reserving a phantom trailing space

### LINKED ISSUE #1011: Replaced content reserves a phantom trailing space, adding a word gap after every <img> and inline <svg>
## What happens

An atomic inline — ``, inline `<svg>` — reserves one full space's width after itself, whether or not the source has any whitespace there. Text immediately following a replaced element is pushed a space to the right, and a single source space becomes two.

Measured at `font: 16px monospace` (one space advance = 6.5977pt), a 13px image laid out at x=10 and therefore ending at x=19.75:

| markup | text drawn at | gap after the image | browser |
|---|---|---|---|
| `<span>X</span>` | 26.3477 | **6.5977pt** (one space) | 0 |
| ` <span>X</span>` | 32.9453 | **13.1953pt** (two spaces) | one space |
| `<span>Y</span><span>X</span>` (control) | contiguous | 0 | 0 |
| `<svg width=13 height=13>…</svg><span>X</span>` | 26.3477 | **6.5977pt** | 0 |

The text control confirms it is specific to replaced content, not to inline flow generally.

## Expected

No gap beyond whatever whitespace the source actually contains. CSS Text 3 applies `word-spacing` at word-separator characters; an atomic inline is not one, and no UA inserts a space around one.

## Root cause

[`CssRect.ActualWordSpacing`](https://github.com/ChrisVanDijk/PeachPDF/blob/main/src/PeachPDF/Html/Core/Dom/CssRect.cs#L137-L139) adds a second, unconditional word-space term for these words on top of the one a trailing space already contributes:

```csharp
public double ActualWordSpacing =>
    (HasSpaceAfter ? … : 0) +
    (ReservesTrailingSpace ? (FirstLineStyle?.ActualWordSpacing ?? OwnerBox.ActualWordSpacing) : 0);
```

`OwnerBox.ActualWordSpacing` here is the measured width of a space, not the CSS `word-spacing` delta, so the term is a whole space even at the initial `word-spacing: normal`. `CssRect.FullWidth` is what inline flow advances by, so the reservation lands directly in the cursor position.

`ReservesTrailingSpace` defaults to `IsImage`, which is true for `CssRectImage` (also backing `<object>`/`<video>`/`<iframe>`), `CssRectSvg` and `CssRectShape` (list markers). I measured `` and inline `<svg>`; the others share the code path.

## Why this is worth filing now

It surfaced while fixing form-field geometry. `CssRectFormField` had been the only phantom-word type leaving `IsImage` false; setting it — which the box-model arithmetic in `CssLineBox.UpdateRectangle` requires — silently inherited this reservation and put a whole space between a checkbox and its label, with no way for an author to remove it (`margin: 0` does not touch it).

That was worked around by splitting the two questions `IsImage` was conflating: `CssRect.ReservesTrailingSpace` is virtual, defaults to `IsImage`, and `CssRectFormField` overrides it false. **The workaround exists precisely so the form-field fix would not depend on resolving this**, and it is the natural seam for the real fix: making the default `false` — or deleting the term — corrects every replaced element at once.

## Scope / risk

This is layout behaviour every `` in every document has had for a long time, so changing it will shift positions in existing tests and showcase output. That is the work, and it is why it was not bundled into the form-field change: the two are independent, and mixing them would have made the form-field regression impossible to review. Expect to re-baseline image-adjacent layout assertions and re-render showcases.

Worth checking as part of a fix whether the term was originally compensating for something else — a bare `` on a `text-align: right` line currently lands 0.25pt short of the container's right edge, which the reservation does *not* explain (alignment reads the rect, not `FullWidth`) and which may be an unrelated rounding issue.

## Repro

```html
<!DOCTYPE html><html><head><style>body { font: 16px monospace; margin: 0 }</style></head><body>
<div><img src="data:image/png;base64,…" width="13" height="13"><span>X</span></div>
<div><span>Y</span><span>X</span></div>
</body></html>
```

Render and compare the x-origin of each `X` against the element preceding it: the second row is contiguous, the first is one space short of it.

ChrisVanDijk:
CssRect.ActualWordSpacing had two terms: one for a real trailing space (HasSpaceAfter) and a second, unconditional one for any IsImage word. The second is inherited from the original HtmlRenderer code and has no spec basis, so every atomic inline -- , inline <svg>, <object>, <video>, <iframe>, and a shape/image list marker -- pushed whatever followed it one whole word space to the right. With source white space present the two stacked, so " text" got two spaces.

Delete the term and the ReservesTrailingSpace property it was hiding behind. CssRectFormField's override goes with it: the field now gets correct behaviour from the base class instead of opting out of wrong behaviour.

Per css-text-3 4.1.1 white space is the only thing that produces an advance between two adjacent inline-level boxes. Measured at 16px monospace (one space = 6.5977pt), gap after the atomic inline:

  text    6.5977 -> 0          (Chromium: 0)
   text  13.1953 -> 6.5977     (Chromium: one space)

Chromium's figures come from driving the existing Playwright dependency directly, so the target is measured rather than assumed.

Inside list markers now take their gap solely from the UA sheet's li::marker { margin-right: 5px }, uniformly across text, shape and image markers -- the phantom space had been defeating that rule for exactly the two marker kinds whose word happens to be IsImage.

Also removes the now-dead trailing subtraction at the end of CssBox.GetMinMaxSumWords, whose only job was cancelling the phantom term, leaving a note on why it must not simply be inverted, and corrects two comments that described FullWidth as adding a space unconditionally.

Two pre-existing deviations the review pass surfaced around this are recorded as accepted gaps rather than fixed here: justify expanding at every word boundary (#1013) and a block's final line counting its trailing space in max-content width (#1014).

Verified: full suite 10832/0 (net8.0), CLI 96/0, generators 119/0, solution rebuild 0 warnings, 100% coverage on the changed lines, probe rasterized through both PDFium and MuPDF, and all 113 showcases pixel-diffed against a baseline built from unmodified main -- exactly four change, each by the intended one-space shift.

Fixes #1011

### DIFF
--- a/.claude/accepted-gaps/final-line-trailing-space-counted-in-max-content-width.md
+++ b/.claude/accepted-gaps/final-line-trailing-space-counted-in-max-content-width.md
@@ -1,0 +1,22 @@
+# A block's own final line counts its trailing space in max-content width
+
+Tracked as **#1014**. Surfaced by the issue #1011 review; pre-existing.
+
+[css-text-3 §4.1.2](https://www.w3.org/TR/css-text-3/#white-space-phase-2) hangs a line's trailing
+white space, so it does not contribute to the line's width. `CssBox.GetMinMaxSumWords` applies this at
+a `<br>` — the one point in its walk where a line is known to have ended — but not at the end of a
+block, so a block whose last line ends in white space measures one space wider than it draws. Every
+shrink-to-fit consumer of that measurement (a table column, a float, an inline-block) inherits the
+extra space.
+
+The obvious repair is wrong and was checked: the walk carries **one** running `maxSum` across a whole
+subtree, so the last word of any given box is not the last word of the line whenever a sibling's
+content follows it there. In `<span>AB </span><span>CD</span>` that trailing space is an ordinary
+inter-word gap, and subtracting it would undercount the line by a space. Closing this properly means
+knowing where the line really ends during the intrinsic-size walk, which that walk is not structured
+to answer.
+
+Until issue #1011 this was masked in one direction: a `!HasSpaceAfter`-guarded subtraction sat at the
+end of that branch, put there to cancel the phantom word space `CssRect.ActualWordSpacing` used to
+give every `IsImage` word. With that term gone the guard selects only words whose spacing is provably
+zero, so the subtraction was removed — see the note left in its place in `GetMinMaxSumWords`.
--- a/.claude/accepted-gaps/justify-expands-at-every-word-boundary.md
+++ b/.claude/accepted-gaps/justify-expands-at-every-word-boundary.md
@@ -1,0 +1,22 @@
+# `text-align: justify` expands at every word boundary, not at justification opportunities
+
+Tracked as **#1013**. Found during the issue #1011 review; pre-existing, not introduced by it.
+
+`CssLayoutEngine.ApplyJustifyAlignment` (and its vertical counterpart in the column path) sums every
+word's own `Width` on a justified line and spreads `(availWidth - textSum) / wordCount` after **each**
+word, consulting `HasSpaceAfter` only as a floor for an overflowing line. Per
+[css-text-3 §7.3](https://www.w3.org/TR/css-text-3/#justification) expansion belongs at
+*justification opportunities* — for `text-justify: auto` in a Latin script, the word separators — not
+at every inline-box boundary.
+
+Measured at `font: 16px monospace`, `width: 200pt`, on a genuinely justified (non-final) line:
+`A<span>B</span> CD EF …` places `B` 6.186pt after `A`, where left alignment places it flush at
+`A`'s right edge and a browser renders `AB` contiguous. The same gap opens on both sides of an
+`<img>`/inline `<svg>` that has no source white space around it.
+
+Out of scope for issue #1011, which was about the *natural* inter-word gap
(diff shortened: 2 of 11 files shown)


jhaygood86 (approved):
Clean, well-evidenced fix. Verified the core claim independently: css-text-3 §4.1.1 (and every browser's actual behavior) has no mechanism that reserves space after an atomic inline beyond real source white space, so the removed term was genuinely spec-less legacy behavior, not a judgment call. Mutation-verified myself: reintroduced the old unconditional `IsImage` term in `ActualWordSpacing`, rebuilt, and reran `AtomicInlineSpacingIntegrationTests` — 8/9 failed exactly as claimed (the ninth being the all-text control, which is unaffected either way). Restored and confirmed all 9 pass again.

Good process discipline throughout: the two related-but-separate pre-existing deviations found along the way (justify expanding at every word boundary, a block's final line counting trailing space in max-content width) are correctly scoped out into their own tracked issues (#1013, #1014) rather than folded into this fix, and both accepted-gap files reference real, correctly-described issues rather than placeholders.

Rebuilt clean (0 warnings), full net8.0 suite passes (10,832, exact match to the claim). Regenerated and rasterized the two affected showcases (`content_image`, `list_style_image`) — no visual regressions, gaps read as intended. CI is green on all three platforms. Approving.

Issue #1017, opened 2026-09-12
A `white-space: nowrap` block does not start a new line in the intrinsic-width walk, summing sibling lines into one


ChrisVanDijk:
`CssBox.GetMinMaxSumWords` measures max-content by running one total across a subtree and resetting it
wherever a new line begins — "widest line wins". `StartsNewLine` decides where that is, and it refuses
to reset for any box with `white-space: nowrap`:

```csharp
// CssBox.cs:7049
private static bool StartsNewLine(CssBox box) =>
    box.DerivedStyle.ActualDisplay != Keywords.Inline
    && box.DerivedStyle.ActualDisplay != Keywords.TableCell
    && box.WhiteSpace.Value != Whitespace.NoWrap;
```

`white-space` says whether a box's content wraps *within* a line. It says nothing about whether the box
begins one. A block-level box always does, `nowrap` or not ([CSS 2.1
§9.4.1](https://www.w3.org/TR/CSS21/visuren.html#block-formatting)), so a `nowrap` block-level sibling has its
line added to the previous sibling's instead of competing with it.

### Reproduction

One character advances 6.5977pt at `font: 16px monospace`. Float widths, PeachPDF vs Chromium 148:

| markup (inside `font: 16px monospace`) | PeachPDF | Chromium |
|---|---|---|
| `<div>AB CD</div><div>EF</div>` | 32.9883 | 32.9883 |
| `<div>AB CD</div><div style="white-space:nowrap">EF</div>` | **46.1836** | 32.9883 |
| the above with the whole thing under `white-space: nowrap` | **46.1836** | 32.9883 |
| …and a third `<div>GH</div>` sibling | **59.3789** | 32.9883 |

The error is one whole sibling line each time (46.1836 = 32.9883 + 13.1953), so it grows linearly with
the sibling count. Only the *later* sibling needs `nowrap` — with it on the first `<div>` only, the
second one's own reset parks the first line correctly and the result is right.

**`paddingSum` loses its per-line scoping the same way**, since `oldPaddingSum` is saved in the same
branch. Giving both `<div>`s a `border-left: 10pt` gives 66.1836 against Chromium's 42.7383 — both
borders counted instead of the widest line's own. That is the Acid2 `#eyes-a`/`#eyes-b`/`#eyes-c`
regression the `oldPaddingSum` comment describes, reachable again whenever the siblings are `nowrap`.

[...]

PR #1018, opened 2026-09-12
Hang a block's final-line trailing space in intrinsic sizing

### LINKED ISSUE #1014: A block's final line counts its trailing white space in max-content width, widening shrink-to-fit boxes

ChrisVanDijk:
Per css-text-3 4.1.2 the white space that ends a line hangs and does not contribute to the line's
width. CssBox.GetMinMaxSumWords applied this at a <br> only, so a block whose own final line ended in
a collapsible space measured -- and drew every shrink-to-fit box around it -- one space wider than the
glyphs it holds. Source indentation before a closing tag produced it as readily as a literal trailing
space did, so ordinary markup hit it.

The walk carries one running maxSum across a whole subtree, so a given box's last word is not the
line's last word whenever a sibling's content follows it there. That is why the rule cannot be applied
where a box's words run out, and it is what the issue concluded the walk was not structured to answer.
It already answers it in one place, the <br> branch, and the missing half is that the walk has a
second line terminator: the block-boundary epilogue, where a box that opened a line of its own is
closing it. Three box kinds never reach that epilogue because StartsNewLine excludes them -- a table
cell, a white-space: nowrap block, and an inline box measured directly -- and GetMinMaxWidth covers
all three with one subtraction after the walk, a no-op wherever the rule already ran.

What makes the subtraction safe is narrower than it first looks. "maxSum at the epilogue is this box's
last line only" is false and the review pass falsified it. The real invariant is that trailingSpace is
non-zero only while the most recently measured word is still the tail of maxSum: every path that puts
something else on the line after that word clears it, and the <br> branch and the block reset clear it
when the line ends. Two paths did not uphold that and now do -- a flex row's item total, and the
explicit-width fold when the declared length raises the line. Both are reachable only under
white-space: nowrap, which is why a first round missed them.

CssLayoutEngine.GetBoxWidth counted the same space independently, through box.Words.Sum(x =>
x.FullWidth), and GetFitContentWidth folds that through GetLargestChildWidth into every shrink-to-fit
box's used width. Fixing the walk alone therefore left the issue's own float repro unchanged at
19.7930pt. The last word's gap ends the box's content and hangs too. This half was found by measuring,
not by reading: fit came back 19.7930 for "AB <br>CD" while max-content was a correct 13.1953, which is
only possible if something other than the walk decides the width.

Measured at 16px monospace (one space = 6.5977pt), float width:

  AB                                        13.1953 -> 13.1953   (Chromium: 13.1953)
  AB<space>                                 19.7930 -> 13.1953   (Chromium: 13.1953)
  AB&nbsp;                                  19.7930 -> 19.7930   (Chromium: 19.7930)
  <span>AB </span><span>CD</span>           32.9883 -> 32.9883   (Chromium: 32.9883)
  <span>AB </span><span>CD </span>          39.5859 -> 32.9883   (Chromium: 32.9883)
  AB <br>CD                                 19.7930 -> 13.1953   (Chromium: 13.1953)
  <div>AB </div><div>CD</div>               19.7930 -> 13.1953   (Chromium: 13.1953)
  white-space: pre / pre-wrap, "AB "        19.7930 -> 19.7930   (Chromium: 19.7930)

The unchanged rows are the ones that say the fix did not over-reach: a space between two inlines is a
real inter-word gap, &nbsp; never hangs, and a preserved space is its own word whose ActualWordSpacing
is zero, so neither subtraction can reach it. The two invariant repairs were measured the same way
under nowrap: an inline-flex row after a space goes 26.3906 -> 32.9883 (Chromium 32.9883) and an
inline-block with width: 50pt goes 63.1953 -> 69.7930 (Chromium 69.7852, the delta being Chromium's own
sub-pixel rounding of the declared length). Chromium's figures throughout come from driving the
existing Playwright dependency directly, so the target is measured rather than assumed.

Eleven tests are added to IntrinsicWidthWalkTests: three confirmed failing against unmodified main,
three more confirmed failing against the first round of this fix, and five that must pass either way,
guarding against over-subtraction.

One pre-existing deviation the review pass surfaced around this is recorded as an accepted gap rather
than fixed here: a white-space: nowrap block does not start a new line in the intrinsic walk, so
sibling lines are summed (#1017). The note carries the measurement showing that deleting the nowrap
clause on its own makes two inline-level cases worse, because StartsNewLine is wrong in a second way
that clause masks. This change is unaffected either way and strictly improves that shape.

Verified: full suite 10844/0 (net8.0), solution rebuild 0 warnings, 100% line and branch coverage on
every changed line including both arms of the new explicit-width guard, and all 113 showcases
rasterized through PDFium and pixel-diffed against a baseline built from unmodified main -- zero pages
differ. Byte-comparing the PDFs is useless here, since all 113 differ on embedded timestamps alone.

The accepted gap this closes is deleted, and a migration note is filed for the next release-notes pass:
a float, inline-block, auto table column or flex item whose content ends in white space is now up to
one space narrower, and content beside it may shift in by the same amount.

Fixes #1014


jhaygood86 (approved):
Reviewed in depth given the blast radius (intrinsic-width measurement feeds every shrink-to-fit box — floats, inline-blocks, auto table columns, flex items). Independently verified this thoroughly:

- Traced the core invariant ("trailingSpace is non-zero only while the most recently measured word is still the tail of maxSum") through every write site in GetMinMaxSumWords — holds, including the subtle case of an empty inline's margin correctly NOT clearing a still-hanging space.
- Confirmed StartsNewLine's three exclusions (table cell, nowrap block, inline box measured directly) never reach the new block-boundary epilogue, and GetMinMaxWidth's outer subtraction correctly covers exactly those cases as a no-op elsewhere.
- Independently re-measured Chromium directly via Playwright for 14 rows spanning the whole claimed table (floats, &nbsp;, multi-span, nowrap flex/inline-block/inline-table/margin cases) — every value matched the PR's table exactly, including the one I was most skeptical of (a hanging space before an empty margin box).
- Confirmed the GetBoxWidth double-count is real and the fix is algebraically correct — reproduced the exact 19.793→13.195pt numbers the PR quotes by reverting just that half.
- Confirmed issue #1017 is a real, accurately-described, deliberately-deferred gap backed by its own measured table, not a shortcut.
- Mutation-tested the full revert (exactly 3/11 new tests fail, matching main's claimed behavior) and the restored fix (all 17 expanded cases pass). One narrative-only nit: the PR's claim of "3 more failing at an intermediate walk-only stage" reproduces as 1, not 3, when reconstructed from the final diff — doesn't affect the shipped code's correctness, just imprecise phrasing in the write-up, not blocking.
- Rebuilt clean (0 warnings), full net8.0 suite passes (10,844, exact match), all 113 showcases rasterized and pixel-diffed against unmodified main with zero differences. CI is green on all three platforms.

Approving.

Issue #1021, opened 2026-09-12
text-align: justify justifies the last line before a forced break


ChrisVanDijk:
[css-text-3 §6.1](https://www.w3.org/TR/css-text-3/#text-align-property) defines `justify` as:

> Text is justified according to the method specified by the text-justify property, in order to exactly fill the line box. **Unless otherwise specified by `text-align-last`, the last line before a forced line break is start-aligned.**

PeachPDF exempts only the last line of the **block**, not the last line before a forced break. `CssLayoutEngine.ApplyJustifyAlignment`'s only exemption is:

```csharp
if (blockFinished && lineBox.Equals(lineBox.OwnerBox.LineBoxes[^1]))
    return;
```

so a line ending at a `<br>` (or a preserved newline under `white-space: pre-line`/`pre-wrap`) is stretched to the full measure where a browser leaves it ragged.

### Reproduction

```html
<div style="text-align: justify; width: 200pt; font: 16px monospace">
  AA BB CC DD EE FF GG<br>next paragraph
</div>
```

The first line is stretched so `GG` ends flush against the 200pt measure. Chromium leaves it at its natural ragged width and starts `next paragraph` on the following line.

`<br>`-separated address blocks, verse, and signature blocks inside an otherwise justified column are the shapes where this shows up in real documents.

### Why it isn't a one-line fix

Two things make it more than adding a second condition to that guard:

- **The forced-break marker lives on the *following* line, not the one that needs exempting.** `FlowBox` wraps *before* placing the synthetic `"\n"` word, so `CssLineBox.FollowsForcedBreak` is set on the line the break starts, not the line it ends. Answering "is this the line before a forced break" means looking ahead one line — cheap only if the index is threaded down from `FinalizeLineBoxes`, which already iterates by index. An `IndexOf` per line would make finalization quadratic in a long block.
- **A forced break landing exactly on a fragmentainer boundary cannot see its successor at all.** The pass that stops at the break finalizes its lines with `blockFinished: false` and the next line does not exist yet; the resumed pass creates it, by which time the previous line has already been aligned. That is the same shape the `blockFinished` parameter exists to handle, and it needs the same kind of carried-over signal — `InlineBreakToken.FollowsForcedBreak` already carries the mirror image of it for `text-indent: each-line`.

### Related

[...]

Issue #840, opened 2026-08-25
ApplyCenterAlignment/ApplyJustifyAlignment share ApplyRightAlignment's pre-#797-style overflow-guard bug


jhaygood86:
## What's wrong

`CssLayoutEngine.ApplyCenterAlignment` and `ApplyJustifyAlignment` (src/PeachPDF/Html/Core/Dom/CssLayoutEngine.cs) have the same one-directional overflow guard `ApplyRightAlignment` had before its recent fix: they only apply their alignment shift when the computed `diff` is positive, silently leaving an overflowing line exactly where natural (always left-to-right-flowing) layout placed it instead of actively centering/justifying it around the overflow.

`ApplyRightAlignment`'s analogous guard was changed from `if (!(diff > 0)) return;` to `if (!(Math.Abs(diff) > 0)) return;`, mirroring the earlier fix already shipped for `ApplyVerticalFlushAlignment` (the vertical-writing-mode equivalent). The reasoning: a nonzero `diff` in either direction is, by construction, exactly the shift needed to flush/center/justify the line — discarding a negative one leaves an overflowing line un-shifted, spilling off the wrong physical edge instead of the correct one.

## Spec

CSS Text Module Level 3 (`text-align`, `text-align-last`) — an overflowing `text-align: center` line should overflow symmetrically past both edges once centered; an overflowing `text-align: justify` line's own alignment behavior for the final/only line should still center/flush per its own semantics rather than silently reverting to un-aligned natural flow.

## Impact

`text-align: center` or `text-align: justify` combined with `white-space: nowrap` (or a single unbreakable token) and content wider than the container renders left-aligned/un-shifted instead of centered/justified, on both LTR and RTL content. This also affects `text-overflow: ellipsis` correctness for `text-align: center`/`justify` content (see #694's own fix, scoped only to `text-align: right`/RTL-default).

## Out of scope of #694

Fixing `ApplyRightAlignment`'s guard (and the identical horizontal-RTL `text-overflow` blocker it caused) was in scope for #694; generalizing the same fix to `ApplyCenterAlignment`/`ApplyJustifyAlignment` was deliberately left out to avoid scope creep, and is tracked here instead.

Issue #797, opened 2026-08-21
RTL auto-height inline-only vertical content anchors words against the provisional page-height bottom, not the final settled one


jhaygood86:
Part of the real vertical `writing-mode: vertical-rl`/`vertical-lr` layout work tracked in #547, surfaced while implementing #768 (the bug itself is pre-existing, from the earlier #761/#778-era `CreateVerticalLineBoxes` work, and is independent of any of #768's five features — it reproduces with plain, unstyled text and no alignment/bidi/hyphenation/float/positioning involved at all).

## The bug

For a vertical box (`writing-mode: vertical-rl`/`vertical-lr`) with `direction: rtl` and an **auto** (not explicit) `height`, whose content is inline-only (dispatched via `CssLayoutEngine.CreateVerticalLineBoxes`), words are placed using the method's own placement-time `frame` — built against `clientTop + wrapLimit`, where `wrapLimit` for an auto-height box is a large page-height fallback (`CreateVerticalLineBoxes`'s own documented reason: "an auto-height box near the bottom of a page must not self-limit to a sliver of remaining space"). Under `direction: rtl`, `WritingModeFrame.InlineStartIsBottom` is true, so every word's physical `Top` is anchored against that *provisional* (large) bottom edge, not the box's own true content-driven size.

After the word-placement loop, the auto-height settle step correctly shrinks `blockBox.ActualBottom` down to the real content-driven height — but nothing re-anchors the already-placed words to the new, much smaller bottom edge. The result: for an RTL auto-height vertical box, every word ends up positioned far outside (below) the box's own final `ClientBottom`, rather than anchored to it.

## Reproduction

```html
<div style="writing-mode: vertical-rl; direction: rtl; width: 200px">
  Some plain text with no other styling.
</div>
```

No `text-align`, no bidi content beyond ordinary RTL, no hyphenation, no floats, no `position` — the box's own final height is small (content-driven), but the words themselves paint far past `ClientBottom`.

## Likely fix shape

[...]

Issue #956, opened 2026-09-08
line-height: normal is a flat 1.2x, but the spec makes it a function of the font's metrics


dchupp:
`line-height: normal` resolves to a flat `1.2 × font-size` for every font. CSS 2.1 §10.8.1 makes it a function of the font's own metrics, and browsers implement it that way.

Raising this as an issue rather than sending a PR, because `LineHeightTypedStorageTests.NormalKeyword_ActualLineHeight_ResolvesToOnePointTwoTimesFontSize` pins the current value deliberately — so changing it is a behaviour decision that is yours, not a defect fix I should just make.

## The spec

> **normal** — Tells user agents to set the used value to a "reasonable" value based on the font of the element.
> — [CSS 2.1 §10.8.1](https://www.w3.org/TR/CSS21/visudet.html#propdef-line-height)

"Based on the font of the element" is the operative phrase. Browsers resolve it from the font's own ascent, descent and line gap — the `hhea` triple, or the OS/2 typo triple when the font sets `USE_TYPO_METRICS`.

## Measured

Same document (`font-size: 20pt; line-height: normal`), four families, baseline-to-baseline distance between two consecutive lines. Chrome 152 headless, read with `pdftotext -bbox`; PeachPDF via `CssBox.ActualLineHeight`:

| font | Chrome 152 | PeachPDF | difference |
| --- | --- | --- | --- |
| Liberation Serif | 23.25pt (1.162×) | 24.00pt (1.200×) | +3.2% |
| Liberation Sans | 23.25pt (1.162×) | 24.00pt (1.200×) | +3.2% |
| DejaVu Sans | 23.25pt (1.162×) | 24.00pt (1.200×) | +3.2% |
| **FreeSans** | **21.75pt (1.087×)** | 24.00pt (1.200×) | **+10.4%** |

The flat constant is never right, and how wrong it is depends on the font — 3% for the Liberation family, over 10% for FreeSans. That last one matters more than it looks: FreeSans is what `fc-match sans-serif` gives on a machine with the GNU FreeFont set installed, so it is not an exotic case.

## Why it is worth fixing rather than living with

It is a **multiplier on every line of body text**, so it compounds down the page rather than cancelling out. A 10% taller line over a 45-line page is four and a half lines of drift, which is the difference between a table fitting on one page and spilling onto two — and between a document paginating the way its author previewed it in a browser and not.

[...]

PR #1022, opened 2026-09-12
Distribute text-align: justify expansion at justification opportunities

### LINKED ISSUE #1013: text-align: justify distributes expansion at every word boundary instead of at justification opportunities

ChrisVanDijk:
Fixes #1013

`CssLayoutEngine.ApplyJustifyAlignment` re-walked a justified line from its start edge, placing each word at a running cursor and adding a flat `(availWidth - textSum) / wordCount` after every one. Three defects followed from that single shape:

- Expansion landed at **every word boundary** rather than at [css-text-3 §6.4.5](https://www.w3.org/TR/css-text-3/#justify-algos)'s justification opportunities, so `A<span>B</span>` — and text beside an `` or inline `<svg>` — opened a 6.186pt gap the source has no white space for. This is the filed bug.
- Dividing by the **word** count rather than the **gap** count left every gap one share too narrow, and the unconditional last-word flush then dumped the shortfall into the final gap. Every justified line the library has ever produced ended with a gap roughly twice the width of the others.
- Re-deriving positions discarded every natural advance the flow had already computed.

All three go away by distributing over the opportunity count and shifting each word by the expansion accumulated so far — the progressive-shift approach `ApplyLeaderFill` already uses. Keeping the flow's own advances is what makes §6.4.1's "space distributed by justification is *in addition to* the spacing defined by the letter-spacing or word-spacing properties" true by construction, and it preserves a float-narrowed line start and an inline box's own padding for free.

## The opportunity model

A word separator, or the boundary between a block-script character and any other one — §6.4.5's minimum requirements for `text-justify: auto`, the only method PeachPDF implements. Block scripts are approximated by `CommonUtils.IsAsianCharacter`, deliberately the same predicate `CssBox.ParseToWords` uses to split CJK text one character per word, so the opportunities recognized are exactly the word boundaries that split produces and no others. A hyphen or `overflow-wrap` split is not one.

**The trap worth knowing about:** a third source of word separators belongs to no word at all. `ParseToWords` emits no word for a collapsible-whitespace-only text node, so in `<span>AA</span> <span>BB</span>` — about as ordinary as markup gets — neither `AA` nor `BB` carries a flag, and `FlowBox` instead advances the cursor for the whitespace-only box itself. The obvious `previous.HasSpaceAfter || word.HasSpaceBefore` test would have silently stopped that markup justifying. Both flows now record `CssRect.PrecededByWordSeparator` as they position each word, assigned rather than accumulated so a repeated layout pass re-derives it.

## Behavior changes

Two shapes deliberately become no-ops, both confirmed against Chromium — see `.claude/migration-notes/2026-09-12-justify-expands-only-at-justification-opportunities.md`:

- A line with **no justification opportunity at all** (typically one long unbreakable word) is §6.4.3's *unexpandable text*: it aligns as `text-align-last`, whose initial `auto` under `justify` is start. It used to be flushed to the end edge.
- A line **already too long for its measure** is start-aligned by §6.1 too ("any content that doesn't fit overflows the line box's end edge").

The overflow-floor machinery from #840/#843 goes with them rather than being ported: the multi-word case #840 fixed lands in the identical position, because its floor-at-natural-spacing walk reproduced natural placement anyway. Only the two *single*-word overflow tests changed, and they changed to what Chromium does.

`ApplyVerticalJustifyAlignment` takes the same model along the inline axis, plus an anchor to the column's inline-start edge that horizontal must not have: an auto-height vertical box lays out against a placeholder far edge and only learns the real one afterwards (#797).

## Verification

Every expected number was measured in Chromium first, through the repo's existing Playwright dependency, at `font: 16px monospace` in a 200pt block:

| fixture | Chromium | before | after |
|---|---|---|---|
| `A<span>B</span>` gap | 0 | 6.186pt | 0 |
| the nine inter-word gaps on that line | all 10.094px | eight at 6.186pt, last at 12.372pt | all 7.561pt |
| lone overflowing word, non-last line | `x = 0` | flushed to the end edge | start edge |
| `一二AB三四 CD 五六…` | 0.297px at *both* a CJK boundary and on top of a space | — | 0.241pt at both |

That last row settled the design: Chromium gives a word separator and a CJK letter boundary the **same** expansion, which is §6.4.1's equal-priority rule and why a single count and a single share are enough.

All 114 showcases were re-rendered and pixel-diffed against a baseline built from stashed code: **exactly 11 changed, and all 11 declare `text-align: justify`.** Both defects are plainly visible in that output — `paged_media_horizontal_reflow`'s body text ended every line with a doubled gap (`sed  do`, `ad  minim`, `ea  commodo`), and `hyphenation_de_de`'s single-word lines sat flushed right in a narrow column, reading as a stray indent.

- Full suite (`--framework net8.0`): 11 036 passed / 0 failed / 9 skipped
- `diff-cover` against `main`: 100% on all 55 changed lines
- `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings, 0 errors
- New `JustificationOpportunityTests` (10 tests), confirmed to fail against the unfixed library before being kept

Closes the `justify-expands-at-every-word-boundary` accepted gap.

## Deliberately out of scope

Reading css-text-3 in full turned up a separate, pre-existing deviation: **the last line before a forced `<br>` break is still justified**, where §6.1 start-aligns it. It needs a look-ahead of one line (the break marker lands on the *following* line) and a carried-over signal for a break on a fragmentainer boundary, so it is filed as #1021 and recorded in `.claude/accepted-gaps/justify-still-justifies-the-line-before-a-forced-break.md`.


jhaygood86:
CI failed on macos-latest only (Ubuntu and Windows pass) — I pulled the actual failure logs rather than assuming a flake:

```
JustificationOpportunityTests.EveryWordSeparatorOnAJustifiedLine_GetsTheSameExpansion [FAIL]
  each separator gap must be wider than the natural space it expands from (gap=4.442)

JustificationOpportunityTests.CjkAndWordSeparatorsOnOneLine_ShareOneExpansionAmount [FAIL]
  expected the separator gap to also carry its natural space (separator=3.533, boundary=0.199)
```

This is a real, deterministic bug in the new test fixtures, not CI flakiness. In both cases, the *consistency* assertion right before the failing one passed (`Assert.All(separatorGaps, gap => Assert.Equal(separatorGaps[0], gap, 3))` succeeded — every gap really is uniform, so the justify algorithm distributed correctly) — it's the *absolute* point-value threshold immediately after (`gap > 6.6`, `naturalSpace > 6`) that failed. That's the signature of a font-metrics difference, not an algorithm bug: `font:16px monospace` and the bare CJK text in the third test resolve to whatever "monospace"/CJK fallback font macOS's CI runner happens to have installed, which apparently has narrower glyphs than whatever Windows/Ubuntu resolve to.

This is a known, previously-fixed trap in this exact codebase — `BundledFonts.cs`'s `FontFaceRule` helper exists specifically for it, and its doc comment cites the prior incident directly: *"an HTML-layer (LayoutHarness) test whose fixture geometry is calibrated against a specific font's real metrics ... must not depend on whatever font the host OS happens to resolve a bare font-family keyword to (issue #956's own line-height:normal tests hit exactly this: identical fixtures passed on Windows and failed on Linux/macOS CI runners purely because they resolved to different default system fonts with different real metrics)."*

`JustificationOpportunityTests.cs`'s fixtures should pin a real font the same way — `BundledFonts.FontFaceRule(BundledFonts.Ttf, "TestFont", "font/truetype")` (or `Otf`) injected via an inline `@font-face` for the Latin/monospace-width assertions, and `BundledFonts.Cjk` for the CJK-mixed test (`CjkAndWordSeparatorsOnOneLine_ShareOneExpansionAmount`) — rather than the bare `font:16px monospace` keyword and unpinned CJK text. Not a defect in the justify-algorithm fix itself (the underlying logic looks correct — gaps are uniform), just in the two tests' font pinning. This needs to be fixed before merge; I'm continuing a fuller review of the rest of the PR separately.

jhaygood86 (changes requested):
Reviewed this in depth given the blast radius (core justify-alignment, both horizontal and vertical paths). Independently verified nearly everything:

- **The opportunity-count vs. word-count algebra**: mutation-reverted the layout logic and reproduced the exact pre-fix numbers (8 gaps at 6.186pt, one at 12.372pt ≈ 2×), then restored and confirmed the fixed gap (7.561pt), matching the PR's claim to 4 significant figures.
- **Both Chromium-matching behavior changes** (lone overflowing word start-aligns; overflowing multi-word line stays start-aligned with natural spacing) independently reproduced via a fresh Playwright build, and the underlying spec citations (§6.1, text-align-last's initial value under justify) verified against the actual fetched spec text.
- **The whitespace-only-text-node trap** traced directly in `CssBox.AppendWordsFromText`/`FlowBox` — confirmed a collapsible-whitespace-only text node genuinely produces no word, and the fix's `PendingWordSeparator` hook is the only place that catches it. (One honestly-disclosed, pre-existing gap found while tracing further: vertical writing mode's flat word collection doesn't apply this same handling — not a regression, not claimed fixed here.)
- **Accepted gap #1021** is real, open, and accurately described; leaving it out of scope is reasonable given `text-align-last`/`text-align-all` aren't implemented at all.
- **The overflow-floor removal (#840/#843)** is algebraically a no-op for the multi-word case — confirmed the old `Math.Max(spacing, word.ActualWordSpacing)` branch always reduced to natural placement when overflowing, so nothing changes there.
- **Mutation testing**: reverting the layout fix fails 9/10 of the new `JustificationOpportunityTests`, restoring makes all 35 relevant tests (justification + overflow + vertical-text-align) pass.
- Full showcase corpus (114) pixel-diffed against a true pre-PR baseline: exactly 11 changed, and cross-checking against which showcases actually declare `text-align: justify` gives the identical 11 — a clean 1:1 correspondence.
- Build: 0 warnings, full suite passes across all three TFMs (net8.0/net10.0/net11.0, not just the one claimed), diff coverage 100%.

**Blocking: CI is currently red on macos-latest**, and I already flagged the root cause in a comment above — this isn't a flake. Two of the ten new tests (`EveryWordSeparatorOnAJustifiedLine_GetsTheSameExpansion`, `CjkAndWordSeparatorsOnOneLine_ShareOneExpansionAmount`) hard-code `font:16px monospace` and rely on unpinned CJK-fallback resolution, then assert absolute point thresholds (`gap > 6.6`, `naturalSpace > 6`). The *uniformity* assertions immediately before those thresholds pass — the algorithm distributes correctly — but macOS's CoreText resolves `monospace`/CJK fallback to narrower glyphs than Windows/Ubuntu, so the hardcoded absolute thresholds don't hold. This repo already has dedicated infrastructure for exactly this trap (`BundledFonts.FontFaceRule`, built after issue #956 hit the identical problem) — please pin these two tests to a bundled font (`BundledFonts.Ttf`/`Otf` for the Latin case, `BundledFonts.Cjk` for the CJK-mixed case) rather than a bare `font-family` keyword.

Minor, non-blocking: the css-text-3 §6.4.5 restatement (in `IsJustificationOpportunity`'s doc comment and `docs/html-css-support.md`) covers block scripts but omits the spec's separate "clustered scripts" bullet. Harmless in practice since PeachPDF has no clustered-script line-breaking at all, so there's nothing to test yet — just worth a one-line caveat if this area gets touched again, not something to fix now.

Everything else here is excellent, thoroughly evidenced work. Happy to take another look once the macOS test fixtures are pinned.

jhaygood86 (approved):
The font-pinning fix is excellent — it goes beyond just pinning the two failing tests: every fixture in the file now embeds a bundled font, and rather than swapping in a new hardcoded absolute threshold, the natural-gap assertions are now derived dynamically by laying the same document out text-align:left and reading the real gap back, which eliminates this entire class of platform-font-metric fragility rather than just patching the two symptoms. Also caught and clearly documented a genuinely new, narrow limitation while rebuilding the CJK fixtures (ParseToWords merging a Latin run into the following ideograph, so the Latin-to-CJK boundary is invisible in that one direction) rather than hiding it, and folded in the css-text-3 §6.4.5 clustered-script citation I flagged as a minor doc nit.

Verified locally: rebuilt clean (0 warnings), JustificationOpportunityTests 10/10 pass, full net8.0 suite passes (11,036, exact match — one unrelated form-field content-stream-ordering test flaked once under full-suite parallel load and passed cleanly on a second run, consistent with this repo's known order-dependent test flakiness, unrelated to this PR's changes). CI is now green on all three platforms, including macOS, which was the one failing before this fix. Approving.

Issue #1027, opened 2026-09-13
text-align is a longhand, not a shorthand over text-align-all / text-align-last


ChrisVanDijk:
### Summary

[css-text-3 §6.1](https://www.w3.org/TR/css-text-3/#text-align-property) defines `text-align` as a
**shorthand** for `text-align-all` (§6.2) and `text-align-last` (§6.3):

> Setting `text-align` to a value other than `match-parent` or `justify-all` sets `text-align-all`
> to that value and `text-align-last` to `auto`. `justify-all` sets both to `justify`.

PeachPDF instead has two independent inherited longhands — `text-align` (playing the role of
`text-align-all`) and `text-align-last`. `text-align-all`, `justify-all` and `match-parent` are not
recognized at all.

Deliberately left out of scope by the `text-align-last` work (#1021); this issue tracks the
deviation.

### What the deviation actually is

The one reachable difference is the shorthand's **reset**. Given

```html
<div style="text-align-last: justify">
  <p style="text-align: center">…</p>
</div>
```

the spec makes `p`'s last line **centred**, because `text-align: center` on `p` reset
`text-align-last` to `auto`. PeachPDF **justifies** it, because the inherited `text-align-last:
justify` is never reset. Declaration order within a single block matters to the spec for the same
reason, and does not here.

### Why it was left out of scope

- **No browser does what the spec says.** Chromium, Gecko and WebKit all ship `text-align-last` as an
  independent longhand and **none of them implements `text-align-all`**, so the reset does not happen
  in a browser either. Implementing the shorthand faithfully would make PeachPDF diverge from every
  renderer an author checks their document against, in exchange for matching prose no engine follows.
- **It is not a cheap change.** `text-align` would have to become a real `ShorthandProperty`
  (`PropertyFactory.AddShorthand`), which changes what `StyleDeclaration.TextAlign` returns —
  `MarginBoxRenderer.ResolveAlignment` reads exactly that for `@page` margin-box alignment. The
  render-side registry entry in `css-properties.json` would have to move from `text-align` to a new
  `text-align-all`, touching the presentational-attribute path (`DomParser`'s `align=` handling writes
  `box.TextAlign` directly) and the UA stylesheet (`th`, `caption`, `center`).

### What would close it

1. Add `text-align-all` as the longhand the render layer reads, and re-point the
   `css-properties.json` entry, `DomParser`'s `align=` handling and `CssDefaults`' UA rules at it.
[...]

PR #1028, opened 2026-09-13
Align a line before a forced break by text-align-last

### LINKED ISSUE #1021: text-align: justify justifies the last line before a forced break

ChrisVanDijk:
Fixes #1021

### The bug

[css-text-3 §6.1](https://www.w3.org/TR/css-text-3/#text-align-property) ends a paragraph at a forced
line break as much as at the end of a block:

> Unless otherwise specified by `text-align-last`, the last line before a forced line break is
> start-aligned.

PeachPDF exempted only the block's own last line, so under `text-align: justify` a line ended by a
`<br>` — or by a preserved newline under `white-space: pre`/`pre-wrap`/`pre-line` — was stretched to
the full measure. A `<br>`-separated address block came out like this:

```
Peach          State          Technologies
   1     Example   Parkway,   Suite    400
        Atlanta,       Georgia     30303
United States of America
```

### The fix

**Record the fact where it is known, rather than deriving it later.** The flow already sets
`CssLineBox.FollowsForcedBreak` on the line a break *opens*; it now sets the mirror,
`PrecedesForcedBreak`, on the line the same break *closes* — in `FlowBox`'s wrap branch and in
`CreateVerticalLineBoxes`'. That is what makes this more than a one-liner tractable:

- no lookahead, which would have cost an `IndexOf` per line and made finalization quadratic in a long
  block;
- it survives a **fragmentation break**, where the pass that stops at the break discards the line the
  `<br>` opened and finalizes the rest with `blockFinished: false` — the successor does not exist yet,
  but the flag was already stated.

**"Start-aligned" is only `text-align-last`'s initial `auto` behaviour**, so the property (§6.3) is now
implemented rather than start alignment being hard-coded a third time. It cost one entry in
`css-properties.json` — inherited, initial `auto`, an `enum-keyword` over the `TextAlignLast` enum the
CSS-OM layer already had and the render layer had never consumed — and the source generator emits the
storage, cascade entry, initial value and getter.

**The "which lines are exempt" decision moved up**, out of `ApplyJustifyAlignment` /
`ApplyVerticalJustifyAlignment` and into one `ResolveUsedAlignment` both dispatchers call. It resolves,
in order: §6.1/§6.3 (a paragraph-ending line is aligned by `text-align-last`), then §6.4.3 (a justified
line with no justification opportunity is *unexpandable text*, also aligned by `text-align-last`, and
start-aligned when `text-align-last` is itself `justify`). The justify routines now only stretch, and
take the opportunity count as a parameter instead of re-walking the line for it.

### RTL corrections that fall out

Three shapes in a justified RTL block used to be left at the physical **left** edge and now flush right:
the paragraph's closing line, an unexpandable line, and one already too long for its measure (§6.1's
overflowing line, which spills past the left edge the way a non-justified RTL overflowing line already
did).

### Out of scope

`text-align` stays an independent longhand rather than a shorthand over
`text-align-all`/`text-align-last`, matching every shipping browser — none implements `text-align-all`.
Tracked as #1027 and recorded in `.claude/accepted-gaps/`.

### Verification

- Full suite `--framework net8.0`: **11 058 passed / 0 failed / 9 skipped**.
- 23 new tests in `TextAlignLastTests`; 5 confirmed to fail against the unfixed dispatchers, including
  the paginated and vertical-writing-mode cases.
- `diff-cover` against `main`: **100%** of 46 changed lines.
- `dotnet build PeachPDF.slnx -t:Rebuild -c Release`: 0 warnings, 0 errors.
- All 115 showcases re-rendered and pixel-diffed against a baseline built from stashed code: **exactly
  one changed**, the new `text_align_last` one. Nothing else in the gallery pairs `justify` with a
  `<br>` or with `direction: rtl`. The new showcase was rasterized through both PDFium and MuPDF, which
  agree.

Docs updated in `docs/html-css-support.md` (`text-align` row rewritten, `text-align-last` row added);
notes added under `.claude/` for the migration, the reasoning, and the invariant a future change must
not violate.


ChrisVanDijk:
Re-ran and all three platforms pass, so the earlier `test (ubuntu-latest)` failure was a flake.

Worth flagging for later though: `GsubLigatureMatchAllocationTests.WalkingARunForLigatures_AllocatesNothingPerPosition`
is fragile as written. It warms with 3 calls and then measures 50, which straddles .NET's tier-0 →
tier-1 promotion (30 calls plus a background call-counting delay), so what lands inside the measured
window depends on machine speed and load. The 4,096-byte bound leaves no headroom for that.

The failure was also nowhere near the regression the test guards against — 6,616 bytes over 10,000
match attempts is 0.66 bytes per attempt, where the defect it exists to catch costs over half a
megabyte.

Happy to harden it in a separate PR off `main`: warm past the promotion threshold before measuring,
rather than raising the bound (raising it would weaken the very assertion the test makes).

Note there's an empty "Re-run CI" commit on this branch — worth squashing or rebasing out before merge.

ChrisVanDijk:
Follow-up: the flaky `test (ubuntu-latest)` failure from earlier now has a fix in #1029, on its own
branch off `main` — no changes to this PR.

Short version: six allocation guards across five test files shared the same hand-rolled shape (warm 3
calls, measure one batch, assert under 4,096 bytes), which straddles tier-0 → tier-1 promotion and so
measures something machine-dependent. #1029 extracts a shared `AllocationProbe` that warms past the
promotion threshold and takes the smallest of several batches.

Sensitivity is unchanged, and checked rather than assumed: with the boxing regression reintroduced into
`TryMatchLigature`, the hardened test still fails at 720,000 bytes against the same 4,096 bound — 176×
over. Raising the bound would have been the easy fix and would have quietly disabled the assertion.

jhaygood86 (changes requested):
Excellent, thoroughly-tested work. Reviewed in depth given the blast radius:

- **css-text-3 §6.1/§6.3 core claims**: fetched the live spec directly — §6.1's exact wording ("Unless otherwise specified by text-align-last, the last line before a forced line break is start-aligned") matches the PR's quote verbatim, as does §6.3's `text-align-last` definition.
- **`PrecedesForcedBreak` tracking**: traced both `FlowBox`'s wrap branch and `CreateVerticalLineBoxes`, confirmed it's genuinely stated when the break closes the line (not derived via lookahead), and confirmed the fragmentation-survival claim — the flag on an already-finalized line survives a page break where the successor line doesn't exist yet. Mutation-tested by dropping both reads: 6/23 `TextAlignLastTests` fail (including the paginated and vertical-writing-mode cases), all pass restored.
- **`text-align-last` end-to-end**: confirmed the CSS-OM enum pre-existed but was unconsumed before this PR; the generator wiring is clean (no PPG010/011 diagnostics), and `right`/`center`/`left`/`justify`/`start`/`end` are all tested explicitly, not just the default.
- **RTL corrections**: all three reproduced and asserted directly on line-box/word geometry.
- **Accepted gap #1027**: real, accurately described, and matches the "no browser implements the literal spec text either" rationale that should also apply to the finding below.
- Full suite passes (11,058, exact match), diff coverage 100%, showcases: exactly one new (`text_align_last`), zero visual differences across all 114 pre-existing ones, rasterized through both PDFium and MuPDF in agreement. CI is green on all three platforms.

**One real, confirmed finding that needs the same treatment as #1027 before merge**: `ResolveUsedAlignment` (`src/PeachPDF/Html/Core/Dom/CssLayoutEngine.cs:4217-4218`) resolves a doubly-unexpandable line (justified, no opportunity, AND `text-align-last` itself resolves to `justify`) to `towardStart`. The code's own remarks (lines 4192-4201) describe only the general "defer to text-align-last, else start" rule and don't mention this case at all — but css-text-3 §6.4.3's actual text has a second, separate clause for exactly this situation: *"(If text-align-last is justify, then they must be aligned as for center.)"* That clause is neither implemented nor disclosed.

I independently verified (cross-checked against Chromium, Firefox, and WebKit via Playwright, not just one engine) that all three real browsers start-align here too — so the runtime behavior is correct and matches the actual web platform, exactly the same situation as #1027's `text-align-all` gap. But per CLAUDE.md's own convention (which this PR already follows correctly for #1027), a confirmed spec deviation needs a tracked issue and an `.claude/accepted-gaps/` file, not an incomplete citation in code comments/docs that makes it look like full compliance. Please file a tracking issue for this (same "no browser implements the spec's literal text either" framing works), add the matching accepted-gap file, and correct the citations in `CssLayoutEngine.cs`'s remarks, the recent-fixes note, and the `text-align-last` row in `docs/html-css-support.md` to say "matches browser behavior" rather than attributing the start-alignment to §6.4.3 as if it were literal compliance.

This doesn't touch any actual behavior — purely a documentation/tracking gap. Happy to take another look once it's filed.

ChrisVanDijk:
Good catch — filed as #1031 with an accepted-gap file, and the citations are corrected in
36397244. No behaviour change.

Two corrections rather than one: the remarks also *misquoted* §6.4.3 as "is start-aligned, unless
`text-align-last` specifies otherwise", which is §6.1's wording. §6.4.3 actually says "must be aligned
as specified by the `text-align-last` property", followed by the parenthetical. Both are fixed.

Disclosed at each place a reader meets the behaviour:

- `ResolveUsedAlignment`'s remarks now quote §6.4.3 accurately and carry a **Deliberate deviation**
  block naming the parenthetical, the three engines that don't implement it, and the gap file; the
  `fallback == Justify` line has its own note so the `towardStart` isn't read as an oversight
- `.claude/accepted-gaps/unexpandable-justified-line-starts-rather-than-centres.md` (new, #1031)
- the recent-fixes note: step 3 of the rule list states the deviation instead of implying compliance,
  and "Not done" lists it
- `docs/html-css-support.md`, `text-align-last` row: "PeachPDF start-aligns it, which is what Chromium,
  Firefox and Safari all do (§6.4.3 asks for centring here, and no browser engine implements that)"
- the RTL test's comment, which no longer attributes the start-alignment to §6.4.3

One thing I made explicit in the gap file, because it's the trap: closing this is a one-word change
(`Center` instead of `towardStart`). That is exactly why it needed writing down — a future reader would
otherwise take the `towardStart` for a bug and "fix" it into a browser-divergent result. The file also
separates the two things that look alike: resolving to *start* rather than leaving the line where the
flow put it is load-bearing under RTL and correct; only start-versus-centre is the deviation.

Suite still 11,058 / 0 and the solution builds warning-free.

jhaygood86 (approved):
Exactly the right fix, and more thorough than I asked for — the accepted-gap file (#1031) is exemplary (states the exact one-line change that would close it, the browser cross-check evidence, and explicitly separates the load-bearing RTL start-resolution from the deviation itself so a future reader can't conflate the two). It also caught and corrected a second citation error I'd missed: the original remarks misattributed §6.1's "is start-aligned, unless text-align-last specifies otherwise" wording to §6.4.3.

Rebuilt clean (0 warnings), full net8.0 suite passes (11,058, exact match). CI is green on all three platforms. Approving.

PR #1029, opened 2026-09-13
Measure allocation guards over steady state, not one loop

ChrisVanDijk:
### The trigger

`GsubLigatureMatchAllocationTests.WalkingARunForLigatures_AllocatesNothingPerPosition` failed on
`test (ubuntu-latest)` during #1028 — a branch that touched no font or shaping code at all — reporting
**6,616 bytes** against its 4,096-byte bound. The `windows-latest` and `macos-latest` legs of the same
`fail-fast: false` matrix passed, the full suite passed locally in Debug, Release, and Release with
coverage collection (the exact CI invocation), and a re-run went green on all three.

The number is what identifies it: 6,616 bytes over 10,000 match attempts is **0.66 bytes per attempt**,
where the regression the test exists to catch costs **720,000 bytes**. The failure was a fixed lump
landing inside the counted window, not per-position allocation returning.

### Why one loop can't measure this

.NET promotes a method out of tier-0 after 30 calls, and only after a background call-counting delay
that restarts while other methods are still being called for the first time — which, in a suite running
test collections in parallel, is continuously. A three-call warm-up therefore leaves the method at
tier-0 and puts promotion somewhere *inside* the loop being counted, at a point that depends on the
machine. That is why a CI runner sees it and a dev box doesn't.

### The change

Six measurement sites across five files had that same hand-rolled shape — warm 3, measure one batch,
assert under 4,096 — each a copy of the same defect. So this extracts `TestSupport.AllocationProbe`
rather than patching the one that happened to flake:

- `Text/GsubLigatureMatchAllocationTests.cs`
- `Text/GposPositionerAllocationTests.cs`
- `CSS/CssDataSelectorMatchAllocationTests.cs` (three sites)
- `CSS/StyleRuleLookupAllocationTests.cs`
- `CSS/TokenValueCapacityTests.cs`

The probe warms past the promotion threshold (64 calls) and returns the **smallest** of several
measured batches. The minimum is the right estimator because of what each kind of allocation does to
it: a genuine per-iteration allocation appears in *every* batch and survives the minimum untouched,
while a one-off transient lands in one batch and is discarded. It cannot hide a regression; it can only
discard a coincidence.

Each site keeps its own iteration count, bound and failure message. Only the measurement is shared.

### Sensitivity is not traded away — proven, not argued

Raising the bound would have been the easy fix and would have destroyed exactly the assertion the test
makes. So the hardened test was run against the *reintroduced* defect — the
`foreach`-over-`IReadOnlyList` boxing plus the eager `skippedOffsets` list in `TryMatchLigature`:

> **720,000 bytes against the 4,096 bound — still fails by 176×.**

`GsubShaper.cs` was then reverted; only test files are touched by this PR.

### Verification

- Full suite `--framework net8.0 --configuration Release`: **11 035 passed / 0 failed / 9 skipped**
- `dotnet build PeachPDF.slnx -t:Rebuild -c Release`: 0 warnings, 0 errors
- The six converted allocation tests in isolation: 13/13 pass

### Notes

Recorded as an invariant rather than a recent-fix note — it is a rule future tests must follow, not a
change that ages out. It cross-references the existing
`testing-a-complexity-guard-asserts-a-count-not-the-clock.md`, which is the same trade made in the
other direction: a bound loose enough to absorb noise is a bound loose enough to pass the regression it
was named for.


jhaygood86 (approved):
Sound test-infrastructure fix, test-only (confirmed GsubShaper.cs itself is untouched in the diff). The minimum-of-several-batches methodology is the statistically correct estimator for this exact noise shape (additive one-off transients vs. genuine per-iteration allocation), and the tier-0/tier-1 promotion explanation matches known .NET tiered-JIT behavior under concurrent load.

I verified the sensitivity claim myself rather than trusting it: reintroduced the described regression in TryMatchLigature (foreach-over-IReadOnlyList boxing plus fresh matched/skipped lists per position instead of the reused scratch fields), rebuilt, and reran WalkingARunForLigatures_AllocatesNothingPerPosition — it failed at 400,000 bytes against the 4,096 bound (~98x over), confirming the hardened probe's minimum-of-batches approach doesn't accidentally mask a real regression. Restored the file and confirmed clean.

All 13 converted tests pass, rebuilt clean (0 warnings), full net8.0 suite passes (11,035, exact match). CI is green on all three platforms. Approving.

Issue #1033, opened 2026-09-13
A float does not contribute to the line it sits beside in the intrinsic-width walk


ChrisVanDijk:
A float is taken out of flow but is still placed *beside* the inline content of the block it is in
([CSS 2.1 §9.5](https://www.w3.org/TR/CSS21/visuren.html#floats)), so for max-content sizing its width
adds to that line. `CssBox.GetMinMaxSumWords` instead treats a float as a box that **opens a line of its
own**, competing with the text beside it rather than adding to it — correctly as far as
`CssBox.StartsNewLine` goes, since CSS 2.1 §9.7 blockifies a float, but wrong for what the walk needs to
know.

### Reproduction

At `font: 16px monospace` (one character advances 6.5977pt):

```html
<div style="float:left"><!-- and again with white-space:nowrap -->
  XY <span style="float:left">ZZZZ</span>
</div>
```

| outer box | before #1017 | after #1017 | expected |
|---|---|---|---|
| plain | 26.3906 | 26.3906 | 46.1836 |
| `white-space: nowrap` | 46.1836 | **26.3906** | 46.1836 |

26.3906 is `max("XY " = 19.7930, "ZZZZ" = 26.3906)`; 46.1836 is their sum. The consequence is the usual
one for an undercount out of this walk: a shrink-to-fit box (a float, a `position: absolute` box, an auto
table column) sized narrower than the content it holds, with the float overprinting its neighbour.

The expected column is the additive result the spec implies, and matches what this engine itself produced
for the `nowrap` row before #1017. A browser was **not** available to the session filing this — a review
pass reported Chromium giving 46.1836, but that figure has not been independently re-measured here and
should be confirmed before it is treated as the reference.

### The `nowrap` row is a regression from #1017, and was previously right by accident

Before #1017, `StartsNewLine` returned false for any `white-space: nowrap` box, so a float inside one did
not open a line and was summed onto it — the right answer for a reason that had nothing to do with floats.
#1017 removed `white-space` from that predicate (it says whether content wraps *within* a line, not
whether a box opens one, and reading it there was wrong for every other shape, measurably so), and the
`nowrap` case joined the plain case's pre-existing defect. Restoring the old predicate is not an option;
the fix has to be about floats.

### Why it was out of scope for #1017

"A float adds to the line" is not unconditionally true, so this is not a predicate tweak. A float between
two block-level siblings does not add to either one's line, and the flat walk has no notion of which
[...]

Issue #1032, opened 2026-09-13
An atomic inline whose content is block-level replaces the line instead of adding to it in the intrinsic-width walk


ChrisVanDijk:
`CssBox.GetMinMaxSumWords` — the intrinsic (min-content/max-content) walk that sizes a float, a
`position: absolute` box, an auto table column and a flex item — carries **one** running line total
across a whole subtree. An atomic inline-level box therefore isn't measured as a unit: the walk
descends into it, and a **block-level** child inside it resets that running total, discarding whatever
was already on the outer line. The epilogue then restores it with `Math.Max` rather than a sum.

An atomic inline contributes its own max-content width to the line it sits on, as one unit
([css-display-3 §2.3](https://www.w3.org/TR/css-display-3/#atomic-inline),
[css-sizing-3 §5.2](https://www.w3.org/TR/css-sizing-3/#intrinsic-contribution)). The walk computes
`max(outer line, inner content)` where it must compute `outer line + inner content`.

### Reproduction

Float widths at `font: 16px monospace` (one character advances 6.5977pt), via the
`IntrinsicWidthWalkTests` harness:

| markup inside `<div style="float:left">` | measured | expected |
|---|---|---|
| `AB <span style="display:inline-block">CD</span>` | 32.9883 | 32.9883 ✅ |
| `AB <span style="display:inline-block"><div>CD</div></span>` | **19.7930** | 32.9883 |
| `AB <span style="display:inline-block"><div>CD EF GH</div></span>` | **52.7812** | 72.5742 |
| `AB <span style="display:inline-block"><div>CD</div><div>EF GH</div></span>` | **32.9883** | 52.7812 |
| `AB <span style="display:inline-table"><div>CD</div></span>` | **19.7930** | 32.9883 |
| `AB <span style="display:inline-grid"><div>CD</div></span>` | **19.7930** | 32.9883 |
| `AB <span style="display:inline-flex"><div>CD</div></span>` | 32.9883 | 32.9883 ✅ |

Every wrong row is exactly `max("AB " = 19.7930, inner max-content)`. The first row is right only
because there is no block-level child to trigger the reset, and a box with nothing before it on the
line (`<span style="display:inline-block"><div>CD EF</div></span>` alone → 32.9883) is right for the
same reason — there is nothing to discard.

### `inline-flex` is the oracle

The last row is the same markup with the same inner block content, and it is **correct** — because
[...]

Issue #1003, opened 2026-09-11
Inline-level flex/grid items are not blockified (CSS Display 3 §2.7)


ChrisVanDijk:
## The rule

[CSS Display 3 §2.7](https://www.w3.org/TR/css-display-3/#blockify) blockifies every in-flow child of a
flex or grid container, as required by
[css-flexbox-1 §4](https://www.w3.org/TR/css-flexbox-1/#flex-items) — "the `display` value of a flex item
is blockified" — and css-grid-2 §6. An item declared `inline`, `inline-block`, `inline-flex` or
`inline-table` computes to `block`, `block`, `flex` and `table` respectively.

## What PeachPDF does

`DomParser.NormalizeFlexOrGridItem` blockifies only the **layout-internal** set (css-display-3 §2.6's
table-internal displays → `block`). An inline-level item keeps its declared computed value.

```html
<div style="display: flex">
  <span id="item">x</span>   <!-- computed display stays `inline`; spec says `block` -->
</div>
```

## Why it is mostly invisible

The flex and grid engines already lay every item out blockified
(`CssLayoutEngineFlex.PerformLayoutBlockified`), so an inline item's own `width`/`height` do apply and it
is measured, aligned and painted as a block. What genuinely broke before this area was last touched was
the *box tree* — CSS 2.1 §9.2.1.1's anonymous block wrapper being generated inside a flex container,
which swallowed the real item — and `DomParser.CorrectInlineBoxesParent` handles that now.

So the remaining deviation is the computed value itself. It is observable through a future
`@supports`-style or computed-style surface, not through layout.

## Why it was deferred

Blockifying inline-level items was implemented and then reverted, because it regresses replaced
elements. A replaced box takes its size from the phantom word carrying its content, and that size only
reaches the box through inline flow. As a block-level box it goes through `CssLayoutEngine.GetBoxWidth`
instead, whose `width: auto` fills the containing block rather than using the element's intrinsic width —
CSS 2.1 §10.3.4's intrinsic width for block-level replaced content is not implemented.

The result is that an `<img>` or inline `<svg>` flex item loses its intrinsic size outright. Five tests
caught it; `FlexboxIntegrationTests.Column_AlignItemsCenter_ReplacedItem_Centers` is the clearest, and
`PixelsPerInchAbsoluteSizingIntegrationTests.ImgAndSvg_DirectFlexItem_NoBlockSibling_ResolveIdenticallyUnderNonDefaultPixelsPerInch`
[...]

PR #1034, opened 2026-09-13
Decide a line start from display, not white-space, in the intrinsic-width walk

### LINKED ISSUE #1017: A `white-space: nowrap` block does not start a new line in the intrinsic-width walk, summing sibling lines into one

ChrisVanDijk:
`CssBox.StartsNewLine` — the predicate `GetMinMaxSumWords` resets its running line total on — had two
wrong answers that partly masked each other, so fixing either alone makes measurable cases worse.

**It read `white-space: nowrap` as "this box does not begin a line."** But `white-space` says whether a
box's content wraps *within* a line; it says nothing about whether the box begins one, and a block-level
box always does ([CSS 2.1 §9.4.1](https://www.w3.org/TR/CSS21/visuren.html#block-formatting)). A `nowrap`
block-level sibling therefore had its line **added** to the previous sibling's instead of competing with
it for "widest line wins" — one whole line too wide per sibling, growing linearly with the sibling count.
`paddingSum` lost its per-line scoping the same way, since `oldPaddingSum` is saved in the same branch.

**It also read every inline-level display as "this box does begin a line,"** which an inherited `nowrap`
happened to undo. So an `inline-block`/`-flex`/`-table`/`-grid` was measured correctly only in a document
that declared a `nowrap` it had no other reason to declare.

Float widths at `font: 16px monospace` (one character advances 6.5977pt), against Chromium 148:

| markup | before | after | Chromium |
|---|---|---|---|
| `<div>AB CD</div><div style="white-space:nowrap">EF</div>` | 46.1836 | 32.9883 | 32.9883 |
| …plus a third `<div>GH</div>` sibling, all under `nowrap` | 59.3789 | 32.9883 | 32.9883 |
| two `border-left: 10pt solid` siblings under `nowrap` | 56.1836 | 42.9883 | 42.7383¹ |
| `AB <span style="display:inline-flex">CD</span>` | 19.7930 | 32.9883 | 32.9883 |
| `AB <span style="display:inline-block;width:50pt">` | 50.0000 | 69.7930 | 69.7852 |

¹ Chromium rounds a border width down to a whole device pixel (10pt = 13.333px → 13px = 9.75pt); the
0.25pt is that rounding, not a measurement disagreement.

## `ActualDisplay` alone is not a sufficient oracle

Deciding "is this box block-level" from `DerivedStyle.ActualDisplay` looks right — it blockifies a float
per CSS 2.1 §9.7, and `DomParser.BlockifyPositionedBox` handles `position: absolute`/`fixed`. But
`NormalizeFlexOrGridItem` deliberately blockifies only the **table-internal** set, leaving an inline-level
flex/grid item's computed display alone (#1003). Left that way, a single-line flex **column** of
`inline-block` items measured as the **sum** of its items rather than the widest of them — 72.5742pt
against the 39.5859pt the same container gives for plain `<div>` items — and took that width out of the
table column beside it.

`IsFlexOrGridItem` asks the **parent's** display instead, since
[css-display-3 §2.7](https://www.w3.org/TR/css-display-3/#blockify) blockifies an item whatever its own
computed value says. That also corrects a case that was already wrong on `main`: a bare `display: inline`
item in a flex column, 72.5742 → 39.5859.

This was caught by the post-change review pass, not by the suite — nothing covered the shape. The new
theory fixture does now, and note what it does and does not prove: only its `inline` case fails against
`main`, because the *old* predicate already had the four atomic-inline displays opening a line (for the
wrong reason). It guards a regression relative to the fix-in-progress.

## Gaps left behind, both tracked

- **#1032** — an atomic inline whose content is block-level is still not measured in isolation. Unchanged
  by this PR either way (same number before and after); the discard happens at the block-level *child's*
  reset, not at the atomic inline's own boundary. `inline-flex` already gets it right via `IsFlexRow`,
  which is the shape the fix should take.
- **#1033** — a float inside a `white-space: nowrap` shrink-to-fit box no longer contributes to the line
  it sits beside: 46.1836 → 26.3906. **A genuine regression from this PR**, and it was previously right by
  accident — the old predicate excused any `nowrap` box from opening a line. Without `nowrap` the same
  shape measured 26.3906 before this change too, so the defect class is pre-existing. Not fixed here
  because "a float adds to the line" is not unconditionally true and the flat walk cannot tell which
  formatting context a child is in.

## Verification

- `IntrinsicWidthWalkTests`, 30 fixtures, 7 new; 9 fail against the merge base.
- Full suite green on net8.0: 11,071 passed / 0 failed / 9 skipped.
- 100% diff coverage on the changed production lines.
- **All 115 showcases byte-identical.** Getting a trustworthy answer there needed four independent
  sources of run-to-run randomness normalized — font subset tag, creation date in both Info and XMP
  form, per-annotation `/NM` GUIDs, and the document `/ID` — each of which produced a false positive:
  the subset tag alone made 2 files differ under MuPDF while PDFium rendered them identically, and the
  GUIDs alone made 9 files differ at object level while PDFium found zero differing pages across all 23.
  That list is recorded in the fix note so the next sweep starts from it.

Fixes #1017


ChrisVanDijk:
I will start fixing #1033 soon.

jhaygood86 (approved):
Reviewed in depth given the blast radius (same intrinsic-width-walk family as #1018/#1022/#1028). Independently verified nearly everything:

- CSS 2.1 §9.4.1 and css-display-3 §2.7 citations confirmed against fetched spec text.
- Traced StartsNewLine/IsFlexOrGridItem directly — block-level-ness now correctly comes from ActualDisplay, and the parent-display lookup for flex/grid items is correctly ordered (own-display check short-circuits first).
- Independently re-measured Chromium for the main before/after table rows and the #1032 comparison rows — all matched exactly.
- Mutation-tested: reverting CssBox.cs to the merge base fails 10 of 30 IntrinsicWidthWalkTests (PR claims 9 — trivial off-by-one in counting, every failure matches the PR's narrative exactly, including the subtle detail that only the flex-column `inline` case fails on the old code since the other four atomic-inline displays were already accidentally excluded). Restored: 30/30 pass.
- The self-disclosed new regression (#1033) reproduces exactly as described in PeachPDF's own before/after numbers (46.1836→26.3906pt), and is honestly filed as a tracked accepted-gap rather than glossed over.
- Full showcase byte-identical claim reproduced: after normalizing a couple of additional randomness sources the PR's own description didn't enumerate (a verbose PDF-writer header timestamp, per-annotation /M dates), all 115 showcases come out genuinely byte-identical.
- Rebuilt clean (0 warnings), full net8.0 suite passes (11,071, exact match). CI is green on all three platforms.

One minor, non-blocking note for a quick follow-up: the accepted-gap file for #1033 already honestly flags its "expected" browser value as unverified — and it turns out to be wrong when actually measured (real Chromium gives 39.5859pt for that repro, not 46.1836pt, because the float there ends up positioned before the preceding text rather than after it). Doesn't change the conclusion that PeachPDF's own before/after numbers are correct and the regression is real and narrow — just worth updating the reference number in `.claude/accepted-gaps/a-float-does-not-contribute-to-the-line-in-the-intrinsic-walk.md` and issue #1033 whenever convenient. Not blocking since the file was already careful to flag it as unverified rather than asserting it.

Approving.

Issue #1040, opened 2026-09-13
The intrinsic walk's padding total is neither an outer width nor the winning line's own padding


ChrisVanDijk:
`CssBox.GetMinMaxSumWords` accumulates border/padding in `paddingSum`, a running total kept
*separate* from `maxSum` (the running line width) and combined across boxes by `Math.Max` rather
than by addition. `GetMinMaxWidth` then returns `paddingSum + Math.Max(maxSum, widestLine)`.

That is not a well-defined quantity, and it is wrong in three independent ways. All three are
pre-existing and none is caught by the suite today.

### 1. A containment chain's paddings are maxed, not added

A child's border box has to fit inside its parent's content box, so both count. The reset at
`StartsNewLine` goes to `0` and the epilogue restores with `Math.Max(paddingSum, oldPaddingSum)`,
so an ancestor's decoration vanishes whenever a descendant declares at least as much.

```html
<div style="position:absolute; padding: 0 2pt"><div style="padding: 0 10pt">ABCD</div></div>
```

resolves 20pt of padding where 24pt has to fit; the inner box overflows the outer one. Identical
with `border` in place of `padding`. The `Math.Max` is right for *siblings* (they are on different
lines, so the widest wins — that is the Acid2 `#eyes-a`/`#eyes-b`/`#eyes-c` case it was added for)
and wrong for nesting; one save/restore currently serves both.

### 2. The padding added is not the winning line's own

`maxSum` and `paddingSum` are chosen independently, so the decoration of one line can be added to
a *different* line's width. The `writing_mode` showcase has this shape:

```html
<div>                                          <!-- flex item -->
  <div class="vbox" style="width: 200px">      <!-- border 2px + padding 8px -->
    <div style="padding: 4px">First.</div>     <!-- and two more -->
  </div>
  <div class="label">vertical-rl: block children stack right-to-left</div>
</div>
```

The item measures 171.81pt. Its max-content is `max(vbox outer = 165pt, label text = 156.81pt)` =
165pt: the widest line is the label's, which has no padding at all, and the 15pt that gets added to
it is the `.vbox`'s. Fixing (1) alone makes this *worse* (177.81pt), because the two defects push in
opposite directions.

### 3. Consumers disagree about what the number is

- `CssLayoutEngineTable.GetColumnsMinMaxWidthByContent` treats it as an **outer** width (its own
  comment at `CssLayoutEngineTable.cs` says `cell.GetMinMaxWidth()` "already includes them").
[...]

Issue #1038, opened 2026-09-13
A float that follows inline content is placed on the next line, not beside it


ChrisVanDijk:
A float that follows inline content in source order is not placed on that content's line. It is
laid out at the block's left (or right) edge, on top of the text it should sit beside, and the
inline content after it starts a new line instead of continuing beside it.

```html
<div style="width: 300pt; font: 16px monospace">XY <span style="float:left">ZZZZ</span> more words</div>
```

Laid out today, at `font: 16px monospace` (one character advances 6.5977pt):

| box | x | y |
|---|---|---|
| `XY` | 20 | 20 |
| the float | 20 | 20 |
| `more words` | 20 | 34.25 |

`XY` and the float are drawn at the same point, and `more words` is on the second line. A browser
puts the float immediately after `XY` on the first line and continues `more words` beside it.

A float whose source position is *before* the inline content is placed correctly and the text does
wrap around it, so this is specifically about a float that follows content on its own line.

## Spec

[CSS 2.1 §9.5](https://www.w3.org/TR/CSS21/visuren.html#floats): a float is taken out of the flow
but is still placed beside the inline content of the block it is in, and §9.5.1 rule 6 puts the
float's outer top no higher than the top of the current line box — the line the content before it
is on. The content after it flows down the float's side rather than starting a new line.

## Where it comes from

`DomParser.CorrectInlineBoxesParent` wraps the inline run in a CSS 2.1 §9.2.1.1 anonymous block,
because §9.7 blockifies the float and `ContainsVariantBoxes` then sees mixed block-level and
inline-level children. That wrapper ends the line at the float, so the float is placed as a
block-level sibling after it rather than on it. Browsers do not generate that wrapper: an
out-of-flow box does not make a block container's children non-inline.

## Out of scope for #1033

#1033 was the same shape in the intrinsic-width walk — the float not adding its width to the line
it sits beside, so a shrink-to-fit box was sized to the wider of the two rather than their sum.
That is fixed, and the box now reserves the room; this issue is that layout does not then put the
float in it. The fix here is a box-tree change (not generating the wrapper for an out-of-flow
child, which means `DomUtils.ContainsInlinesOnly`, `CssBox.LayoutContents`' inline-vs-block
dispatch, and `CssLayoutEngine.FlowBox` all have to handle a floated child among inline content),
which is a different and much larger change than the measurement one.

Recorded in `.claude/accepted-gaps/a-float-after-inline-content-is-placed-on-the-next-line.md`.

PR #1042, opened 2026-09-13
Add a float to the line it sits beside in the intrinsic-width walk

### LINKED ISSUE #1033: A float does not contribute to the line it sits beside in the intrinsic-width walk

ChrisVanDijk:
Fixes #1033

A float is taken out of the flow but is still placed *beside* the inline content of the block it is
in ([CSS 2.1 §9.5](https://www.w3.org/TR/CSS21/visuren.html#floats)), so for max-content sizing its
width **adds** to that line. `CssBox.GetMinMaxSumWords` saw only that §9.7 blockifies it — which
`StartsNewLine` reads correctly — and had it open a *competing* line, measuring `max(text, float)`
where the answer is `text + float`. A shrink-to-fit box was then sized to the wider of the two, and
the float drew over the content beside it.

Float widths at `font: 16px monospace` (one character advances 6.5977pt):

| markup inside `<div style="float:left">` | before | after |
|---|---|---|
| `XY <span style="float:left">ZZZZ</span>` | 26.3906 | **46.1836** |
| …the same under `white-space: nowrap` | 26.3906 | **46.1836** |
| `<span style="float:left">AAA</span><span style="float:left">BBB</span>` | 19.7930 | **39.5859** |
| `<p>AB CD</p><span style="float:left">EF</span>` | 32.9883 | 32.9883 (unchanged, and must be) |

46.1836 is `"XY" + space + "ZZZZ"` — the same seven characters as one plain run. The `nowrap` row is
a regression from #1017, where the old predicate excused any `nowrap` box from opening a line and so
summed the float onto it for a reason that had nothing to do with floats.

## What it took

"A float adds to the line" is **not** unconditionally true — one between two block-level siblings is
on neither one's line — so this is not a predicate tweak. The walk needed one fact it did not have:
**which of a box's block-level children are [CSS 2.1 §9.2.1.1](https://www.w3.org/TR/CSS21/visuren.html#anonymous-block-level)
anonymous wrappers around its own inline run.** `DomParser.CorrectInlineBoxesParent` generates one of
those around `XY ` here, because the blockified float makes `ContainsVariantBoxes` true, and nothing
on the box recorded it. `CssBox.IsInlineRunWrapper` now does, set at the single place that creates
them, and two predicates read it:

- `FloatsShareTheLine(box)` — every in-flow child is inline-level content (an inline box, or one of
  those wrappers), so a float child is beside it. Vacuously true for a box whose children are all
  floats, which is right: they do sit side by side.
- `SharesItsLineWithAFloat(box)` — the same question from the wrapper's side, so `StartsNewLine`
  returns false for it and the line is not closed at the float. This is what keeps the space:
  without it the wrapper's epilogue hangs `XY `'s trailing space (css-text-3 §4.1.2) and the answer
  is 39.5859, one space short.

The float is then measured in isolation via its own top-level `GetMinMaxWidth` and added — the
mechanism `IsFlexRow` already uses per flex item, including its `trailingSpace = 0`. `min` takes the
wider of the running unbreakable run and the float's own min-content, since a float is both a wrap
opportunity and an unbreakable unit.

## Why the padding is folded rather than added

`paddingSum` is a running total kept separate from `maxSum` and combined across boxes by `Math.Max`
rather than by addition, so the isolated measurement's decoration cannot simply ride onto the line —
that *adds* what the recursive descent would have *maxed*. `GetMinMaxWidth` gained a `decoration`
out-parameter for exactly this, and the branch splits it back out and folds it the way the descent
did, reproducing the existing accounting exactly. **All 115 showcases are byte-identical** after
normalizing the four known run-to-run randomness sources, which is the evidence that it really is
unchanged.

Two alternatives were built and measured first, and both shipped a different defect — details, with
numbers, in `.claude/recent-fixes/2026-09-13-a-float-contributes-to-the-line-it-sits-beside.md`.

## Deliberately out of scope, tracked

- **#1038** — a float that *follows* inline content is still placed on the next line rather than
  beside it. The box now reserves the room; layout does not yet put the float in it. That is a
  box-tree change, not a placement tweak. Two adjacent floats are already placed correctly, so the
  measurement fix is fully visible there.
- **#1040** — the walk's padding total is neither an outer width nor the winning line's own padding,
  in three interacting ways. The visible cost of leaving it here: two floats that each declare
  padding lose one of the two paddings from the line they share.

Both have accepted-gap notes, and the user-facing limitations are described in
`docs/html-css-support.md` without citing the issues.

## Verification

- `IntrinsicWidthWalkTests`: 6 new fixtures, 5 of which fail against the merge base; the sixth is
  the block-level-sibling contrast case, which must keep passing and does.
- Full suite green on net8.0 — 11,077 passed / 0 failed / 9 skipped.
- 100% diff coverage on the changed production lines.
- All 115 showcases regenerated and compared byte-for-byte.


jhaygood86 (changes requested):
The direction here is right — a float genuinely adds its width to the line it sits beside (CSS 2.1 §9.5), and `FloatsShareTheLine`/the isolated `GetMinMaxWidth` + `decoration` accounting are sound. But the headline numbers are wrong, and it's specifically because `SharesItsLineWithAFloat` does two things at once that need to be separate.

**The bug:** `SharesItsLineWithAFloat` feeds into `StartsNewLine`, which this PR now also uses to decide whether the wrapper's trailing space hangs. But css-text-3 §1.5 says *"intervening inline box boundaries and out-of-flow elements must be ignored"* for whitespace-processing/adjacency purposes. A float is out-of-flow — so for hanging purposes, the trailing space in `"XY "` before `<span style="float:left">ZZZZ</span>` genuinely is at the end of the line's only in-flow content, and css-text-3 §4.1.2 says it should still hang. The recent-fix doc's own framing — *"without it the wrapper's epilogue hangs `XY`'s trailing space... and the answer is 39.5859, one space short, because the wrapper wrongly believes its line ended"* — has this backwards: the wrapper's belief is correct per spec, not wrong.

I checked this against real browsers rather than trust either number:

| markup | this PR | independently measured (Chromium + Firefox) |
|---|---|---|
| `XY <span style="float:left">ZZZZ</span>` | 46.1836pt | **39.5859pt** |
| …same under `white-space: nowrap` | 46.1836pt | **39.5859pt** |
| two adjacent floats (`AAA`+`BBB`) | 39.5859pt | 39.5859pt ✓ (unaffected — no space precedes either float) |
| `<p>AB CD</p>` + float (contrast case) | 32.9883pt | 32.9883pt ✓ |

I also reverted just the production code (keeping this PR's new tests) and confirmed the 5 failing-against-merge-base fixtures fail with exactly the old/new values the PR states (26.3906→46.1836, 19.7930→39.5859), so the mechanics of "float width now adds to the line" are correctly wired — it's specifically the trailing-space suppression riding along on the same `StartsNewLine` check that overshoots by one space width, in the one case (a space-terminated run immediately followed by a float) that's actually the common one.

**Requested change:** keep `FloatsShareTheLine`'s effect on whether the float's width sums onto the line, but don't let `SharesItsLineWithAFloat` suppress the wrapper's ordinary trailing-space-hang at `StartsNewLine` — a float being on the same line as the wrapper's content doesn't mean the wrapper's line hasn't ended for whitespace purposes; §1.5 says the float should be transparent to that decision. That likely means splitting the "does this box's line still take the float's width" question from the "does this box count as a line boundary for hanging" question, since `SharesItsLineWithAFloat` currently answers both.

All six new `IntrinsicWidthWalkTests` fixtures' expected values will need to be recomputed against real browser output once this is fixed — two of them (`AFloat_AddsToTheLineItSitsBeside`, `AnOutOfFlowOrUndisplayedSibling_DoesNotStopTheFloatSharingTheLine`) currently assert the wrong number in both cases; `ConsecutiveFloats_AddUp_RatherThanCompeting`, `AFloatBetweenBlockLevelSiblings_StillGetsItsOwnLine`, and `AFloatsDeclaredWidthAndMargins_LandOnTheLine` are unaffected (no space immediately precedes their floats) and I confirmed their current values independently against Chromium. The recent-fix doc and the migration note will also need their numbers/framing updated.

Everything else — the §9.5/§9.2.1.1 citations, the `IsInlineRunWrapper` plumbing, the padding-fold accounting, issues #1038/#1040 and their accepted-gap files, build (0 warnings) and full suite (11,077/0/9) — checks out and doesn't need changes.

ChrisVanDijk:
You're right, and the §1.5 citation is the part I should have gone looking for myself. Fixed in c16afeb7.

I checked the spec text before changing anything: css-text-3 §1.5 Text Processing does say
"intervening inline box boundaries and out-of-flow elements must be ignored", and §4.1.2 hangs an
end-of-line collapsible space under `normal` and `nowrap` alike. So the wrapper's belief that its
line ended is correct, and landing the float's *width* on the line does not make the float *content*
following that space. The recent-fix doc had that backwards exactly as you describe.

Worth recording why: 46.1836 came from reading Blink's `ComputeInlinePreferredLogicalWidths` from
memory rather than from a browser — and the accepted-gap file this branch deletes had *already*
flagged that number as reported-but-never-independently-measured. I built on it anyway. That's the
actual mistake.

**The split.** `SharesItsLineWithAFloat` keeps only "does the running line total carry on through
this box"; the float branch now hangs the space explicitly:

```csharp
maxSum -= trailingSpace;
trailingSpace = 0;

maxSum += floatMax + floatMargins;
```

rather than the flex-row branch's `trailingSpace = 0` — a flex row is in-flow and genuinely is
content following the space, a float is not. Both the predicate's doc comment and the float branch
now say which of the two questions they answer.

**Why the predicate is still needed.** Deleting it outright also yields 39.5859 for the headline
case (the wrapper's own epilogue hangs the space, then the float adds), so nothing in the previous
fixture set would have noticed. What it costs is a second run of inline content *after* the float:
the two runs land in separate §9.2.1.1 wrappers that each open a line, and
`XY <span style="float:left">ZZZZ</span> more` measures 39.5859 while holding three things.
`InlineContentEitherSideOfAFloat_StaysOnOneLine` pins that and fails against the merge base.

**One number I could not measure.** That new fixture's expectation is *derived*, not browser-checked
— css-text-3 §4.1.1 collapses the two spaces around the out-of-flow float into one, which is then
interior to the line, so the text measures as the single run `"XY more"` and the float adds to it
(72.5742pt at the fixture font). The fixture comment says so explicitly. If you have the browser
harness handy, that is the one value in the PR worth a second pair of eyes.

**Fixtures.** `AFloat_AddsToTheLineItSitsBeside` and
`AnOutOfFlowOrUndisplayedSibling_DoesNotStopTheFloatSharingTheLine` now compare against
`FloatWidthAsync("XYZZZZ")` — six characters, without the space — instead of `"XY ZZZZ"`. The other
four are untouched, as you predicted. Seven fixtures now, six failing against the merge base; the
seventh is the block-level-sibling contrast case, which must keep passing and does.

Docs updated: the recent-fix note's numbers and its whole "the wrapper wrongly believes its line
ended" framing, the migration note, and the trailing-space invariant file — which now records the
float branch as the one place the rule's two halves come apart, with the 46.1836-vs-39.5859 symptom,
since that is exactly the trap a future change would fall into again.

Re-verified: full suite green on net8.0 (11,078 / 0 / 9), solution rebuild 0 warnings, 100% diff
coverage on the changed production lines, and all 115 showcases regenerated and byte-identical.

jhaygood86 (approved):
Confirmed fixed. The follow-up commit correctly splits the two questions `SharesItsLineWithAFloat` was conflating: it now keeps only "does the running line total carry through this box," while the float branch explicitly hangs the trailing space (`maxSum -= trailingSpace; trailingSpace = 0;`) rather than treating it as content-following-the-space the way the flex-row branch's zeroing does.

Independently re-verified:
- `XY <span style="float:left">ZZZZ</span>` and its `nowrap` twin now measure 39.5859pt, matching real Chromium and Firefox.
- The new `InlineContentEitherSideOfAFloat_StaysOnOneLine` fixture (72.5742pt) — the one value you flagged as derived rather than browser-checked — checks out against both Chromium and Firefox exactly. Good catch adding that fixture; it's a genuinely distinct case my review didn't cover (I confirmed by mutation-testing that it would *not* have been caught by the first fix).
- Mutation-tested the new fixture against the pre-follow-up code: fails by exactly one un-hung space-width (6.5977pt), confirming it's testing what it claims to.
- Full rebuild: 0 warnings. Full suite: 11,078/0/9, matches your numbers. 100% diff coverage on the changed lines.
- Docs/recent-fix/migration-note/invariant files all correctly updated with the corrected numbers and the corrected "the wrapper's belief was right, not wrong" framing.
- CI green on all three platforms.

Nothing else needs changes — the earlier-verified §9.5/§9.2.1.1 citations, `IsInlineRunWrapper` plumbing, padding-fold accounting, and issues #1038/#1040 are all unaffected by this commit. Merging.

PR #1044, opened 2026-09-14
Align Acid2 rendering with Chrome

### LINKED ISSUE #1040: The intrinsic walk's padding total is neither an outer width nor the winning line's own padding

ChrisVanDijk:
Fixes #1040

## Summary

- size line boxes from CSS `line-height`, including first-line overrides and vertical writing modes
- correct block-end margin collapse, formatting-context float containment, absolute sizing, and HTML5 stray `</p>` handling exercised by Acid2
- keep intrinsic content widths paired with the decoration of the same candidate line
- paint stack-level-zero positioned descendants in document-tree order across positioning schemes
- preserve author-defined heights when descendants overflow while retaining the pinned flex/grid fragmentation fallback
- render the Acid2 showcase with zero page margins for browser-comparable geometry

## Validation

- `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0`: 11,404 passed, 9 platform skips
- diff coverage against `origin/main`: 99%
- `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings, 0 errors
- regenerated all 121 showcases and compared Acid2 with Chrome using PDFium and MuPDF at 96 and 192 DPI


jhaygood86 (changes requested):
This is excellent, well-evidenced work — I independently spec-checked and mutation-tested all five fixes (line-height line-box sizing against CSS 2.1 §10.8/§10.8.1, block-end margin collapse against §8.3.1, float containment against §10.6.7, the shrink-to-fit double-count against §10.3.7, the stray `</p>` handling against the WHATWG "in body" insertion mode, and the positioned-descendant paint order against Appendix E's step 8). Every spec citation checked out against real fetched text, every numeric claim reproduced exactly when I reverted the relevant production lines and reran the new tests, and the Acid2 face rasterizes identically through both PDFium and MuPDF. This is real, durable progress on spec compliance.

There are two things blocking merge, though — one a genuine regression, one a docs gap the PR's own changes opened.

## 1. Blocking: the new line-box sizing and word placement have drifted apart (causes the 9 macOS/Ubuntu CI failures)

CI is currently red — the exact same 9 tests fail identically on `macos-latest` and `ubuntu-latest`, while `windows-latest` is clean. That's not a flake pattern (a real flake wouldn't reproduce identically across two independent OSes); I had this root-caused and it's a genuine mechanism introduced by the first commit.

`LineBoxExtentOf` now correctly sizes a line's block-axis extent from `ActualLineHeight` (CSS 2.1 §10.8 — exactly right, this is the actual fix this PR set out to make). But individual **word placement, painting, and clip-reporting** still derive from the older "natural" glyph height (`RFont.Height`, sourced from OS/2 **win** ascent/descent) rather than the same line-height-based extent. Before this PR both quantities were identical, so they could never disagree; now they can, and on fonts where the gap between "natural" (win-metrics) height and "normal" line-height (hhea/typo-metrics, per issue #956's existing convention) is large enough — which happens to be true for Linux/macOS's default-font substitutes but stays under tolerance on Windows' real Arial — a word placed at its natural size overflows a box/line/fragment that was sized only to the smaller line-height value.

This is the root cause for all 9 failures, directly (`VerticalWritingModeLayoutIntegrationTests` × 3, both `FirstLinePseudoElementIntegrationTests` first-line-classification tests, `ClipReportTests`) and as a second-order pagination echo (`TableBorderPaint_LastPage_OuterBottomBorderIsDrawn`, `PulledRun_FromAPassThatResumedIntoAParagraph_ReEntersThatPass`, `TableSpannedBandRepetitionTests` — box heights leaking through into page-break math). None of these 9 fixtures were added or touched by this PR, and several already carry comments documenting prior font-metric hardening with margin this change has now exceeded — so this is a real magnitude increase from this PR's code, not latent flakiness surfacing coincidentally.

The line-height-based sizing itself is correct and should stay — the fix belongs on the word-placement side: either grow a word's placed rect to `Math.Max(natural height, LineBoxExtentOf)` the same way line/box thickness already does (so any overflow is symmetric on both sides of one computation), or reconcile `RFont.Height`'s metric source with `RFont.NormalLineHeight`'s for the `line-height: normal` case so they can't diverge in the first place.

## 2. Blocking: stale doc claim after closing #1040

The final commit's float-decoration rework (`inheritedDecoration`/`descendantDecoration` threaded through the nesting chain instead of `Math.Max`-reset) actually closes issue #1040 — I confirmed with a scratch repro of the gap file's own case (two padded floats sharing a line) that both paddings now count (58.20pt) instead of only the larger one (the old 50.20pt bug). The commit correctly deletes `.claude/accepted-gaps/intrinsic-padding-total-is-not-the-winning-lines-own-padding.md`, but per CLAUDE.md, closing a gap means deleting the file **and** removing the matching limitation note from `docs/**` — and `docs/html-css-support.md`'s `float` row still has the old sentence: *"Where two or more floats share a line and each declares its own padding/border, only the largest of those is counted in that measurement, so such a box can be sized short of what it draws."* That's no longer true and should come out.

## Non-blocking, worth a look but not required for this PR

- `HasElementInButtonScope`'s scope-boundary set is missing `select` from the WHATWG list — narrow and effectively unreachable given HTML's content model, so not worth blocking on.
- The extended `.claude/accepted-gaps/margin-collapsing-ignores-clearance.md` note (like the pre-existing entry it extends) doesn't carry its own tracking issue per CLAUDE.md's "genuine spec deviation must be tracked as an issue" convention — could use a follow-up issue at some point.

Build is clean (0 warnings) and diff coverage is 99% with the 2 uncovered lines being minor, low-impact edge cases (a defensive unreachable branch and an untested foster-parent corner case) — neither needs action.

Once the word-placement/line-box-extent split is fixed and the stale docs sentence is removed, this should be good to go — the underlying correctness work is solid.

jhaygood86:
Progress — the "Stabilize line-height tests across platforms" commit fixed 7 of the 9 failing tests, and its underlying claim (glyph ink may legitimately overhang its line box under negative leading, so word rects rather than line-box extents were the wrong thing to assert on) is a reasonable read of CSS 2.1 §10.8. But CI is still red on macOS/Ubuntu (Windows clean), and the 2 remaining failures are genuine geometry mismatches, not test-tolerance issues:

- `TableBorderPaint_LastPage_OuterBottomBorderIsDrawn`: the fixture now derives `lastPageIndex` from the real fragment count (`FragmentTree.Fragmentainers.Count - 1`) rather than `ActualBottom / pageHeight`, which resolved the earlier out-of-range crash — but the expected border position (Y=216.125) still isn't found among the painted lines (`[1.8, 39.9, ..., 195.9]`). The table's actual last-page geometry and the test's expectation still disagree.
- `VerticalRl_FloatLeft_NarrowsLaterColumnsAndNoWordOverlapsIt`: even under the loosened, arguably-more-correct "does the *line box* overlap the float" check (rather than individual glyph rects), the line containing `'Nu'` still genuinely overlaps the float on macOS/Ubuntu. That's not an assertion-strictness artifact — the column really isn't avoiding the float there.

Both still reproduce identically on macOS and Ubuntu with Windows clean, so this reads as the same underlying mechanism (or a closely related one) as before, just not fully resolved by the production-side behavior for these two cases. Holding the review as-is (changes requested) until CI is green — happy to take another look as soon as it is.

jhaygood86:
Latest commit ("Fix cross-platform line geometry") moved in the wrong direction — CI failures went from 2 to 6 on macOS/Ubuntu (still clean on Windows), same identical set on both platforms:

- The 2 from before are still failing: `VerticalRl_FloatLeft_NarrowsLaterColumnsAndNoWordOverlapsIt`, `TableBorderPaint_LastPage_OuterBottomBorderIsDrawn`
- 4 new ones appeared: `EarlyBreakLayoutIntegrationTests.PulledRun_FromAPassThatResumedIntoAParagraph_ReEntersThatPass` (the one the previous commit had specifically pinned a bundled font for and gotten green), `MulticolLayoutIntegrationTests.AnUnbreakableChildTallerThanTheBalancedTarget_EmitsAllOfItsContent`, `SuppressedPassFragmentainerTests.ATableInAMultiColumnContainer_EmitsCellContent`, `TableSpannedBandRepetitionTests.EveryBandOfASlicedRun_KeepsItsDecorations(tall: 246)` (also previously fixed by pinning `line-height:12pt` in that fixture)

Notably, two tests that the prior "stabilize" commit had already gotten green (via a pinned font and an explicit `line-height`) are failing again — suggesting this commit's line-geometry change has a broader footprint than the two it targeted. Given the pattern so far (each fix narrowing scope but touching more surface), it might be worth stepping back to look at whether `LineBoxExtentOf`/word-placement need a more structurally consistent fix rather than iterating case-by-case — happy to dig into the mechanism together if useful. Holding review as changes-requested until this is green.

ChrisVanDijk:
Running tests locally under Ubuntu right now: the final candidate passes five of the six CI failures in WSL; only the multicol case remains, and it also fails against the prior production baseline on this Ubuntu image. I’ll instrument that one fixture to compare each missing word’s line origin and captured column region, then fix the ownership mismatch rather than pinning another font.

ChrisVanDijk:
Apologies for the unnecessary CI churn. I misdiagnosed the remaining cross-platform failure and iterated on the assertion twice after the correct direction—using the repository's existing bundled font for this host-font-sensitive fixture—had already been identified. The final update pins the bundled font, gives it an explicit line height, and keeps the fixture clear of the separately tracked boundary-geometry limitation in #1048.

jhaygood86 (approved):
Green now, and I re-confirmed it's a real fix rather than a loosened one.

Production code (`src/PeachPDF/`) is byte-identical to `971e7f11`, the commit I already spec-checked and mutation-tested in full (line-height line-box sizing, block-end margin collapse, float containment, shrink-to-fit double-count, stray `</p>` handling, positioned-paint order — all confirmed against real spec text at that point). Everything since then has been test/docs work:

- The broad "Fix cross-platform line geometry" attempt that regressed 2→6 failures was cleanly reverted (`6d4f751f`) once it became clear it crossed too many subsystems (inline layout, table continuation, multicolumn capture, fragmentation) for this PR's scope — good call, and the honesty to back out rather than keep patching forward.
- The remaining two failures (`TableBorderPaint_LastPage_OuterBottomBorderIsDrawn`, `VerticalRl_FloatLeft_NarrowsLaterColumnsAndNoWordOverlapsIt`) are now fixed legitimately: pinned bundled fonts instead of relying on host defaults, and assertions derived from the real materialized fragment/line geometry (`FragmentTree`, `tableFragment.Lines`, `PageBreakBottoms`) instead of the whole-box arithmetic that was quietly assuming font-metric consistency it didn't have.
- The remaining structural gap (word rects/vertical-float exclusion/nested-fragment membership not yet sharing one geometry contract with the new line-height-based line boxes) is honestly captured in a new accepted-gap file and tracked as #1048, with a clear, specific "expected direction" for whoever picks it up — real issue, well-scoped, not a vague placeholder.
- The stale `docs/html-css-support.md` sentence I flagged about float-padding accounting is removed.

Own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings, 0 errors. `dotnet test --framework net8.0` — 11,398 passed / 0 failed / 9 skipped. CI green on all three platforms.

Merging.

Issue #1054, opened 2026-09-14
Negative leading overflows a line box downwards only, never upwards


ChrisVanDijk:
## Spec rule

CSS 2.1 [§10.8.1](https://www.w3.org/TR/CSS21/visudet.html#line-height): when a `line-height` is shorter than the font's own height the leading is negative, and half of it is subtracted from *each* side — the glyph content area then overflows its line box both above and below. Chrome does exactly that: a `10pt/5pt` line is 7 CSS px tall with its content area starting 5px **above** the line.

## Current behaviour

PeachPDF applies the full negative leading to the line's **height** (so `line-height: 5pt` really is a 5pt line, and the next line follows 5pt down) but holds the line's topmost ink at the line box's own top edge, so the overflow is downwards only.

The floor is applied **line-wide**, not per box — see `CssLayoutEngine.ApplyVerticalAlignment`'s `escape`, and the `Math.Max(0, …)` in `HalfLeadingOffsetOf` on the flow path — so every box on the line moves by the same amount and the baseline they share is preserved exactly; only the whole line's ink shifts down inside its own box.

## Why

This engine decides which fragmentainer a word belongs to from **the word's own rectangle** (`Fragmentation.FragmentEmitter.ClaimsWord`), not from the line box that owns it. Ink that escapes above its line box therefore escapes the fragmentainer the line was placed in. At a page or column boundary the consequences are not cosmetic:

- a line placed exactly at a page's content top has its words claimed by the page **above** it, or — their bottoms being past that page's own end — by *neither*, and they vanish entirely;
- a line box is a monolithic break unit ([css-break-3 §4.1](https://www.w3.org/TR/css-break-3/#possible-breaks)), so its content has to go with it.

This was not hypothetical: `ResumableInlineLayoutIntegrationTests.LineMixingTallAndShortContent_MovesWholeInsteadOfStrandingItsShorterWords` fails without the floor, with the line's text on page 0 and the image beside it on page 1.

Note that `line-height: 1` is already negative leading for most fonts (Arial's content area is about 1.15em), so this is a common configuration rather than an exotic one — which is why the failure mode had to be closed off rather than accepted.

## What closing it would take

Making **the line box**, not the word, the unit of fragmentainer membership: `ClaimsWord` and the `box.Words` walk around it would have to resolve a word's slot through the line that owns it. That is the same ownership problem #1048 describes from the other side, and it should probably be closed with that one rather than separately.

[...]

Issue #1053, opened 2026-09-14
Replaced and atomic inline content is not baseline-aligned


ChrisVanDijk:
## Spec rule

CSS 2.1 [§10.8.1](https://www.w3.org/TR/CSS21/visudet.html#line-height) aligns a baseline-aligned atomic inline by its **bottom margin edge**: an `<img>`, an inline `<svg>`, MathML, a form control or an `inline-block` sits *on* the line's baseline, with the text beside it sharing that baseline. For an `inline-block` with in-flow line boxes and `overflow: visible`, the baseline is its own last line's baseline rather than its margin edge.

## Current behaviour

Non-replaced inline boxes are baseline-aligned, with the leading split half above and half below the content area. Replaced and atomic content is left exactly where the flow put it — at the line box's **top** edge.

`CssLayoutEngine.FirstNonReplacedWordOf` is the single place that decides this: a box whose content on the line is entirely replaced (including an inline ancestor wrapping nothing else, so a `<span>`'s background cannot part company with the image inside it) gets no baseline offset, and `LineBoxContributionOf` correspondingly leaves such content out of the line's own baseline extent — it grows the line through `MaxBottom` at the call site instead.

Visible effect: `text <img style="height:20pt"> text` puts the image's top level with the text's line top, where a browser puts its bottom on the text baseline.

## Why it was left out of scope

The baseline-alignment change it came out of already reached line-box sizing, word placement, fragmentation membership and seven tests' expectations. Extending it to atomic inlines is separable and materially different work:

- An atomic inline's own words are flowed into the **same** line coordinates as its parent's, with `coordinates.CurrentY` pre-shifted by its border and padding (`ApplyAtomicInlineVerticalInsets`). There is no separate line box to take a baseline from, so "the baseline of its last line box" has to be reconstructed rather than read.
- Moving replaced content changes where every image, form control, MathML run and inline `<svg>` in the test suite and the showcases lands — a large, separate verification job, and a bad one to bundle with a change whose own correctness is argued from text positions.

Doing it half-way would be worse than not doing it: aligning text to the baseline while leaving an image on the same line at the top is neither the old behaviour nor the correct one.

`BaselineAlignmentLayoutIntegrationTests.ReplacedContentStaysAtTheLineTop` states the current behaviour directly, so closing this is a visible, deliberate edit rather than an accident.

Recorded in `.claude/accepted-gaps/replaced-inline-content-is-not-baseline-aligned.md`.

Issue #1048, opened 2026-09-14
Reconcile line, word, and nested-fragment geometry across font metrics


ChrisVanDijk:
## Problem

Line-box sizing now follows CSS 2.1 §10.8/§10.8.1, but word rectangles, vertical-float exclusion, and nested fragment ownership can still diverge when the resolved font's natural metrics differ materially from the used `line-height`.

This is exposed by Linux/macOS fallback-font metrics:

- vertical writing-mode float avoidance samples a column edge rather than the prospective full line-box span, so glyph/line geometry can overlap a float;
- a table continued inside a balanced multi-column fragment can leave the first resumed line outside the captured column band, so laid-out words are emitted nowhere;
- pagination-sensitive fixtures can change pass and page membership solely from host fallback-font metrics.

The attempted fix in PR #1044 broadened line/word geometry and fragmentation behavior and regressed additional table, multicolumn, and early-break cases. It was reverted so the Acid2-focused work can land independently; affected tests now use a bundled font and deterministic geometry.

## Expected direction

Use one structural contract for line-box extent, placed word overflow, fragmentation membership, and captured nested-fragment geometry. In particular:

1. Keep CSS 2.1 §10.8/§10.8.1 line boxes sized from used `line-height`, including negative leading.
2. Ensure word painting/clipping does not alter the line's fragmentation ownership.
3. Test vertical-float exclusion against the prospective full line-box block-axis span.
4. Ensure table/cell continuation inside a multi-column fragment resumes at and is captured within that nested fragmentainer's content band.
5. Add Linux/macOS coverage using deliberately different bundled font metrics, without relying on installed fonts.

## Why separate

This crosses inline layout, tables, multicolumn capture, fragmentation, and painting. It is substantially broader than PR #1044's Acid2 rendering scope and needs focused mutation and cross-platform testing rather than another incremental fix in that PR.

PR #1056, opened 2026-09-14
Share one baseline across a line box, and split its leading

ChrisVanDijk:
Found by rendering the `marker_styling` showcase against headless Chrome on the same HTML. Two cells were wrong — `::marker { font-size: 13pt }` left the marker's digits sitting below the text they numbered, and `list-style-position: inside` with a 20pt marker put the item's first line 12.8pt higher than Chrome's — and both turned out to be the same defect, with nothing to do with `::marker`.

## The defect

`CssLayoutEngine.ApplyVerticalAlignment` computed `baseline = Math.Max(rect.Top)`. The variable was named `baseline`; the quantity was the lowest **top edge** on the line, and every inline box was placed flush with it. Top-alignment and baseline-alignment coincide exactly while every font on a line is the same size, which is why it survived — and `line-height: normal` resolves from the same ascent + descent + gap that `RFont.Height` is built from, so the leading is ~0 there too and an ordinary paragraph never moved.

Three parts of [CSS 2.1 §10.8.1](https://www.w3.org/TR/CSS21/visudet.html#line-height) were missing, and they are one rule:

1. **Half-leading** — an inline box's content area is centred in its `line-height`, not hung from the top.
2. **A real baseline** — every box on the line hangs its own content area from one shared baseline.
3. **A line box is a pair of extents, not a height** — the two sides are maximised independently, so a line can exceed *every* `line-height` on it. Chrome makes `30pt/30pt` text containing a `10pt/40pt` span 56.328px tall, which a max-of-line-heights model cannot produce.

`LineBoxExtent` holds the pair, `LineBoxContributionOf` returns one inline box's share, and `CssLineBox.BaselineExtent` accumulates the line's. It is nullable deliberately: either side can legitimately be negative, and accumulating into a zeroed pair floored it at zero, which made a line with a short `line-height` taller than it had declared.

An outside `::marker` is not in its item's inline flow but does sit on that line's baseline, so it is placed there and unions its **ascent** into the item's first line. Ascent only, measured rather than assumed: in Chrome, `<li>Item` at 8.5pt with `::marker { font-size: 20pt }` takes the line's extent above the baseline from 10px to 24px while leaving the below side at the item's own 2px — the marker hangs outside the principal box, so its descender has nothing under it to push down. css-lists-3 §3.5 leaves both the alignment and the line-height interaction *expressly* undefined, so browsers are the reference rather than a rule.

## Two things that only showed up by running it

**Placing the ink after the line closed painted words twice.** Fragmentation decisions — `WouldStraddleFragmentainer`, `OverflowsEveryFragmentainer`, and the emitter's own `ClaimsWord` — are all asked of the word's own rectangle. Judging a word to fit at the line's top and only *then* moving it down by the half-leading put it across a column boundary after the decision had been made, and the emitter claimed it in both columns. `FlowBox` now applies the offset the moment the word is placed; `ApplyVerticalAlignment` re-derives every offset from the closed line and remains the authority.

**That broke an unstated invariant.** A good deal of this engine read a word's own top as though it were its line's — `DropATrailingForcedBreaksOwnLine`, and `CssLineBox.LineTop`, which the orphans/widows slot checks and fragment invalidation ask. 12 tests failed at once across four unrelated areas. `CssLineBox.FlowTop` now records the line box's own top as the flow places it, `LineTop` prefers it, and all 12 went green together.

## A pre-existing bug this uncovered

Fixing `vertical-align: top`/`middle`/`bottom` to align against the line box rather than the topmost ink exposed a **double application**: a `<td>`'s own `vertical-align` was being read both by the table algorithm (CSS 2.1 §17.5.3) and by this inline pass, because the styled-box walk stops at the first box carrying an `HtmlTag`, which for text directly inside a `<td>` is the cell itself. It was invisible while inline `top`/`bottom` were no-ops on a single-box line. The inline pass now reads a cell's value as `baseline`; `vertical-align` is not inherited, so a cell's value reaching that point can only ever have been the cell's own.

## Deliberately out of scope

Both are recorded as accepted-gap notes with the reasoning and the tests that state the current behaviour:

- **Replaced and atomic inline content** still sits at the line's top rather than with its bottom margin edge on the baseline — #1053. Doing it half-way would be worse than not doing it.
- **Negative leading overflows downwards only** — #1054. The floor is line-wide, so the shared baseline is preserved exactly and only the whole line's ink moves. Without it, a line landing at a page's content top has its words claimed by the page above — or, their bottoms being past that page, by neither, and they vanish.

#1054 is the same line-vs-word ownership problem #1048 describes from the other side, and should be closed with it. Neither is closed by this PR.

## Existing tests that changed

Seven asserted **glyph-ink** positions where they meant **line-box** positions — a distinction that did not exist while ink was flush with the line's top. `LayoutHarness.LineTopOf` is the shared helper they now use. One is a behaviour change rather than churn: `FragmentEmitterTests.InlineSpanningAPageBreak_KeepsCountingEdgesAcrossPages` now fits **two** lines on page 0 rather than three, because the third line box spans 60–90pt against a page ending at 80pt and so moves whole ([css-break-3 §4.1](https://www.w3.org/TR/css-break-3/#possible-breaks)). It used to be kept because its *glyphs* fitted.

## Charts.css label follow-up

A small-font chart value exposed one more boundary case in `LineBoxContributionOf`: when the word is owned by the line's block itself, the inline-ancestor walk must stop before that block's parent. It previously pulled a 12pt table-cell strut into an 8pt value line, shifting the value down inside its bar. `ColumnBarValue_StartsAtBarTop` guards the corrected placement.

## Verification

- After the Charts.css label follow-up: 42 focused baseline, flex-item, and first-line tests passed on each of .NET 8 and .NET 10.
- New `BaselineAlignmentLayoutIntegrationTests` (10 tests). Confirmed against neutered code that the 8 defect-targeting ones fail while 3 guards pass either way — the negative-leading floor, `line-height: normal` not moving, and replaced content staying at the line top.
- Every expectation was taken from headless Chrome on the same markup first. Values depending on the resolved font (which is not Chrome's here) are asserted as relationships, each preceded by the metric precondition that makes it non-vacuous.
- Full suite before the Charts.css label follow-up: 11,416 passed, 0 failed, 9 skipped (pre-existing platform skips).
- `diff-cover` against `origin/main`: 100% on the changed lines. Reaching it found real dead code (`BaselineAlignment.GetFirstBaselineY`, `CssLineBox.AboveBaseline`/`BelowBaseline`), deleted rather than covered.
- New `baseline_alignment` showcase, rasterized through **both** PDFium and MuPDF and compared against Chrome's own print-to-PDF of the identical HTML: the eight shaded one-line-box bands measure `[35.0, 35.0, 14.5, 35.5, 55.5, 40.5, 66.5, 33.5]` (PDFium) and `[34.5, 34.5, 14.0, 34.5, 55.0, 40.0, 66.5, 33.0]` (MuPDF) against Chrome's `[35.0, 35.0, 14.0, 35.5, 56.0, 40.0, 68.0, 34.0]`.


jhaygood86 (approved):
Independently verified this end to end.

**Core fix**: fetched CSS 2.1 §10.8.1 directly — the spec's line-box height rule really is "distance between the uppermost box top and the lowermost box bottom" after each box's half-leading/baseline placement resolves, not `max(line-height)`. For baseline-aligned boxes that reduces exactly to the "pair of independently-maximized extents" (`LineBoxExtent`) this PR implements, so the architecture is a correct spec reading, not an invented model. I reproduced both flagship numbers against real Chromium myself: the 30pt/40pt mixed-line-height case measures exactly 56.328125px (unreachable by any max(line-height) model), and the `::marker{font-size:20pt}` case shows the above-baseline extent growing 10→24px while below-baseline stays fixed at the item's own 2px — both match the PR's claims exactly.

**The fragmentation-ordering fix, the `FlowTop`/`LineTop` invariant fix, and the table-cell double-`vertical-align` fix** all check out — I mutation-tested the `FlowTop`/`LineTop` one myself and it broke 13 pre-existing tests across 7 classes when reverted, consistent with the PR's "roughly a dozen, several areas" claim.

**Minor, non-blocking documentation inconsistencies** in the PR's own self-reported verification (worth a look, don't block this): the body's "8 defect-targeting tests fail / 3 guards pass" doesn't quite square with the shipped `.claude/recent-fixes/` note's own "5 fail / 3 pass" for what reads as the same claim, and the "dead code deleted" mention (`GetFirstBaselineY`, `CssLineBox.AboveBaseline`/`BelowBaseline`) references symbols that don't appear to exist in either the merge-base or this branch — doesn't affect the actual diff or its genuine 100% coverage, just a stale or miscopied note.

**The follow-up commit** ("Keep direct text line boxes within their own formatting context") is a real, well-targeted fix — I mutation-tested it myself: reverting just that one line reproduces the exact claimed symptom (an 8pt flex value inflating to 15.75pt from its 12pt ancestor's font, instead of staying in the 8-12pt range its own text should produce).

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. `dotnet test --framework net8.0` — 11,424 passed / 0 failed / 9 skipped. CI green on all three platforms.

Merging.

PR #1059, opened 2026-09-14
Reuse SVG artwork across a document, and paint border-image slices correctly

ChrisVanDijk:
Two related pieces of work on how SVG artwork reaches the PDF: storing it once per
document instead of once per placement, and fixing the 9-slice painting that reuse
made visible.

## SVG artwork is cached in document-local Form XObjects

A parsed `SvgDocument` painted more than once — a `position: fixed` logo on every
page, an `` repeated down a list, a `background-image`, a
`border-image` source — used to be written into the PDF again at every placement.
`SvgRenderer.RenderCachedInto`/`GetOrCreateForm` now render it once into a Form
XObject and invoke that form wherever it appears.

The cache is keyed on the parsed document's identity plus the size and
`PixelsPerPoint` it is painted at, and lives in a `ConditionalWeakTable` keyed on
the owning PDF document, so a form never escapes the document it belongs to and a
finished document plus its forms can be collected. Identity rather than structural
equality is deliberate: two separately built inline SVGs can differ only in a
resolved `currentColor`. A graphics with no document to own a form — a measuring or
recording pass — still gets the individual drawing calls, as does artwork too small
for a form to be worth it.

## border-image painted the whole source into every one of its nine regions

`BorderImageDrawHandler` computes a source rectangle per region, but the backend
threw it away: `XGraphicsPdfRenderer.DrawImage(image, destRect, srcRect, unit)`
carried a `// TODO: incomplete - srcRect not used` and drew the entire image into
the destination. So every corner, every edge and every repeat tile painted the whole
texture squashed into its own band — a frame far denser and differently coloured
than the author wrote. The existing tests all asserted destination geometry, and the
recording mock did not record `srcRect` at all, so nothing caught it.

PDF has no operator for drawing a sub-rectangle of an image or Form XObject, so the
crop is composed in `GraphicsAdapter`: clip to the destination, then place the whole
image at the scale and offset that lands `srcRect` exactly on it. A whole-image
`srcRect` — what every background layer passes — short-circuits to the same
`XGraphics` call as before, so un-cropped draws emit byte-identical operators.
Nine-slice painting also turns interpolation off for its duration: clipping does not
stop a smoothing sampler reading half a source pixel past the cut, which at the
6x-and-up scales a slice-to-border stretch reaches was bleeding a quarter of the
border width.

Second defect in the same picture: an SVG `border-image-source` was treated as
sizeless and stretched over the border-image area before slicing, distorting the
artwork and giving every edge tile the wrong aspect. It is now sliced at its own
intrinsic size per CSS Images 3's default sizing algorithm; only a genuinely
sizeless SVG falls back to the area, as a gradient still does. A bare `<number>` in
`border-image-slice` counts CSS pixels on a vector or generated source, so
`ResolveSlice` takes the unit to count them in.

## Showcase

The 12-page document exercising all of this was a hand-written HTML file force-added
into `docs/showcase/`, a gitignored build-output directory. It now goes through
`SaveShowcaseAsync` like every other showcase and appears on the showcase page. Its
repeated SVG `border-image` is what makes it a real test of reuse: a border-image
invokes its source once per corner and once per edge tile, so one form is placed
dozens of times per page rather than once.

Measured by rendering the showcase's own generated HTML with the form cache switched
off, everything else identical: **24 Form XObjects and 176,077 bytes without reuse,
against 2 and 126,666 with it** — the no-reuse file is 39% larger. The saving is
smaller than the object count suggests because content streams are compressed and
this artwork is light; a heavier repeated illustration widens the gap.

## Verification

- Full suite on net8.0: 11,493 passing. Two `LineClampIntegrationTests` failures
  pre-date this branch and reproduce on a stashed tree.
- 100% diff coverage on the library changes.
- Rasterized through both PDFium and MuPDF per this repo's paint-verification
  convention: a synthetic 3×3 colour grid now puts each colour in its own region (it
  previously put all nine in each), the `border_image` showcase renders as an actual
  picture frame, and gradient frames run continuously around the border instead of
  repeating the whole gradient nine times.
- New tests cover the crop arithmetic and its emitted clip/placement operators,
  per-region source rectangles, SVG intrinsic sizing both ways, and the slice number
  unit. The recording mock now captures `srcRect` and the interpolation flag as of
  each call.

`docs/html-css-support.md` and `docs/architecture.md` updated; reasoning and traps
recorded under `.claude/recent-fixes/` and `.claude/invariants/`.


jhaygood86 (changes requested):
The actual feature work here is solid — I independently verified both pieces rather than trusting the description, since this is squarely the kind of PDF graphics-state change CLAUDE.md is strictest about:

- **SVG form reuse**: confirmed the cache is genuinely identity-keyed (`SvgDocument` has no `Equals`/`GetHashCode` override, so the dictionary key uses reference equality), confirmed it's scoped to a `ConditionalWeakTable` on the owning `PdfDocument` (safe across concurrent `PdfGenerator` instances, collectible once a document is done), and confirmed the measuring-pass/too-small-for-a-form fallbacks are real and tested. I built a standalone repro — two identical `fill="currentColor"` circles under different ancestor `color:` — and rasterized through both PDFium and MuPDF: both correctly show one red and one blue circle, no accidental form-sharing from the `currentColor` divergence this identity-keying is meant to guard against.
- **border-image crop**: hand-checked the placement-transform math against `ComputeCroppedPlacement` for a concrete example and it's correct; confirmed the whole-image-srcRect path is structurally a no-op (`XGraphicsPdfRenderer.DrawImage` never reads `srcRect` at all, so any whole-image draw is byte-identical to before); confirmed the interpolation toggle is scoped to the nine-slice paint only, not a permanent change. Built a synthetic 3×3 colour-grid border-image and rasterized through both renderers — all 8 border regions show their own distinct colour with no bleeding, which is the decisive visual proof this bug is actually fixed rather than just no longer showing up in a content-stream substring check.
- CSS Images 3's default-sizing algorithm is applied correctly for the SVG-intrinsic-size fix (checked against the fetched spec text).
- Mutation-tested both fixes independently; reverting either reproduces the exact symptom described.

**Blocking: `GraphicsAdapterImageCropTests.CroppedDraw_ClipsAroundTheDestinationBeforeInvokingTheForm` is locale-dependent and CI confirms it — failing on `windows-latest`/`macos-latest`, passing on `ubuntu-latest`.** `NewPage()` calls `document.AddPage()` with no explicit page size, and `PdfPage`'s default is `RegionInfo.CurrentRegion.IsMetric ? PageSize.A4 : PageSize.Letter`. The test hardcodes A4-derived coordinates (assuming an 842pt-tall page), so it only passes on a metric-locale machine. Two of three CI platforms are apparently resolving non-metric here. Fix is small — pass an explicit page size (A4, matching the test's existing assumption) into `NewPage()` rather than relying on OS locale.

Two minor, non-blocking notes, not worth holding up the fix above:
- The PR's byte-savings figure (176,077 → 126,666 bytes, 39% smaller) didn't reproduce exactly on independent re-measurement (I got closer to a 20% gap) — the Form XObject counts (24 → 2) matched exactly, so this looks like measurement-methodology drift (in-process vs. CLI-against-saved-HTML) rather than a wrong claim; the PR's own recipe note already flags this figure as re-derivable/drift-prone.
- Diff coverage measured at 97%, not the claimed 100% — two uncovered lines in `SvgRenderer.cs` (a zero-size-viewport guard and a null-tile-creation branch), both defensive edge cases, comfortably above the 90% gate.

Once the locale-dependent test is fixed and CI is green, this should be good to merge.

jhaygood86 (approved):
The locale-dependent test failure is fixed correctly — `page.Size = PageSize.A4` pinned right after `AddPage()`, plus a well-written invariant note explaining the trap (`PdfPage.Initialize()` follows `RegionInfo.CurrentRegion.IsMetric`, so an unpinned hand-built `PdfDocument` fixture silently depends on the machine's locale) so it doesn't resurface in a future test built the same way.

CI green on all three platforms, my own `dotnet build PeachPDF.slnx -t:Rebuild` (0 warnings) and `dotnet test --framework net8.0` (11,495 passed / 0 failed / 9 skipped) both clean. Combined with the earlier round's independent verification of both pieces of feature work — the SVG Form XObject reuse (identity-keyed cache, confirmed no `currentColor` cross-contamination via rasterization through PDFium and MuPDF) and the border-image slice cropping (hand-checked transform math, confirmed via a rasterized 3×3 colour-grid test that each region now paints from its own part of the source rather than the whole image) — this is good to merge.

PR #1061, opened 2026-09-14
Position a fixed box against the page area, and demote presentational attributes

ChrisVanDijk:
Two placement bugs, both found by putting a paginated document side by side with a browser rendering
the same markup: a fixed header sat on the paper edge instead of the content edge, a full-width rule
under it stopped short of the right margin, and an SVG logo rendered 37% oversized.

## A fixed box is positioned against the page area

A `position: fixed` box resolved its `left`/`top` from the **sheet corner**, while its width, height
and `left: 0; right: 0` fill all resolved against the page **area** (the sheet minus its margins).
Both halves cannot be right, and the disagreement was visible: a `left: 0; right: 0` rule was given
the content width but drawn from the paper edge, ending one margin short of the right content edge.

Measured out of the content stream on an A4 probe with `@page { margin: 38px 48px 42px }`:

```
before:  0 826.89 273.75 15 re    # `top: 0; left: 0` box, at the sheet corner
         0 745.89 523.28  1.5 re  # correct measure, drawn 36pt too far left
after:  36 798.39 273.75 15 re    # the content corner
        36 717.39 523.28  1.5 re  # 36 -> 559.28, exactly the measure
```

[CSS 2.1 §10.1](https://www.w3.org/TR/CSS21/visudet.html#containing-block-details) makes the page area
the containing block for fixed positioning in paged media ("the viewport in the case of continuous
media or the page area in the case of paged media"), which is also where a browser puts it when
printing. The origin is now the page's own content corner, and `FragmentEmitter`'s per-page correction
uses that slot's own resolved margins rather than the document's base ones.

Two of the ten tests that encoded the old convention needed more than new numbers:

- `FixedAbsoluteOffset_StaysTheSameDeltaOnEveryPage_RegardlessOfSizeOverrides` asserted that
  `left: 80pt` produced an identical rectangle on two pages with *different* margins. Under a page-area
  containing block that identity is wrong — the box is 80pt into each page's own area, which is two
  different distances from the sheet edge. Renamed and re-asserted accordingly.
- `FixedPercentOffset_ResolvesToEachPagesOwnArea_OnAMarginOnlyOverrideDocument` stopped proving
  anything: the centre of a page area with symmetric margins *is* the centre of the sheet, so its
  `left: 50%` fixture agreed across pages by coincidence. It uses 25%, which has no such symmetry.

## Presentational attributes no longer outrank author rules

`width`, `height`, `bgcolor`, `align`, `border`, `color`, `face`, `nowrap` and friends were applied
*after* the author stylesheet, so they beat every author rule — `<svg width="500" height="110">`
ignored a `.logo { width: 365px }` rule, and only an inline `style` could override it. HTML maps these
to declarations at the very start of the author origin with zero specificity, so any author rule wins.
They are now applied between the UA sheet and the author sheets, deliberately after the revert
snapshot so `revert` in an author rule still rolls back past them to the UA value.

Nothing in the suite depended on the old order, which is worth knowing before assuming the cascade is
a risky place to touch.

## Behaviour changes

Both change rendered output for documents that already worked, so `.claude/migration-notes/` carries a
note for release notes:

- Every fixed box moves inward by the page margins. A running header written as
  `position: fixed; top: 0` now starts at the top content edge; to put it back on the paper edge, use a
  negative offset or a zero page margin. On a document whose pages have different margins or sizes, a
  fixed box now follows each page's own area instead of holding one absolute position on every sheet.
- A stylesheet rule that was previously being silently ignored now takes effect — usually the intended
  result, and the same one a browser was already showing.

## Verification

- Full suite on net8.0: 11,499 passing. The two `LineClampIntegrationTests` failures reproduce on a
  clean branch off `main`, so they pre-date this work.
- Zero warnings from `dotnet build PeachPDF.slnx -t:Rebuild`.
- New tests state both rules directly: a fixed box's zero insets landing on the page area's corner, a
  `left: 0; right: 0` box spanning the full measure, and four presentational-hint cascade cases
  (attribute loses to a class rule, still wins where no rule touches the property, still loses to an
  inline style).
- The SVG reuse showcase is re-rendered through PDFium and MuPDF; its fixed header cluster is retuned
  in the last commit, having been authored against the old origin and the oversized logo.

`docs/html-css-support.md` updated (the `position` row and the Positioning section); reasoning and the
two traps above recorded under `.claude/recent-fixes/`.


jhaygood86 (approved):
Both fixes independently verified and correct.

**Fixed-position containing block**: fetched CSS 2.1 §10.1 directly — it says exactly what the PR claims, "the page area in the case of paged media," not "page" or "sheet." Reproduced the before/after content-stream numbers myself on an equivalent A4/asymmetric-margin probe and got matching values on both sides (sheet-corner origin on revert, page-area-corner origin with the fix). Worked through the two changed tests' reasoning by hand rather than taking it on faith: the old `FixedAbsoluteOffset_...` identity-across-pages assertion really was the bug once margins differ, and the `left:50%` fixture really was a coincidence that a page-area-centered model happens to share with a sheet-centered one under symmetric margins — 25% breaks that coincidence correctly.

**Presentational-attribute cascade order**: fetched the WHATWG HTML rendering spec directly — "author-level zero-specificity presentational hints" is the actual spec phrase, matching both the PR's framing and the repo's own code comments. Mutation-tested with a table `width` attribute vs. a low-specificity author rule and confirmed the attribute now correctly loses, still wins where no author rule touches the property, and still loses to an inline style either way.

Mutation-testing both fixes in isolation reproduces the exact claimed symptoms when reverted.

One thing worth a quick look, not blocking: `CssBox.cs`'s `CommitBlockChildOffset` derives its origin from the document's base `@page` margins rather than the specific page slot's own resolved margins (inconsistent with the width/height basis in the same method, which does use the slot's own margins). This doesn't produce wrong output — `FragmentEmitter.ComputeFixedPageOffset` independently re-derives the correct absolute position per slot and applies it as a delta to every slot including slot 0 — and it's already covered by a passing test (`FixedPercentOffset_ResolvesToEachPagesOwnArea_OnAMarginOnlyOverrideDocument`), so it's latent/cosmetic rather than a live bug.

The PR's claim that the two `LineClampIntegrationTests` failures are pre-existing didn't reproduce across several runs on either `main` or this branch — worth a note for next time (this is the same shape of claim that turned out to be a real, PR-specific bug in #1059), but nothing here suggests this PR's own tests are hiding anything; the full suite is clean on both my own run and CI.

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. `dotnet test --framework net8.0` — 11,501 passed / 0 failed / 9 skipped. CI green on all three platforms. Rasterized the retuned SVG-reuse showcase through both PDFium and MuPDF — logo correctly sized and positioned inside the page margins, no clipping.

Merging.

PR #1063, opened 2026-09-14
Draw a block's text decoration over its inline content

ChrisVanDijk:
The bookmarks showcase underlines its table-of-contents entry all the way to the right page margin. The entry is `a.toc-entry { display: block }`, and paint drew text decoration over each of a fragment's decoration rectangles. For an inline box those are its per-line rectangles — the text — but a block-level box has exactly one, its own border box, which is the right area for a background and a border and the wrong one for a decoration line.

[css-text-decor-3 §2.4](https://www.w3.org/TR/css-text-decor-3/#line-decoration) does not decorate the block itself: a block container's decoration propagates to an anonymous inline box wrapping its in-flow inline-level content, and through its in-flow block-level descendants to theirs. So the geometry needed is the inline content's, one span per line box.

## The change

`FragmentPainter.PaintPropagatedDecoration` handles the one-rectangle case, recognized at paint time as `lines is [{ Line: null }]` — only the own-bounds branch emits a `LineFragment` with no `CssLineBox`. It walks the fragment's children collecting one union rectangle per line box and draws a line over each.

- **The spans come from the fragment tree, not the box tree**, so a block broken across pages decorates exactly the lines that landed on the page being painted, with no coordinate mapping — fragment rectangles are already fragmentainer-local.
- **A descendant hosted on a line box contributes its rectangles and is not descended into.** Its rectangle already covers its content: an inline box's own, including its padding and border; an atomic inline's whole border box, which §2.4 draws the line across without propagating into its contents. Anything else is a block whose own rectangle is a border box again — the thing this avoids reading — so it is descended through.
- **Out-of-flow descendants are skipped**, per the same section.
- **`PaintDecoration` gained `ownDecorationArea`.** A propagated span is inline content geometry, already inside the box's padding; the padding/border insets and the `bottomInset` compensation exist for a rectangle measured from the box's *border* edges and would displace a span. That is what makes an underlined block with padding correct rather than merely different.
- The subtree walk is gated on the box declaring a decoration at all (its own or its `::first-line`'s), so the overwhelming majority of blocks never start it.

## What a document author sees

- A block wider than its text is underlined only under the text; a wrapped block draws one line per line box instead of one across the whole block.
- The span follows `text-align`, so centred and right-aligned blocks are decorated where their text sits.
- Floats and absolutely positioned descendants are not decorated, and a block with no inline content draws no line at all where it used to draw a full-width one.

## Found by running it

An opaque atomic inline (an ``, an inline-block with a background) **covers** the propagated line, because the line is painted with the block and the atomic inline paints later. With the background removed the same line is continuous underneath, so this is CSS 2.1 Appendix E's existing paint order showing through rather than the new geometry — deliberately not changed here.

Union-of-tops for a span's y-position was the other worry: a tall inline-block or a baseline-aligned image drags the union's top upward and could lift the line. Rasterizing both cases put the line at text level, so no per-descendant special case was added.

## Verification

- 8 new tests in `TextDecorationPaintIntegrationTests.cs` — span extent vs. the block's width, one span per line box, several inline children on one line unioning into a single span, padding, an in-flow block descendant, a skipped float, an empty block, and a block with no decoration. All 19 in the file pass.
- Full net8.0 suite: 11546 passed, 2 failed — both `LineClampIntegrationTests`, failing identically on `main` with this change stashed.
- `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings.
- Diff coverage: every changed line is hit.
- Rasterized through both PDFium and MuPDF (they agree): the bookmarks fixture, plus wrapped, centred, right-aligned, padded, nested-block, floated, atomic-inline, image and empty cases.


jhaygood86 (changes requested):
The core fix is correct and well-verified — I independently checked it against css-text-decor-3 §2.4's actual propagation mechanics and rasterized six scenarios through both PDFium and MuPDF (the TOC/bookmark repro, a wrapped block, centered/right-aligned text, a padded block, a block with float/absolute siblings, and the atomic-inline-coverage case) — all match the PR's description exactly. Mutation-testing confirms the fix: reverting reproduces the full-block-width underline symptom on 6 of the 8 new tests (the other 2 don't discriminate old/new behavior in their current fixtures — harmless, but worth tightening at some point, non-blocking).

**Blocking: the new `.claude/recent-fixes/2026-09-14-block-text-decoration-propagates-to-inline-content.md` misstates css-text-decor-3 §2.4.** It says an atomic inline's "whole border box... §2.4 draws the line across without propagating into its contents." The actual spec text is unambiguous: *"Atomic inlines, such as images and inline blocks, are not decorated"* — no line at all, not "drawn across but not propagated into." I confirmed by rasterizing an atomic inline with `opacity:0` (rather than an opaque background) that PeachPDF's decoration line genuinely runs straight through the atomic inline's geometry (verified against the raw content stream, not just a screenshot) — it's normally invisible only because most atomic inlines paint opaque content on top of it. This is a real, pre-existing spec deviation, not something this PR introduces or worsens (identical on `main`) — but since this PR's own new doc file is the one making the (incorrect) claim about it, and CLAUDE.md is explicit that "an accepted gap that is a genuine spec deviation must be tracked as an issue, not only recorded as prose," this needs two small things before merge:
1. Correct the inaccurate sentence in the recent-fixes note.
2. Add a proper `.claude/accepted-gaps/` file for "atomic inlines are drawn through by a decoration line rather than excluded" with a filed tracking issue, since this is exactly the kind of gap CLAUDE.md's convention exists for — otherwise it'll just quietly disappear when the recent-fixes note expires in 30 days with no permanent record.

Everything else — the propagation-to-inline-content mechanism itself, the out-of-flow exclusion, the `ownDecorationArea` padding-inset fix, the decoration-gate short-circuit, the nested-decorating-box color-propagation behavior (which I separately confirmed matches css-text-decor-3's own worked example almost exactly) — checks out. Full suite, build (0 warnings), diff coverage (100% on changed lines), and CI are all clean; the `LineClampIntegrationTests` pre-existing-failure claim in the PR body didn't reproduce on either `main` or this branch in a clean full-suite run, similar to a couple of recent PRs — worth being more careful about that specific claim going forward, but not blocking here since nothing points to it masking an actual regression in this PR.

Once the doc correction and accepted-gap/issue are in, this is good to merge.

jhaygood86 (approved):
The follow-up commit is exactly right, and better than I asked for. It corrects the inaccurate spec claim in both places it appeared (the recent-fixes note and the code's own doc comment on `PaintPropagatedDecoration`), and files a genuinely well-reasoned accepted-gap file — I checked, issue #1066 is real and open. It correctly identifies this as pre-existing (the old full-width block decoration ran through atomic inlines the same way before this PR's propagation mechanism existed) rather than something this PR introduces, cross-references the related already-filed `text-decoration-skip-ink` gap (#1064, also verified open) since both need the same "subtract from the span" mechanism, and even flags the two specific details a future fix will need to get right (margin box vs. border box; recognizing an atomic inline by display type across two different layout paths). The docs update states the limitation without an issue citation, matching the repo's split between reader-facing docs and internal tracking.

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. `dotnet test --framework net8.0` — 11,549 passed / 0 failed / 9 skipped. CI green on all three platforms.

Merging.

PR #1068, opened 2026-09-14
Fix line-clamp fixture width in comma-decimal locales

ChrisVanDijk:
When the test runner uses a comma as its decimal separator, BuildFixtureNeedingExactlyOnePopAsync interpolates its measured width into an invalid CSS length such as width:75,873pt. The paragraph then stays wide enough that the line-clamp tests never exercise the intended word-pop path.

Format the computed width with InvariantCulture so the fixture remains valid CSS across locales while retaining its font-dependent measurement.

Validation: dotnet test src/PeachPDF.Tests/PeachPDF.Tests.csproj --no-restore --filter "FullyQualifiedName~LineClampIntegrationTests" (15 passed on net8.0 and 15 passed on net10.0).


jhaygood86 (approved):
Correct, minimal fix. `ToString()` without an explicit culture uses the current thread's culture, so under a comma-decimal locale the interpolated width produced invalid CSS (`width:75,873pt`) — the paragraph stayed too wide for the fixture to ever exercise the intended word-pop path, which is exactly why this test's occasional CI failures kept getting (correctly) waved off as unrelated across the last several PR reviews in this queue. `ToString("R", CultureInfo.InvariantCulture)` fixes it while preserving the round-trip precision of the font-dependent measurement.

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. Full suite on net8.0 — 11,543 passed / 0 failed / 9 skipped, `LineClampIntegrationTests` alone — 15/15 passed. CI green on all three platforms.

Merging.

Issue #1075, opened 2026-09-15
text-decoration is drawn horizontally across the top of a vertical-writing-mode column instead of along the inline axis


ChrisVanDijk:
### Summary

Under `writing-mode: vertical-rl` / `vertical-lr`, a `text-decoration` on a block box is painted as a short **horizontal** stroke across the top of the first column, instead of running alongside the column parallel to the inline (text-flow) axis.

### Reproduction

```html
<div style="writing-mode: vertical-rl; height: 300pt; font: 20pt serif;
            text-decoration: underline">Vertical text with an underline</div>
```

**Expected** (browser behaviour): the underline runs vertically down the length of the column, on one side of the glyphs.

**Actual**: a short horizontal line is drawn across the top of the column, crossing the first glyph. Verified by rasterizing the generated PDF with PDFium.

### Why

`FragmentPainter.PaintDecoration` derives the decoration geometry entirely on the x-axis — the span is `new DecorationInterval(rectangle.X, rectangle.Right)` and each keyword picks a single `y`, then strokes `g.DrawLine(pen, segment.Start, y, segment.End, y)`. Under a vertical writing mode a box's physical x-range is the column's *thickness*, not its extent along the line, so both the position and the orientation are wrong.

### Spec

[css-text-decor-3 §2](https://www.w3.org/TR/css-text-decor-3/#line-decoration) defines underline/overline relative to the text's own baseline, and §2.1 notes that "In [vertical writing modes](https://www.w3.org/TR/css-writing-modes-4/#vertical-writing-mode), `text-underline-position` can cause the underline and overline to switch sides" — i.e. the decoration is expected to be laid along the inline axis of whatever writing mode is in effect, with underline/overline on opposite sides of it.

### Scope note

This is a pre-existing gap, not a regression. It is called out here because the `text-decoration-skip-ink` (#1064) and atomic-inline-exclusion (#1066) work deliberately guards both of its new subtractions behind a horizontal-writing-mode check (`FragmentPainter.PaintDecoration`'s `IsHorizontalWritingMode`): under a vertical mode those subtractions would remove the decoration entirely rather than break it, since they are also x-axis reasoning. That guard keeps the new rules from making this worse, but does not fix the underlying orientation problem.

Related vertical-writing-mode gaps: #769 (nested inline box insets, which also affects the decoration box), #771, #796.

PR #1076, opened 2026-09-15
Implement text-decoration-skip-ink and break decoration lines around atomic inlines

### LINKED ISSUE #1064: text-decoration-skip-ink is unimplemented: underlines run through descenders
`text-decoration-skip-ink` (css-text-decor-4 §2.5) is not implemented — the property is not in `css-properties.json`, is not mentioned in `docs/html-css-support.md` in either the supported table or the unsupported list, and `FragmentPainter.PaintDecoration` draws one unbroken `DrawLine` per decoration span. Every underline and overline therefore runs straight through the descenders of `g`, `j`, `p`, `q`, `y` and through `f`/`t` at overline height, where a browser (and the property's `auto` initial value) leaves a gap.

This is an assessment of what implementing it would take, written up so the work doesn't have to be re-derived. Estimate: **medium, roughly a day. No research risk — both hard prerequisites already exist in the codebase.**

## Why it is not hard

The expensive part of skip-ink is normally getting real glyph ink positions. That already exists and is already shipping:

- `GraphicsAdapter.GetTextOutline` (`src/PeachPDF/Adapters/GraphicsAdapter.cs`) walks `descriptor.Shape(str, features)`, applies each `ShapedGlyph`'s GPOS `XOffset`/`YOffset`, decodes the outline through `GlyphOutlineDecoder.TryGetGlyphOutline`, and maps design units into user space at the correct scale. SVG text-as-outline (`SvgRenderer`) already depends on it, so the positions are known-correct against what `DrawString` actually paints.
- `GlyphOutlineDecoder` already returns closed contours of lines and cubics in design units, y-up, with quadratics already elevated to cubics — so a flattener has one segment kind to handle.
- A word's baseline is recoverable from what paint already has: `DrawString` renders the baseline at `point.Y + font.Ascent`, and `FragmentPainter.PaintWordSequence` has each word's fragment rect and `baselineAdjust`.
- The "which content does this decoration line cover" walk now exists as `FragmentPainter.PaintPropagatedDecoration`, which already descends the fragment subtree skipping out-of-flow boxes.

## The one genuinely new routine

Given a band (the decoration line's thickness, dilated by the skip gap), return the x-intervals where glyph ink crosses it: flatten the cubics, scanline the contours with nonzero winding at a handful of y values across the band, union the intervals, dilate by the gap, merge, subtract from the span, drop slivers, draw the remaining segments. Roughly 100 lines, unit-testable without producing a PDF.

## Work breakdown

| Piece | Size |
|---|---|
| Property plumbing — one `css-properties.json` entry (inherited, initial `auto`, `auto \| none \| all`), storage on `CssBox`, a `docs/html-css-support.md` row | small; the source generator does the dispatch |
| `RGraphics.GetInkCrossings(...)` — **virtual**, defaulting to an empty result so the existing `RGraphics` test mocks need no edit and simply degrade to "no skipping"; real implementation in `GraphicsAdapter` reusing `GetTextOutline`'s walk | medium |
| Painter integration — collect the `TextFragment`s under a decoration span, map each to (text, font, baseline, letter-spacing, shaping features), subtract the intervals, draw N segments instead of one `DrawLine`. Needs doing for the inline per-line path as well as the block-propagated one | medium |
| Tests — interval math at the adapter, `TestRecordingGraphics` segment-count/extent assertions at the painter, plus PDFium + MuPDF rasterization per this repo's paint-verification convention | medium |
| Showcase, docs, migration note | small |

## Sharp edges to know before starting

- **CFF/OTF outlines are not decodable here.** `TryGetGlyphOutline` reads `glyf`/`loca` only, so a CFF-outline font yields no ink and nothing to skip. `auto` is defined as UA discretion, so declining to skip is conformant — but `text-decoration-skip-ink: all` would silently not skip. That is an accepted gap with its own tracking issue, not a blocker.
- **Per-codepoint font fallback happens below the adapter.** `GetTextOutline` does not follow it either, so a codepoint painted by a fallback face would get no skip. Same limitation the SVG outline path already has — consistent, and worth documenting rather than fixing here.
- **`auto` must skip to match browsers, which changes the default rendering of every underline in every document.** That is the real cost of this change, and it is verification cost rather than code: a migration note plus a broad two-renderer raster sweep.
- `line-through` is never skipped.
- Vertical writing mode wants a vertical band and should stay out of scope (it is adjacent to #769).
- Performance: an extra shape + outline decode per decorated word. Gate it on the box actually declaring a decoration, the same way the propagated-decoration subtree walk is gated.

### LINKED ISSUE #1066: A text decoration line is drawn through atomic inlines, which css-text-decor-3 §2.4 says are not decorated
## The gap

A text decoration line propagated to a line box is drawn continuously across any atomic inline on that line — an ``, an `inline-block`, an `inline-table`. [css-text-decor-3 §2.4](https://www.w3.org/TR/css-text-decor-3/#line-decoration) says it should not be:

> Atomic inlines, such as images and inline blocks, are not decorated.

The line should break around the atomic inline's whole margin box and resume after it. It does not; it runs straight through.

## Evidence

The deviation is normally invisible, because the common atomic inlines paint opaque content exactly where the line wrongly is — an image or a background-carrying inline-block covers it, so a rasterized page shows the gap the spec asks for and hides the reason it is there. Giving the atomic inline `opacity: 0` and reading the raw content stream exposes it:

```html
<div style="text-decoration:underline">AA <span style="display:inline-block;width:60pt;opacity:0">hidden</span> BB</div>
```

with `@page { size: a4; margin: 10mm }` and `body { font: 11pt Arial, sans-serif; margin: 0 }` produces exactly one line segment in the page content stream:

```
28.3465 802.89 m
123.8064 802.89 l
```

28.35pt is the content-left edge; the segment is 95.5pt long, which is `"AA "` + the full 60pt inline-block + `" BB"`. One unbroken line across the atomic inline, where the spec wants two segments with a 60pt gap.

## Status

Pre-existing and unchanged by #1063, which made a block container's decoration cover its inline content instead of its own full width. Before that change the block's own full-width rectangle ran through the atomic inline in exactly the same way, so #1063 neither introduces nor worsens this — it inherits it. Recorded as an accepted gap (`.claude/accepted-gaps/decoration-line-runs-through-atomic-inlines.md`) rather than fixed there, to keep that PR to the propagation mechanism it was about.

## What fixing it would involve

`FragmentPainter.CollectDecorationSpans` currently unions every line-hosted descendant's rectangle into one span per `CssLineBox`. An atomic inline would instead need to be *subtracted* from the span, turning one span into several segments — the same "one decoration line becomes N drawn segments" shape that `text-decoration-skip-ink` (#1064) needs, so the two are worth building on one mechanism rather than separately.

Two details not to lose when doing it:

- The exclusion is the atomic inline's **margin** box, not its border box; the span rectangles currently collected are border boxes.
- An atomic inline is identified by its own display type, not by whether the fragment happens to have children — an `inline-block` whose content is inlines-only reaches paint through a different layout path than one whose content is block-level (`CssLayoutEngine.FlowAtomicBlockContentChild` vs. the ordinary inline path), and both must be excluded.

ChrisVanDijk:
Implements `text-decoration-skip-ink` and stops decoration lines being drawn through atomic inlines.

Fixes #1064
Fixes #1066

## What changes

**Underlines and overlines now skip glyph ink.** `text-decoration-skip-ink` is parsed as an inherited longhand (`auto` | `none` | `all`, and correctly *not* part of the `text-decoration` shorthand), and `auto` — the initial value — breaks the line around descenders. The ink is measured for real: each run is shaped, each glyph's outline decoded, and a scanline pass over the nonzero winding finds where the glyph actually crosses the decoration band. `line-through` is never skipped, per css-text-decor-4 §2.5. The horizontal clearance around crossed ink grows with the resolved decoration thickness (up to Chromium's 13px cap), so heavy underlines leave the wider gaps browsers do.

**Decoration lines break around atomic inlines.** css-text-decor-3 §2.4 says atomic inlines are not decorated; previously a line was drawn straight through an `` or `inline-block`. One decoration line is now a list of segments with the atomic inlines subtracted.

**Thick underlines now grow away from the baseline.** The underline position accounts for the resolved stroke thickness, keeping its top edge below the alphabetic baseline instead of expanding half the stroke upward through ordinary glyphs. The position is anchored to the exact fractional baseline used by PDF text painting rather than the rounded ascent used for layout. This preserves the intended visible gap and prevents `text-decoration-skip-ink: auto` from breaking a heavy underline merely because the line itself was positioned too high, matching Chrome's continuous 6px result in the existing `text_decoration_thickness` showcase.

**The skip shape is per glyph, not per ink run.** A glyph that puts several disjoint runs of ink in the band — the two sides of an `o`, the bowl and tail of a `g`, the three dots of U+2026 — is hulled into a single break. css-text-decor-4 §2.10.5 leaves the skip shape to UA discretion and names this exact choice, noting that hiding the line within a glyph's enclosed areas "gives a cleaner look to the type" and that following each contour can leave "typographically-awkward wisps of underline" — which is precisely what per-run spans produced. Chrome and Firefox were both measured on `o`, `g`, `n`, `v`, `H` and U+2026 (by diffing the red decoration pixels between `skip-ink: auto` and `none`, since glyphs paint over the line and hide the truth); both hull per glyph, U+2026 included. As a side effect the merged runs emit far fewer strokes — a 20 000-word fully-underlined fixture dropped from 790 KB to 438 KB.

## Design notes

The ink measurement is cached per (font, text, band-relative-to-baseline, letter-spacing, shaping features), with the run's absolute position factored out, so the same word on a later line is a cache hit rather than a second full shape-and-decode. Without it this feature cost +57–80% on a fully-underlined document, since it re-shaped text `DrawString` had already shaped; with it, +13%.

The hull is applied in `GraphicsAdapter.MeasureInkCrossings`, deliberately not in `GlyphInkScanner`. The scanner stays an honest, unit-testable report of where a glyph is painted, and a glyph's own two sides are only distinguishable from two adjacent letters during the glyph walk.

The band is anchored to the font's unrounded ascent, computed the same way `XGraphicsPdfRenderer.DrawString` does, rather than the rounded `RFont.Ascent` — an underline's band is one unit tall at the default thickness, so a rounded anchor could shift it by half its own height and flip whether a grazing descender is skipped.

## Known limitations

Both are documented in `docs/html-css-support.md` and recorded under `.claude/accepted-gaps/`:

- Fonts with CFF outlines report "ink unknown" rather than "no ink", so nothing is skipped — `auto` is UA discretion, so declining to skip is conformant.
- Skip-ink and the atomic-inline break are both gated to horizontal writing modes. Vertical decoration geometry is a pre-existing gap, tracked separately in #1075.

## Verification

Full suite green (11 658 passing, 0 warnings on a solution rebuild). **100% diff coverage** across the branch — including the cross-glyph union arm, which nothing reached until a fixture was added that overlaps two glyphs' spans.

Per this repo's conventions the tests assert on the ordered sequence of `RGraphics` calls and real box geometry rather than PDF content-stream substrings, and the output was rasterized through both PDFium and MuPDF. The new `text_decoration_skipping` showcase covers the feature, and the existing `text_decoration_thickness` showcase confirms heavy underlines remain continuous.


jhaygood86 (approved):
Ambitious feature, and it holds up. This is real glyph geometry, not an approximation — I independently confirmed `GlyphInkScanner` decodes actual `glyf`/`loca` TrueType outlines and does genuine nonzero-winding scanline crossing detection (verified the nonzero-winding claim specifically: a reversed-wound inner contour correctly reads as a hole, not ink). The most consequential empirical claim in the PR — that per-glyph hulling of disjoint ink runs (both sides of `o`, bowl+tail of `g`, all three dots of `…`) matches real Chrome's behavior — was independently reproduced against actual Chromium rendering, not taken on the PR's word: all six named glyphs (`o`, `g`, `n`, `v`, `H`, U+2026) collapse to one hulled gap under `skip-ink:auto` in real Chrome, exactly as claimed. Mutation-testing both the ink-crossing return path and the glyph scanner itself reproduces the exact claimed symptom when disabled.

Also independently verified: `line-through` is genuinely exempted from skipping (matches css-text-decor-4's real prose, quoted accurately even though — see below — the section numbers citing it are wrong); the ink-measurement cache key genuinely excludes absolute position and is genuinely invalidated by a different decoration band; the atomic-inline margin-box exclusion is correctly documented as this engine's own implementation choice rather than a spec mandate (checked against the actual spec text, which says nothing about which box bounds an atomic inline's exclusion); vertical-writing-mode gating is real and tested; and the thick-underline baseline fix produces a visually continuous line under ordinary glyphs at 6px thickness with only real descender-driven gaps, confirmed by rendering the actual `text_decoration_thickness`/`text_decoration_skipping` showcases through both PDFium and MuPDF.

Non-blocking cleanup worth a follow-up, none of it affecting correctness:
- The css-text-decor-4 section citations throughout (code comments, test comments, and the user-visible showcase label) say "§2.5" where the actual property section is §2.10.4/§2.10.5/§2.10.6 depending on which specific claim — the quoted prose and spec anchor links are accurate, just the section numbers are off. Worth fixing since one instance is user-visible in the showcase.
- `.vscode/settings.json` is an empty file with no real content — safe to drop.
- A stray UTF-8 BOM landed on two unrelated pre-existing files (`PropertyNames.cs`, `Map.cs`) as an incidental side effect of editing them.
- The recent-fixes note mentions a separately-surfaced bug ("an empty or inlines-only `inline-block` ignores its declared width") as "worth filing on its own" — I don't see a corresponding issue filed yet; worth doing before that note expires in 30 days.

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. `dotnet test --framework net8.0` — 11,661 passed / 0 failed / 9 skipped. Diff coverage confirmed 100% including the cross-glyph-union arm. CI green on all three platforms.

Merging.

PR #1092, opened 2026-09-15
Remove collapsible white space at the beginning of a line

### LINKED ISSUE #1087: Collapsible whitespace after <br> indents the beginning of the new line
### Summary

Collapsible source whitespace immediately after a `<br>` is laid out as a leading space on the new line. Browsers remove collapsible whitespace at the beginning of a line, so formatting the HTML source across lines must not indent the rendered content.

This was exposed while retesting #31 after its inline-block width problem was fixed by #951. It is independent of that width fix.

### Reproduction

This is the exact sample from #31:

```html
<div style="display:inline-block; color:red; width:450px;">Field name:</div><div style="display:inline-block;">short</div>
<br/>
<div style="display:inline-block; color:red; width:450px;">Long field name:</div><div style="display:inline-block;">long</div>
<br/>
<div style="display:inline-block; color:red; width:450px;">Very long field name:</div><div style="display:inline-block;">very long</div>
```

Compare it with the same markup minified so that no source newline follows either `<br>`.

### Expected

All three label boxes begin at the same x-coordinate, and all three value boxes begin at the same x-coordinate.

Chromium gives identical positions for the multiline and minified forms. With its default 8px body margin and each label at `width: 450px`, every value begins at `x = 458px`.

CSS Text Module Level 3's [Phase II white-space processing](https://www.w3.org/TR/css-text-3/#white-space-phase-2) removes a sequence of collapsible spaces at the beginning of a line.

### Actual

PeachPDF retains the collapsed newline after each `<br>` as one leading space. In the multiline sample, the second and third rows are shifted right by approximately **3.013pt** relative to the first row. In the minified form, all three rows align exactly.

The inline-block widths themselves are now correct: each value starts 337.5pt (450 CSS px) after its corresponding label. The defect is that PeachPDF offsets the entire post-`<br>` line before placing that label.

ChrisVanDijk:
Fixes #1087.

[css-text-3 phase II](https://www.w3.org/TR/css-text-3/#white-space-phase-2) removes a sequence of collapsible spaces at the **beginning** of a line. `CssLayoutEngine.FlowBox` had no notion of where a line began, so the source newline between a `<br>` and the element after it was laid out as a leading space — indenting the whole line the `<br>` opened by one space (3.013pt at the default 12pt font). Formatting the source across lines changed the rendered output; the same markup minified laid out correctly.

## The load-bearing idea

**`Line.Words.Count == 0` is the test that looks right and is wrong.** A forced-break word is placed on the line it *opens*, not the one it closes: the wrap branch closes the current line, creates the new `CssLineBox`, stamps its `FollowsForcedBreak` from that same word, and only then positions it and calls `ReportExistanceOf` — on the **new** line. So after a `<br>` the line already holds one word, and every "is this line empty?" question answers "no" for exactly the line those questions are about.

`IsAtLineStart` asks it correctly: no word on the line that isn't a zero-width break marker, and no `Line.Rectangles` entry (an atomic inline-level box contributes one and no word at all).

## Three defects, one root cause

Fixing the reported one surfaced two more. Both were confirmed by rendering before being touched, and both are fixed here.

**1. The collapsed space was rendered at the line start.** A collapsed space is never a word here — it survives as a cursor advance added at two points in `FlowBox`: the whitespace-only-box tail (`<br>`↵`<div>`) and the `DomUtils.IsBoxHasWhitespace` branch (`<br><span> b</span>`, where the space sits inside the following inline). Both shapes reproduce; fixing only the first leaves the second indented.

**2. A forced break counted as a justification opportunity.** `CssRectWord.IsSpaces` is "every character is whitespace" and the marker's text is a newline, so `IsSpaces` *and* `IsLineBreak` are both true of it. `IsJustificationOpportunity`'s `previous.IsSpaces` arm — written for *preserved* white space under `pre`/`pre-wrap` — counted it, so `text-align: justify` spent an expansion share at the head of every line a `<br>` opened. Measured: `first line here<br>alpha …` at `width: 300pt` put `alpha` at 23.453pt while the lines above and below started at 20.0pt — **with no white space in the source at all**, so this survived the phase II fix and needed its own.

`PrecededByWordSeparator` is now never set for the word that opens a line. Suppressing the flow's own `PendingWordSeparator` was not enough: `word.HasSpaceBefore` is a fact about the *source text*, true of `alpha` in `…<br>\nalpha…` whichever line it lands on.

**3. An overlong word wrapped off the line the break opened, blanking it.** `HasOrdinaryWrapOpportunityBefore` took the same `IsSpaces` arm, reporting a soft wrap opportunity at the very *start* of a line, which [css-text-3 §5](https://www.w3.org/TR/css-text-3/#line-breaking) does not allow. `<p style="width:70pt">AA<br>Supercalifragilisticexpialidocious</p>` put the word two line-heights below `AA` — the `<br>`'s own line blank — where a space instead of the `<br>` put it one. Under `overflow-wrap: break-word` the emergency split was blocked and the line wasted entirely.

## Verification against Chromium

Driven through the repo's existing Playwright dependency, with the bundled Source Code Pro embedded via `@font-face` for **both** engines so neither reads host fonts (#956's trap). Deltas in CSS px on `getBoundingClientRect().left`:

| fixture | Chromium | PeachPDF | delta |
|---|---|---|---|
| issue sample, **multiline** — labels / values | 0 / 450 | 0 / 450 | **0.000** |
| issue sample, **minified** — labels / values | 0 / 450 | 0 / 450 | **0.000** |
| `<br>`↵`<span>` | 0 | 0 | 0.000 |
| `<br><span>` `b</span>` | 0 | 0 | 0.000 |
| leading white space at block start | 0 | 0 | 0.000 |
| `<span>AA</span> <span>BB</span>` (mid-line) | 28.812 | 28.800 | 0.012 |
| space after an atomic inline | 28.812 | 28.800 | 0.012 |
| space after a left float | 53.328 | 53.333 | 0.005 |
| justified line opened by `<br>` | 0 | 0 | 0.000 |
| `pre-wrap`, preserved spaces after `<br>` | 28.812 | 28.800 | 0.012 |

The multiline and minified rows being identical **in both engines** is the issue's acceptance criterion. The 0.012px residuals are Chromium's sub-pixel rounding of one space advance, present on the unchanged rows too — those rows are the over-reach guards: a space mid-line, after an atomic inline, after a float, or preserved under `pre-wrap` must still render, and does.

The issue's own sample was also rendered end to end through the CLI and rasterized with **both PDFium and MuPDF**; both agree, all three labels flush and all three values flush.

## Deliberately not done

- `childHasLeadingWhitespace` is not itself cleared, only its advance suppressed — it also tells `HasOrdinaryWrapOpportunityBefore` that a wrap opportunity exists before the child's first word, which the space offers whether or not it is rendered.
- A float is not counted as line content, and Chromium agrees (53.328 vs 53.333px).
- `TryApplyLineClamp`'s own `Words.Count == 0` guard is untouched: it reads as the same question, but guards a `Words[^1]` access, and `IsAtLineStart` returns false for a line holding only an atomic inline — whose `Words` list *is* empty — so the swap would index off the end.
- The vertical (`CreateVerticalLineBoxes`) path is untouched: it never adds this advance and does not place the break word on the following line, so none of the three defects exist there.

## Testing

- Full suite **11 671 passed / 0 failed** (net8.0); `dotnet build PeachPDF.slnx -t:Rebuild` **0 warnings**.
- **100% line and branch coverage on all 19 changed executable lines.**
- 9 new tests — 6 in `WhiteSpaceLayoutIntegrationTests`, 3 in `OverflowWrapIntegrationTests`. The behavioral ones were confirmed failing against unmodified `main` before being kept; the rest are over-reach guards that must pass either way.

## Note for reviewers

An unrelated, pre-existing defect turned up next door and is **not** addressed here — filed separately as #1093: an inline box whose content is narrower than its border box has its left padding and border added to the line twice. With `<span style="padding-left:20pt;border-left:1pt"></span><span>BB</span>` (no white space at all), Chromium puts `BB` at 27.656px and PeachPDF at 56.000px — exactly double the 28px box. Confirmed **identical on unmodified `main`** for every fixture, so this change neither causes nor worsens it; it needs its own look at `FinalizeFlowBoxExit`'s "hack for actual width handling" tail rather than a white-space fix.


jhaygood86 (approved):
Thorough, well-diagnosed fix. Verified the root cause independently: `Line.Words.Count == 0` is genuinely the wrong test for "is this line at its start," since a forced-break word is placed on the line it opens rather than the one it closes — `IsAtLineStart` (no non-marker word AND no `Line.Rectangles` entry) is the right question.

I confirmed css-text-3's phase II removal rule is quoted accurately ("a sequence of collapsible spaces at the beginning of a line is removed" is verbatim spec text), and independently reproduced the flagship Chromium comparison (bundled-font-pinned, per the repo's own #956 convention) — multiline and minified renders of the issue's sample agree exactly in both PeachPDF and Chromium, plus the justified-line-after-`<br>` and overlong-word-after-`<br>` cases. Mutation-tested all three defects independently by reverting each fix in isolation: each reproduces its own specific claimed symptom without needing the other two, confirming they're genuinely distinct bugs sharing one root cause rather than one fix incidentally masking the others. Also confirmed the three "deliberately not done" calls are sound — in particular, `TryApplyLineClamp`'s `Words.Count == 0` guard really would index off the end if swapped for `IsAtLineStart`, since that method immediately reads `Words[^1]`.

One non-blocking thing worth a quick follow-up: `HasOrdinaryWrapOpportunityBefore`'s XML remarks quote a sentence in quotation marks as if citing css-text-3 §5 directly ("a line must not begin ... with a soft wrap opportunity"), but that exact sentence doesn't appear anywhere in css-text-3 or css-text-4 — I checked the full spec text. The underlying behavior is still correct (independently confirmed against Chromium), it's just not a real quote and should either lose the quotation marks or cite something that actually exists.

Also non-blocking: the disclosed pre-existing empty-inline-with-padding/border bug (mentioned in the PR's "note for reviewers") reproduces on a clean `main` as claimed and is unrelated to this diff, but doesn't appear to have its own tracked issue yet — worth filing so it doesn't get lost, similar to a note from the previous PR's review.

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. `dotnet test --framework net8.0` — 11,671 passed / 0 failed / 9 skipped, matching the PR's claim exactly. Diff coverage confirmed 100% on both line and branch coverage. CI green on all three platforms.

Merging.

PR #1094, opened 2026-09-15
Charge an inline box's left padding and border to the line once

### LINKED ISSUE #1093: An inline box's left padding and border are added to the line twice
### Summary

An inline box whose own content is narrower than its border box has its left padding and border added to the line **twice**, so everything after it on that line is pushed right by one padding+border width. An empty inline carrying `padding-left`/`border-left` is the clearest case: it advances the line by exactly double its width.

Found while fixing #1087 (collapsible white space at the beginning of a line) and deliberately left out of scope there — it is a box-model/advance defect, not a white-space one, and it is present on `main` independently.

### Reproduction

```html
<span style="padding-left:20pt; border-left:1pt solid red"></span><span id="b">BB</span>
```

No white space anywhere in the markup, so nothing about white-space processing is involved.

### Expected

`padding-left: 20pt` + `border-left: 1pt` = 21pt = 28px, so `#b` begins at **x = 28px**. Chromium puts it at 27.656px (its own rounding of the 1pt border to a device pixel).

### Actual

PeachPDF puts `#b` at **x = 56.000px** — exactly 2 × 28.0.

### More cases

Measured at `font: 16px` with the bundled Source Code Pro embedded via `@font-face` for **both** engines (so neither is reading host fonts — every advance is 9.6px). `left` is `getBoundingClientRect().left` in Chromium and the first word's left edge in PeachPDF, converted to CSS px.

| markup (probe is the last `<span>`) | Chromium | PeachPDF | error |
|---|---|---|---|
| `<span PAD></span><span>BB</span>` | 27.656 | **56.000** | +28.0 |
| `<span PAD></span> <span>BB</span>` | 27.656 | **65.600** | +28.0, +9.6 |
| `<span PAD>X</span> <span>BB</span>` | 46.875 | **65.600** | +18.4 |
| `<span>Z</span> <span PAD></span> <span>BB</span>` | 46.875 | **84.800** | +28.0, +9.6 |
| `<span></span> <span>BB</span>` (no padding — control) | 0.000 | 0.000 | — |

`PAD` = `style="padding-left:20pt; border-left:1pt solid red"`.

The control row shows the error appears **only** when the inline carries left padding/border. Content *inside* the padded inline is positioned correctly — in row 3 PeachPDF puts `X` at 28.000px, exactly right. It is the advance *past* the box that is wrong.

### Likely cause

`CssLayoutEngine.FinalizeFlowBoxExit`'s "hack for actual width handling" ([CssLayoutEngine.cs#L2522-L2527](https://github.com/jhaygood86/PeachPDF/blob/main/src/PeachPDF/Html/Core/Dom/CssLayoutEngine.cs#L2522-L2527)):

```csharp
if (opensHere && box.IsInline && 0 <= coordinates.CurrentX - startX && coordinates.CurrentX - startX < box.ActualWidth)
{
    coordinates.CurrentX += box.ActualWidth - (coordinates.CurrentX - startX);
    coordinates.Line.Rectangles.Add(box, new RRect(startX, trueStartY, box.ActualWidth, box.ActualHeight));
}
```

The two sides of that comparison are not the same quantity. `box.ActualWidth` includes the box's left padding and border, but `coordinates.CurrentX - startX` does **not**: `startX` is captured at the top of `FlowBox`, and the parent's own per-child dispatch already advanced `CurrentX` past the child's `leftSpacing` before calling it. So the delta measures a *content-only* advance and is compared against a *border-box* width, and the difference — the padding and border already paid for — is added a second time.

That accounts for all four failing rows exactly:

- row 1: content advance 0 < 28 → adds 28 → 28 + 28 = **56.0**
- row 3: content advance 9.6 (`X`) < 28 → adds 18.4 → 28 + 9.6 + 18.4 = 56, then the space 9.6 → **65.6**
- row 4: `Z` 9.6 + space 9.6 = 19.2, box adds 28 → `startX` 47.2, content 0 → adds 28 → 75.2, then the space 9.6 → **84.8**

### Note on the white space in rows 2 and 4

Chromium renders no space between the padded empty inline and `BB` in either row, because [css-text-3 phase I](https://www.w3.org/TR/css-text-3/#white-space-phase-1) collapses a white-space sequence *across* intervening inline box boundaries — the spaces on both sides of the content-less inline are one sequence, collapsed to a single space. PeachPDF renders both. That is a second, smaller divergence (9.6px) and is independent of the doubling above; fixing the advance alone will not close it.

### Environment

`main` at bcee172f, net8.0, Chromium via the repo's own Playwright dependency.

ChrisVanDijk:
Fixes #1093. Builds on #1092 (merged).

`FinalizeFlowBoxExit`'s "hack for actual width handling" compared two quantities that are not the same kind of thing, so every inline carrying `padding-left` or `border-left` had them charged to the line **twice** — pushing everything after the box right by one padding+border.

```html
<span style="padding-left:20pt; border-left:1pt solid red"></span><span>BB</span>
```

`BB` began at 42pt (2 × 21pt) instead of 21pt. No white space anywhere in the markup, so nothing about white-space processing is involved.

## The load-bearing idea

**`startX` is the box's *content*-box left edge, not its border-box left edge.** `FlowBox`'s per-child dispatch adds a child's `leftSpacing` (margin + border + padding) to the cursor *before* recursing into it, and `FlowBox` captures `startX = coordinates.CurrentX` at its own top — so by then the cursor is already past all of it. The matching `rightSpacing` is added by that same dispatch *after* `FinalizeFlowBoxExit` returns.

So `CurrentX - startX` measures **content width used**. It was being compared against `box.ActualWidth`, which is `ActualBoxSizingWidth` = `Size.Width + padding + border` — the **border-box** width.

The right-hand side is `Size.Width`: the declared *content* width, which is what the branch's own comment ("whether a box's content came out narrower than its declared width") always described. For a plain inline `Size.Width` is 0, so the branch now correctly does nothing at all.

The `Line.Rectangles` entry it registers is the box's border box — it is what background and border paint from — so it starts one `border + padding` back from the content edge `startX` names. That was wrong in the same way and is corrected alongside. Margin is excluded from both, since `ActualWidth` excludes it too.

## The arithmetic is the proof

The diagnosis was confirmed *before* any code changed, by predicting every failing fixture from it (monospace, each advance 9.6px, the padded box 28px):

- empty padded inline: content advance 0 < 28 → adds 28 → 28 + 28 = **56.0**
- with `X` inside: advance 9.6 < 28 → adds 18.4 → 28 + 9.6 + 18.4 = 56, then the space 9.6 → **65.6**
- mid-line: `Z` 9.6 + space 9.6 = 19.2, box adds 28 → `startX` 47.2, advance 0 → adds 28 → 75.2, then the space 9.6 → **84.8**

All three matched the measured values exactly, including the non-obvious 18.4.

## Verification against Chromium

Through the repo's existing Playwright dependency, with the bundled Source Code Pro embedded via `@font-face` for **both** engines. `getBoundingClientRect().left` of the box following the padded one, in CSS px:

| fixture | Chromium | before | after |
|---|---|---|---|
| empty padded inline, no space | 27.656 | 56.000 | **28.000** |
| empty padded inline, space after | 27.656 | 65.600 | **28.000** |
| padded inline with `X`, space after | 46.875 | 65.600 | **47.200** |
| padded inline mid-line | 46.875 | 84.800 | **56.800** |
| control: empty, no padding | 0.000 | 0.000 | 0.000 |
| control: `padding: 0 10pt`, no border | 36.266 | — | **36.267** |
| control: declared `width: 60pt` | 80.000 | — | **80.000** |

**The residual ~0.33px is Chromium's own, not ours** — and the controls are what establish that: it appears on exactly the fixtures carrying `border-left: 1pt` (Chromium renders that 1.333px border as ~0.989px) and disappears on the two border-free controls, which agree to 0.001px and 0.000px.

Row 2 is a bonus from #1092 composing with this: with the spurious `Line.Rectangles` entry gone, `IsAtLineStart` now sees the line as empty after a content-less padded inline, so the collapsible space after it is removed — which is what Chromium does, giving the spaced and space-free forms the same 27.656px.

## Testing

- Full suite **11 677 passed / 0 failed** (net8.0); `dotnet build PeachPDF.slnx -t:Rebuild` **0 warnings**.
- **100% line and branch coverage** on all 6 changed executable lines (8/8 branches).
- 6 new tests in `InlineLeftSpacingAdvanceTests`. **4 confirmed failing against `main`**; the 2 that pass either way are the deliberate controls — no padding, and a declared width wider than its content, which is the case this branch exists for and must keep working.

## Still open, deliberately — filed separately

Row 4 keeps a 9.6px divergence and it is a **different rule**, not a remnant of this one. Chromium renders no space between a content-less inline and what follows it, because [css-text-3 phase I](https://www.w3.org/TR/css-text-3/#white-space-phase-1) collapses a white-space sequence *across* intervening inline box boundaries — the spaces either side of the empty inline are one sequence, collapsed to a single space placed before it. PeachPDF collapses per text node and renders both. That is a white-space-processing change rather than a box-model one and needs its own fixtures across `pre`/`pre-wrap` and nested inlines, so it is tracked separately as #1095 rather than folded in here.


jhaygood86 (approved):
Precise, well-diagnosed fix — and this is exactly the pre-existing bug flagged as worth filing in the previous PR's review, now properly done as #1093.

I independently re-derived the diagnosis rather than just reading it: `startX` really is the content-box left edge (the dispatch adds a child's `leftSpacing` before recursing), so `CurrentX - startX` measures content width used, and comparing that against `box.ActualWidth` (border-box width) was comparing two different kinds of quantity. Working through the "before" arithmetic by hand from the old code — for the three fixtures in the PR's table — I got 56.0, 65.6, and 84.8 exactly matching the PR's claimed pre-fix values, and working through the "after" arithmetic with `box.Size.Width` (content width) as the comparison target, I got 28.0, 47.2, and 56.8, exactly matching the claimed post-fix values. The corrected `Line.Rectangles` entry (reconstructing the border-box left edge via `startX - ActualBorderLeftWidth - ActualPaddingLeft`) is also correct.

Mutation-tested it myself: reverted just this branch's comparison/correction back to `box.ActualWidth`, and 4 of the 6 new tests failed with exactly the claimed doubled-charge symptom (e.g. 21pt expected vs. 42pt actual — precisely 2×), while the 2 deliberate controls (no padding; a declared width wider than content) passed either way as designed. Restored the fix and all 6 pass again.

CI note: macOS failed once on this PR's run with an unrelated crash (`IndexOutOfRangeException` in `GlyphOutlineDecoder`, from the just-merged skip-ink feature, nothing to do with this diff) — confirmed as a flake by rerunning the identical job, which passed clean, and `main`'s own CI has stayed green throughout. Filed a follow-up to look into it separately since it's a real crash in a production code path, just not something this PR needs to address.

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. `dotnet test --framework net8.0` — 11,677 passed / 0 failed / 9 skipped, matching the PR's claim exactly. CI green on all three platforms (after the macOS rerun).

Merging.

Issue #1097, opened 2026-09-15
A percentage width on an inline-content inline-block is ignored and the box shrink-to-fits (CSS 2.1 §10.3.9)


ChrisVanDijk:
### Summary

A percentage `width` on an `inline-block` whose own content is inlines-only is ignored: the box is
sized and reserves room as though `width: auto` were declared, i.e. shrink-to-fit.

[CSS 2.1 section 10.3.9](https://www.w3.org/TR/CSS22/visudet.html#inlineblock-width) sends a
non-replaced inline-block to shrink-to-fit only when `width` is `auto`. A percentage is not `auto`:
it resolves against the width of the containing block
([section 10.2](https://www.w3.org/TR/CSS22/visudet.html#the-width-property)) and is then used
normally, exactly as an absolute length is.

An absolute length is handled — `CssLayoutEngine.ResolveAtomicInlineDeclaredWidth` assigns it, and
the flow then reserves it on the line and paints the border box at it (#1091). That method
deliberately declines a percentage.

### Reproduction

```html
<!doctype html>
<div style="width: 400pt">
  <span style="display: inline-block; width: 50%; background: #e8f4fb">x</span>|after
</div>
```

### Expected

The inline-block's content box is 200pt wide. Its background spans that width, and `|after` starts
200pt past the container's content edge.

### Actual

The box is shrink-to-fit: its background is the width of the `x` glyph, and `|after` sits directly
after that glyph.

An `inline-block` holding **block-level** content takes a different code path
(`CssLayoutEngine.FlowAtomicBlockContentChild`, which calls `GetBoxWidth`) and does resolve a
percentage, so this affects only the inlines-only case.

### Why it was out of scope

The width is resolved at the point the surrounding line places the box, where the containing block a
percentage resolves against is not available — the flow holds the block being flowed and its cursor,
not the box's own containing block. Resolving it needs either that containing block plumbed to the
placement point or the whole used-width resolution moved to somewhere that already has it, which is
a larger change than #1091 (making an absolute declared width size the box at all) was.

`InlineBlockDeclaredWidthTests.APercentageWidthIsLeftAlone` and
`InlineBlockDeclaredWidthGeometryTests.APercentageWidthLeavesThePaintedBoxToTheContent` pin the
current behavior, so closing this means changing those two fixtures rather than discovering the
exclusion.

PR #1099, opened 2026-09-15
Size an inline-block's painted box from its declared width

### LINKED ISSUE #1091: Declared width does not size an empty or inline-content inline-block's painted box
### Summary

An `inline-block` with an explicit absolute `width` advances following inline content by that width, but its own fragment/painted border box remains content-width. If the box is empty, its painted width collapses to zero (or to only its border widths).

This affects `inline-block`s whose content is empty or consists only of inline content. An `inline-block` containing a block-level child takes a different layout path and receives the expected geometry.

### Reproduction

```html
<!doctype html>
<style>
  .box {
    display: inline-block;
    width: 80px;
    padding: 4px;
    border: 2px solid #c33;
    background: #fee;
  }
</style>

<p>Inline content: <span class="box">x</span><span>after</span></p>
<p>Empty: <span class="box"></span><span>after</span></p>
```

A minimal geometry-only form is:

```html
<span style="display:inline-block;width:60pt">x</span>
<span style="display:inline-block;width:60pt"></span>
<span style="display:inline-block;width:60pt;border:1pt solid"></span>
```

### Expected

The declared width sizes the inline-block's content box. In the first sample, both `.box` elements therefore paint the same 80px-wide content area, plus 8px padding and 4px borders.

Following inline content should be positioned after that same painted border box.

[CSS 2.1 section 10.3.9](https://www.w3.org/TR/CSS22/visudet.html#inlineblock-width) only uses shrink-to-fit sizing for a non-replaced inline-block when `width` is `auto`; an explicit width is used normally.

### Actual

On current `main`, the 60pt geometry-only cases measure as follows:

- Inline content (`x`): painted/fragment width is approximately **5.0488pt**, the width of the glyph; `ActualBoxSizingWidth` is `0`.
- Empty: no box rectangle is emitted; `ActualBoxSizingWidth` is `0`.
- Empty with a 1pt border: painted/fragment width is only **2pt**, the two borders; `ActualBoxSizingWidth` is `0`.

Despite this, following text advances by the declared 60pt width. This makes the line position look correct while backgrounds, borders, overflow clips, and other geometry-based behavior use a content-width or zero-width box.

### Relationship to #31

#951 fixed #31's reported alignment symptom by using an absolute declared width when advancing the line cursor. It deliberately did not assign normal box geometry on this plain inline-content path, so #31's exact sample now aligns correctly while this narrower painted-geometry defect remains.

This is separate from #1032 (block-level content in the intrinsic-width walk), #1053 (atomic-inline baseline alignment), and #771 (vertical line flow).

ChrisVanDijk:
Fixes #1091.

[CSS 2.1 §10.3.9](https://www.w3.org/TR/CSS22/visudet.html#inlineblock-width) hands a non-replaced
inline-block to shrink-to-fit only when its `width` is `auto`; an explicit width is used as declared.

An `inline-block` whose own content is inlines-only is flowed into the surrounding block's line boxes
rather than laid out as a box of its own, and nothing on that path assigned it a used width. #951
made a declared width move the line cursor but deliberately left the box's geometry alone, so the box
still painted its background, border and overflow clip at whatever its content happened to measure —
nothing at all, for an empty one, while the text after it had already been advanced past the declared
width.

## The change

`CssLayoutEngine.ResolveAtomicInlineDeclaredWidth` resolves the used width once, into `CssBox.Size`,
before the box's content is placed. The cursor advance and the painted rectangle both read it back
from there, so there is one number rather than two derivations of it, and `ActualBoxSizingWidth`
reports the box's real width instead of zero.

The rectangle is corrected **on the inline axis only**:

* a box that placed no word has its border box stated by the flow
  (`ReserveAtomicInlineUsedWidth`) — it has nothing for `BubbleRectangles` to bubble;
* every other box keeps the rectangle its own words produced, and `WidenAtomicInlineRectangles` —
  a new step between `BubbleRectangles` and `ApplyVerticalAlignment` — widens it in place.

The first shape tried was the obvious one: state the whole rectangle at the advance point and let
`UpdateRectangle`'s later union widen it. That union is min/max on every edge, so a rectangle built
from the flow's `trueStartY`/`ActualHeight` pulled the top up from the ink to the line's flow top
(`Y = 20.1845` → `Y = 20` on a 12pt line) — the flow runs before `ApplyVerticalAlignment` and does not
know where the box's content finally settles. Widening after `BubbleRectangles` also inherits
`text-align`'s per-line shift for free, since the rectangle's left edge has already been through it.

`box-sizing: border-box` is honored, and a border-box width narrower than the box's own border and
padding is floored at their sum ([css-ui-3 §6.2](https://www.w3.org/TR/css-ui-3/#box-sizing)). Content
wider than the declared width still overflows rather than being cut back to it.

A percentage `width` is still treated as `auto` — it resolves against a containing block the line
being flowed does not know. That deviation is now tracked as #1097 and documented for readers in
`docs/html-css-support.md`.

## Relationship to #1094

Rebased onto it. #1094 fixed the same branch's two box-model errors generically — comparing a content
advance against `Size.Width` rather than the border-box `ActualWidth`, and anchoring the rectangle at
the border-box left — so this does not repeat them. Two things still separate an atomic inline from
that branch, which is why `ReserveAtomicInlineUsedWidth` exists rather than the branch simply being
reused:

* `Size.Width` is a **border**-box width under `box-sizing: border-box`, so it is not what to reserve
  between a content-box left edge and a trailing spacing the caller adds afterwards
  (`AvailableWidth` is, under either value);
* that branch states the rectangle whenever the content came out narrower, which for a box whose
  words *did* land on the line is the block-axis merge described above.

One consequence worth calling out: since #1094 an empty `auto`-width padded inline-block advances the
line correctly but paints nothing at all, having no words to derive a rectangle from. It paints its
own border box again here.

## The regression the test suite did not catch

Removing the old parent-loop advance, which looks redundant once the exit path reserves the width,
broke the *other* dispatch path. That advance sat outside the `if`/`else if` chain, so it also ran
after `FlowAtomicBlockContentChild` — and it was compensating a short advance there
(`CurrentX = b.ClientRight` comes out one leading padding short of the declared content box). The
`cascade_layers` showcase's 150px cards went from advancing 145.5pt to 121.5pt, which narrowed the
document enough to stop triggering shrink-to-fit, rendering the *whole page* at a different scale.

The full suite stayed green throughout; only showcase content-stream parity caught it. The advance is
restored (now fed by `ResolveAtomicInlineDeclaredWidth`'s return value, since
`FlowAtomicBlockContentChild` overwrites `Size`), and the dependency is recorded as an invariant so
the next change does not rediscover it.

## Verification

* `InlineBlockDeclaredWidthGeometryTests`, 14 fixtures: the issue's three shapes plus padding/border,
  the fragment tree carrying the width, the border-box left edge, filled-vs-empty parity,
  `box-sizing: border-box` and its floor, overflow not cut back, the percentage exclusion,
  `text-align` carrying the flow-stated rectangle, the empty auto-width box, and a nested empty child.
  Twelve fail against this branch's base (96dd810f); the two that pass are the exclusion guards.
* Full suite green on net8.0 (11,691 passed), 100% diff coverage, `dotnet build -t:Rebuild` with 0
  warnings.
* **All 129 existing showcases are byte-identical** in their decompressed content streams — the real
  check here, given the regression above.
* A new `atomic_inline_width` showcase covers the shapes that changed (fixed-width label columns,
  empty checkbox glyphs, `box-sizing` against a 120pt rule), rasterized through both PDFium and MuPDF.


jhaygood86:
Heads up — this PR is now showing merge conflicts against `main`. Likely cause: #1094 ("Charge an inline box's left padding and border to the line once"), which merged shortly before this PR was opened, touches the same area of `CssLayoutEngine.cs` (`FinalizeFlowBoxExit`'s inline/atomic-inline width and rectangle handling). Worth a rebase before this goes further — I'll hold off reviewing the substance until it's mergeable, since the diff I'd be looking at right now may not reflect what actually lands.

jhaygood86 (approved):
Solid, carefully-composed fix. I independently verified CSS 2.1 §10.3.9 (shrink-to-fit only when `width:auto`) and css-ui-3 §6.2 (border-box floored at 0 content, i.e. at border+padding) against the actual fetched spec text — both accurate.

The interaction with #1094 (merged just before this) was the thing I most wanted confirmed, and it checks out: this PR routes `inline-block` through its own new branch (`ReserveAtomicInlineUsedWidth`) *before* the `else if` that #1094's fix touches, so plain inlines (where `Size.Width` is always 0) are the only boxes that ever reach #1094's branch now — no double-charge, no regression of that fix. This composition is explicitly documented in the new invariant file, and reproducing it myself confirmed the reasoning holds.

The self-reported regression story is the most interesting part of this PR, and it reproduces exactly as described: removing just the compensating advance (while leaving everything else) produces precisely the claimed 24pt shortfall on a `cascade_layers`-shaped fixture, and — critically — the full 11,691-test suite stays 100% green through that mutation. That's a genuine test-suite blind spot the PR's own showcase-parity discipline caught and the unit suite couldn't, which is exactly the kind of thing worth recording as an invariant, and it is.

Also confirmed independently: the `box-sizing:border-box` floor at border+padding, that overflow past a declared width still isn't clipped back, and the union-vs-widen Y-coordinate design decision (stating an empty box's rectangle immediately vs. merging it via `CssLineBox.UpdateRectangle`'s min/max union) — the union really would pull the Y-coordinate to the pre-alignment flow-top, since `ApplyVerticalAlignment` hasn't run yet at that point in the flow, so widening in a separate later pass is the right call, not just an arbitrary implementation choice.

One thing surfaced along the way, not a regression and out of scope for this PR (title and #1091 are explicitly width-only): a declared `height` on an empty/inlines-only atomic inline-block doesn't get applied to its painted rectangle the same way width now does — `Size.Height` stays 0 through the flow that matters, only getting set later by unrelated machinery, too late to affect painting. Worth its own follow-up issue given this PR just fixed the symmetric width case.

I did not independently reproduce the "all 129 showcases byte-identical" claim (didn't regenerate and diff all of them), so I'd treat that one as PR-reported rather than independently confirmed — though the diff coverage (100%, confirmed), the 12-of-14 mutation-test split (confirmed, close enough to the PR's own count that the discrepancy is just a Theory-row counting difference), and the rasterized checks all corroborate the fix broadly.

My own verification: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings. `dotnet test --framework net8.0` — 11,691 passed / 0 failed / 9 skipped. CI green on all three platforms.

Merging.

PR #1102, opened 2026-09-15
Resolve percentage widths on inline-content inline-blocks

### LINKED ISSUE #1097: A percentage width on an inline-content inline-block is ignored and the box shrink-to-fits (CSS 2.1 §10.3.9)

ChrisVanDijk:
Fixes #1097

Summary:
- resolve non-auto inline-block widths against the child containing block using the current page-aware width basis
- preserve the existing single used-width path for line advance, painted geometry, and box-sizing
- update regression tests, support documentation, showcase coverage, and internal gap/migration notes

Verification:
- focused InlineBlockDeclaredWidth tests pass on net8.0 and net10.0
- full net8.0 suite: 11,714 passed, 9 platform-specific skips
- all changed executable lines covered
- all 131 showcases generated successfully
- full solution rebuild: 0 warnings, 0 errors


jhaygood86 (approved):
Verified independently, not just from the PR's own claims:

**Spec accuracy** — fetched CSS 2.1 §10.3.9 and §10.2 directly. §10.3.9: shrink-to-fit only applies when `width` is `auto` ("If `width` is `auto`, the used value is the shrink-to-fit width..."), so a declared percentage must size the box. §10.2: a percentage resolves against the containing block's width. Both match what the fix does and what the docs edit says.

**No duplicate resolver introduced** — `ResolveAtomicInlineDeclaredWidth` now calls the same `PageAwareWidthBasis` helper `GetBoxWidth` already uses elsewhere, rather than growing a second percentage-resolution path (this repo's CLAUDE.md explicitly flags that as an anti-pattern).

**`blockTop`/`coordinates.CurrentY` sourcing** — traced both branches (this one and the existing `FlowAtomicBlockContentChild` path) and confirmed they capture the same Y at the same point in `FlowBox`'s per-child walk before this child's width is resolved. No staleness/off-by-one-pass issue under multi-page reflow.

**Composition with today's #1094/#1099 changes to the same `FlowBox`/`FinalizeFlowBoxExit` code** — safe. This diff only touches the guard clause and percentage-basis computation in `ResolveAtomicInlineDeclaredWidth` plus the one call site's new argument; everything downstream (`WidenAtomicInlineRectangles`, `ReserveAtomicInlineUsedWidth`, the border-box flooring arithmetic) consumes `child.Size` uniformly regardless of whether it came from an absolute or resolved-percentage length.

**Mutation test** — reintroduced the old `EndsWith('%')` exclusion; the two rewritten tests failed with the exact old shrink-to-fit symptom (~10-17pt instead of 200pt). Restored.

**box-sizing: border-box** — not covered by the PR's own two tests (both are content-box only), so I verified this combination with an ad hoc test of my own (50% width + border-box + padding/border inside a 400pt containing block → correctly floors to 200pt border-box width). Confirms the fix is correct there too, but it's a real gap in the PR's own test coverage — non-blocking, worth a quick follow-up.

**Build** — `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings, 0 errors.
**Tests** — `dotnet test --framework net8.0`: 11,714 passed, 0 failed, 9 skipped (platform-specific), matching the PR description exactly.
**CI** — green on all three platforms.

Non-blocking nits: a missing blank line before the `### Stacking Context` heading left over from the docs edit (docs/html-css-support.md), and the border-box+percentage test gap noted above. Neither blocks merge.

Nice, tightly-scoped follow-up to #1099 — good that this landed same-day while the containing-block plumbing context was still fresh.

PR #1103, opened 2026-09-15
Baseline-align atomic inlines, and lay out a wrapping inline-block as a box of its own

### LINKED ISSUE #1053: Replaced and atomic inline content is not baseline-aligned

ChrisVanDijk:
## Summary

CSS 2.1 [§10.8.1](https://www.w3.org/TR/CSS21/visudet.html#line-height) aligns a baseline-aligned atomic inline by its own box rather than by font metrics. This engine left such content at the line's top, and had no single place where that rule lived.

- **Align replaced content by its bottom margin edge** — raster images, inline SVG, MathML and form controls. An inline wrapper moves with its atomic child, so its background and border stay attached.
- **Align an `inline-block` by its last in-flow line's baseline** while `overflow` is `visible`, and by its bottom margin edge when it is empty or its overflow is not visible.
- **Lay out an `inline-block` whose content cannot fit on one line inside it as a box of its own.** Its words used to be handed to the surrounding block's line breaker, so they wrapped at *that* block's measure: the box's content escaped it and its border box was drawn as two disjoint line rectangles, the upper one landing across the text above. Such a box is now routed through `FlowAtomicBlockContentChild` — the path an `inline-block` holding block-level content already took — so it breaks its lines at its own width and takes one place on the line that holds it.
- **`AtomicInlineBaselineOf` is now the one place the rule lives**, read both to size the closed line around the box and to decide how far the box moves, so the two cannot disagree.

Four further defects were found while closing that, each by rendering rather than by the suite, which was green for all of them:

- An `overflow: hidden` `inline-block` painted **completely empty**, and it took two fixes. Its clip and the text inside it were given different baseline deltas, so they moved apart — that is real, and a descendant now takes the enclosing atomic box's shift instead of deriving one from its own font metrics. **The box still painted empty afterwards.** `FragmentEmitter.OverflowClipOf` built the clip from `CssBox.Bounds`, and an `inline-block` flowed into the surrounding formatting context never has `Location`/`Size` assigned at all — its bounds are its own padding and border at the page origin, so the clip was a well-formed 8×4pt rectangle in the top-left corner and every word inside the box fell outside it. The clip is now built from the box's per-line rectangles, the geometry the painter already draws its border from. Pre-existing; reproduces identically on `main`.
- An `inline-block` holding block-level content was aligned by its bottom margin edge, where §10.8.1 (and Chrome) use the baseline of its last line box.
- The box establishing a line's formatting context was taking a place on its own line, aligning every line inside it to the last one and spreading its content out of its own border box.

## Behaviour a document author can notice

- Content wider than a declared `width` now **wraps inside the box** instead of widening it, and what follows on the line starts at the declared width. Content that genuinely cannot be broken — a single long word — still overflows without moving what comes after it. Both match Chrome on the same markup; two tests asserted the opposite and were rewritten against Chrome rather than loosened.
- An `inline-block` with `overflow: hidden` now paints its text at all.
- Recorded in `.claude/migration-notes/2026-09-16-a-wrapping-inline-block-is-now-a-box-of-its-own.md`.

## Deliberately out of scope

An atomic box is still never moved onto a line of its own when it does not fit in what is left of the current one, and the `box-sizing` correction coupled to it is therefore applied only to boxes routed here by their own inline content wrapping. Correcting it alone makes the rendering worse — at their correct width the `cascade_layers` cards no longer fit four to a row, and with no wrap the fourth runs off the page. Both belong in one change: #1105, with the reasoning in `.claude/accepted-gaps/an-atomic-inline-does-not-wrap-onto-a-line-of-its-own.md`.

## Testing

- `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0` — **11,727 passed, 0 failed, 9 skipped**
- `dotnet build PeachPDF.slnx -t:Rebuild` — **0 warnings**
- `diff-cover` against `origin/main` — **100%** on the changed lines
- All **131 showcases** regenerated and compared against `main` by decompressed content stream: 18 differ, each one inspected — every difference is an atomic inline taking the baseline. `baseline_alignment` gains a section on atomic inlines (which is what found two of the defects above) and `atomic_inline_width` a section on content too wide for its box; both were rasterized through PDFium **and** MuPDF.
- Expectations taken from headless Chrome on identical markup for: a multi-line `inline-block` after a wrapped caption, an `inline-block` holding a `<div>`, a narrow box holding breakable text, the same box holding one unbreakable word, and the four `baseline_alignment` samples.
- Mutation-tested: stubbing `AtomicInlineBaselineOf`, the ancestor walk in `BaselineShiftOf`, the `FirstReplacedWordOf` fallback, and the clip's rectangle fallback each fails the tests that claim them.
- The clip defect is covered by `InlineBlockOverflowClipPaintTests`, which reads the page's own content stream and checks that the clip the box pushes **encloses the text drawn inside it**. A geometry-only test of the same fixture passed throughout the failure — nothing about the box's layout positions was wrong, only what paint was handed — and the broken clip has real area, so it satisfies every check for a clip being emitted at all. Recorded as an invariant.

Closes #1053.


jhaygood86 (changes requested):
Verified independently against fetched CSS 2.1 §10.8.1 text and by reproducing behavior directly, not just from the PR's own claims.

**Spec accuracy — correct.** The relevant §10.8.1 text:

> "The baseline of an `inline-block` is the baseline of its last line box in the normal flow, unless it has either no in-flow line boxes or if its `overflow` property has a computed value other than `visible`, in which case the baseline is the bottom margin edge."

The PR's three-way branch order (overflow-not-visible → bottom margin edge unconditionally, even with in-flow content; no line box → bottom margin edge; otherwise → own last line's baseline) matches this exactly, including the easy-to-get-wrong part: overflow-not-visible wins even when the box *does* have in-flow content, and the code checks that first. Good.

**Blocking: multi-line inline-block content still hits the exact bug this PR claims to fix, and it's undisclosed.**

This engine flows an inline-block's own content directly into the *parent's* line-box coordinate space — there's no separate line-box tree for it. `box.FirstHostingLineBox == box.LastHostingLineBox == lineBox` therefore only detects "this inline-block's content never wrapped past one outer line box." When an inline-block's own text wraps across 2+ outer line boxes, that condition is false on every line it touches, but `HasWordOnLine` is *true* on all of them (the words are there, just not satisfying the single-hosting-line condition) — so the new loop's `else { continue; }` fires everywhere, the box contributes nothing to `baselineExtent`, and `BaselineShiftOf` falls through to the pre-PR word-based path. Net effect: a multi-line inline-block is left exactly where the old, broken behavior put it — its top border/padding pulled up into the preceding block, the same symptom this PR fixes for the single-line case.

This isn't hypothetical — I reproduced it directly with a narrow-width inline-block whose own text wraps to multiple lines, placed after a wrapped caption (same shape as the PR's own passing test, just wider content): it failed with

```
multi-line inline-block's margin-box top must not overlap the caption above it
(marginTop=22.825439453125, caption text bottom=25.124267578125)
```

Neither of the PR's two new tests exercises this shape — both use single-line inline-block content ("filled right up", "x", empty, clipped "x"). The updated accepted-gap file is accurately worded for what it *does* say (narrowly scoped to "one hosting line"), but it never names multi-line content as a still-broken case, and the PR description's "preventing its top border and padding from being pulled into the preceding block" reads as covering inline-block generally, which overclaims. Per this repo's CLAUDE.md, a genuine spec deviation left out of scope needs to be tracked as an issue and referenced from the accepted-gap file — that's missing here.

To resolve: either extend the fix to handle an inline-block whose content spans multiple hosting line boxes (comparing each hosting line's own baseline against that line's position), or explicitly narrow the accepted-gap file to name this case, file a tracking issue for it, and add a test that pins current (broken) behavior so it reads as a decision rather than an oversight.

**Everything else checks out:**
- `HasWordOnLine` is a correct, order-independent ancestor-walk (same pattern as the existing `FirstNonReplacedWordOf`), handles arbitrary nesting. One narrow, untested edge case: an inline-block whose only line content is a bare replaced element (e.g. just ``, no text) satisfies neither branch and silently keeps old behavior — low severity, not blocking on its own.
- No interaction risk found with #1056 (`LineBoxExtent.Union` is order-independent, this is a direct generalization), #1094 (vertical border/padding is charged unconditionally every pass, unlike the horizontal case #1094 fixed, so no double-counting), or #1099 (declared-width only affects horizontal extents, this loop only reads vertical).
- Mutation test: reverting the new loop and `BaselineShiftOf`'s inline-block branch makes both new tests fail with exactly the claimed old symptoms. Restored afterward.
- Build: 0 warnings, 0 errors. Tests: 11,716 passed, 0 failed, 9 skipped — matches the PR description.

**Non-blocking:**
- No `.claude/migration-notes/` file, despite this being exactly the kind of visible rendered-output position change this repo's convention calls for one (compare the notes added for #1056 and #1099 earlier today).
- `docs/html-css-support.md`'s atomic-inline-layout/vertical-align coverage wasn't updated.
- The existing `atomic_inline_width` showcase (whose exact fixture the PR's own new test copies) wasn't updated to reflect or demonstrate the fix, despite CLAUDE.md's convention that showcases are how several prior rendering bugs were actually caught.

The single-line fix itself is well-targeted and spec-accurate — happy to re-review once the multi-line gap is either closed or explicitly disclosed and tracked.

jhaygood86 (changes requested):
Re-reviewed the new commit. The replaced-content part (images/SVG/MathML/form controls) is good work, but the specific gap I flagged on the previous revision is still unresolved, and the PR now makes it worse by removing all disclosure of it.

**The multi-line inline-block bug is still present, confirmed with a fresh reproduction against this exact commit.** The `ApplyVerticalAlignment` loop's inline-block branch is unchanged:

```csharp
else if (FirstNonReplacedWordOf(box, lineBox) is { } word
    && ReferenceEquals(box.FirstHostingLineBox, lineBox)
    && ReferenceEquals(box.LastHostingLineBox, lineBox))
{
    atomicBaseline = word.Top + (word.FirstLineStyle ?? word.OwnerBox).ActualFont.Ascent;
}
else if (!HasWordOnLine(box, lineBox))
{
    atomicBaseline = marginBottom;
}
else
{
    continue;   // still the fallthrough for a box spanning 2+ parent line boxes
}
```

An inline-block whose own content wraps across 2+ of the parent's line boxes hits neither branch on any of the lines it touches (`FirstHostingLineBox != LastHostingLineBox`, but it does have words, so `HasWordOnLine` is true) — it falls to `continue` and is left at the old, pre-fix top-aligned position.

This time I went further than the numeric check: I rendered the reproduction (a narrow inline-block whose text wraps across two lines, placed after a wrapped caption) to an actual PDF and rasterized it with **both PDFium and MuPDF**, per this repo's paint-verification convention — both agree, and the result is a real, visible defect: the inline-block's border/background is drawn across the bottom of the caption's last line of text, not just an off-by-a-few-points numeric mismatch.

**This is now a more serious problem than on the last revision**, because this commit:
- Deletes `.claude/accepted-gaps/replaced-inline-content-is-not-baseline-aligned.md` entirely, rather than narrowing it to name this case,
- Changes the description to "Closes #1053" — a full-closure claim,
- Adds no test, gap file, or tracking issue for the multi-line case.

A verified-reproducing, visually-confirmed overlap bug is shipping completely untracked while the PR asserts full resolution. Per CLAUDE.md's accepted-gaps convention, a deliberate gap needs its own file and a tracking issue — this isn't even disclosed as deliberate anymore.

**The new replaced-content alignment itself is solid** — verified independently:
- CSS 2.1 §10.8.1 text confirms the bottom-margin-edge rule for replaced/atomic content; rendered `text  text with descenders gjpqy` and rasterized with both PDFium and MuPDF — the image's bottom edge sits exactly on the text baseline, with descenders correctly extending below it into the reserved strut descent.
- Mutation-tested the `box.IsImage` branch and `FirstReplacedWordOf` fallback — `AReplacedElementsBottomMarginEdge_SharesTheTextBaseline` and `OtherAtomicReplacedElements_ShareTheTextBaseline` (all 3 variants) correctly fail when disabled.
- Half-leading interaction is sound — `LineBoxExtent.Union` is a per-side max, not a sum, so folding in the atomic extent can't double-count against the text-derived extent.
- Build: 0 warnings, 0 errors. Tests: 11,721 passed, 0 failed, 9 skipped — matches the PR description.

**Non-blocking test-quality nit**: two of the new tests are tautological per mutation testing — `AnInlineWrapperMovesWithItsBaselineAlignedImage` still passes with the shift logic fully disabled (its assertion is a containment check that holds trivially when nothing moves), and `ATallReplacedElement_ReservesTheDescentBelowItsBaseline` also still passes disabled (its `>=` doesn't pin the value tightly enough — the stronger existing `LineHeightLineBoxExtentTests.ReplacedInlineContent_SizesTheLineFromItsMarginBoxAndTheSharedStrutDescent` already covers this correctly with `Assert.Equal`). Worth tightening, not blocking on its own.

**To resolve the blocking issue**: either fix the multi-line inline-block case (the baseline needs to come from the box's content on `LastHostingLineBox` specifically, not require `FirstHostingLineBox == LastHostingLineBox`), or restore a narrowed accepted-gap file naming this case, file a tracking issue for it, and walk the PR description back from "Closes #1053" to a partial-closure framing. Happy to re-review quickly once either lands — the replaced-content half of this is ready to go.

jhaygood86 (changes requested):
Re-reviewed the new commit (c1f1083). Good news first: **the multi-line inline-block overlap bug from the last two rounds is genuinely fixed.** Reproduced the exact same scenario (a narrow inline-block whose own text wraps across multiple internal lines, placed after a wrapped caption), rendered it end-to-end through `PdfGenerator`, and rasterized with both PDFium and MuPDF — the box now paints as a single clean bordered rectangle below the caption, no disjoint rectangles, no overlap. Mutation-tested the routing condition in `LaysOutAsAnAtomicBox` and confirmed the PR's own new regression test depends on it, not a coincidence. The `#1105` accepted-gap file and its tracking issue are accurate and genuinely distinct from #1053 — good disclosure this time.

**Blocking: the `overflow: hidden` inline-block clip fix claimed in the commit message does not actually work.** The commit message says: *"An `overflow: hidden` inline-block painted completely empty... [now fixed, a] descendant takes the enclosing atomic box's shift instead of deriving its own."* I rendered the PR's own literal new test fixture (`InlineBlockWithNonVisibleOverflow_KeepsItsWordsInsideItsClip`'s HTML) through `PdfGenerator` and rasterized with both renderers: **the box still paints completely empty.** The word never reaches the content stream at all (no `Tj` for it), and the clip PeachPDF emits is degenerate (`0 842 m W* n` — a single moveto, zero area).

Root cause: for an inline-block whose content fits on one line (the ordinary, non-`FlowAtomicBlockContentChild` path), the baseline-alignment shift only writes the shifted position into `CssLineBox.Rectangles`/word positions — never into `CssBox.Location`/`Size` (that assignment only happens when `LastOwnLineBaselineOf(box)` is non-null, which is false for this shape). But `FragmentEmitter.OverflowClipOf`/`BoundsOf` compute the overflow clip from `CssBox.Bounds` (`Location + Size`), which for this shape is never assigned (`Location=(0,0)`, `Size=(0,0)`, confirmed by instrumenting layout output directly).

To be fair to the fix: this exact failure reproduces byte-for-byte identically on `main` before any of this branch's work, so it's a genuinely pre-existing gap, not something this PR introduced. But the PR explicitly claims to have fixed it, and it hasn't — the new regression test for this only checks `CssBox`/`CssRect` layout geometry, never `Location`/`Bounds` or actual rendered output, so it passes while real output stays empty. This is exactly the content-stream/geometry gap CLAUDE.md's testing conventions warn about, and it's why the dual-renderer check matters here — a purely numeric/geometric test missed it.

To resolve: either fix `FragmentEmitter.OverflowClipOf`/`BoundsOf` to read the position the alignment shift actually produced for this box shape (rather than the never-updated `Location`/`Bounds`), or drop the claim from the commit message/PR description and file it honestly as its own tracked gap, the same way #1105 was just handled well.

**Everything else re-verified clean:**
- `AtomicInlineBaselineOf` correctly consolidates CSS 2.1 §10.8.1's rules in one place, read consistently by both the line-sizing and shift-computation call sites.
- The "aligns every internal line to the last one" fix is present and consistent with correct multi-line rendering.
- Independently verified both Chrome-behavior-reversal claims against real Chrome (not just the commit message): an inline-block with breakable content wraps inside its declared width; one with a single unbreakable word overflows without widening. Both match the PR's rewritten tests.
- Round-2's replaced-content (image/SVG/MathML/form-control) baseline work is untouched and still correct — reran the full `BaselineAlignmentLayoutIntegrationTests` suite and re-rasterized a fresh image+text fixture.
- Build: 0 warnings, 0 errors. Tests: 11,725 passed, 0 failed, 9 skipped.

Real, valuable progress on the hard part of this PR — just need the overflow:hidden claim to actually be true (or honestly withdrawn) before this is mergeable.

jhaygood86 (approved):
Fourth review pass, now approving. This PR went through real iteration and got materially better each round — worth summarizing the full arc since three rounds of REQUEST_CHANGES preceded this:

- Round 1: found and reproduced a real bug in the inline-block-only version — content wrapping across 2+ parent line boxes produced disjoint/overlapping painted rectangles.
- Round 2: the added replaced-content alignment (images/SVG/MathML/form controls) was verified correct, but the round-1 bug was still present, and the PR had deleted its accepted-gap file while claiming full closure with no disclosure.
- Round 3: the wrap-through-`FlowAtomicBlockContentChild` routing genuinely fixed the round-1 bug (verified via dual-renderer rasterization + mutation testing). But the same commit's second claimed fix — an `overflow:hidden` inline-block painting empty — was false: verified by rendering the PR's own test fixture through `PdfGenerator` with both PDFium and MuPDF, the box still painted empty. Traced the root cause to `FragmentEmitter.OverflowClipOf` reading `CssBox.Bounds`, which is never assigned for this box shape.
- Round 4 (this one): the new commit fixes `FragmentEmitter.OverflowClipOf`/`ClipSourceBoundsOf` to build the clip from per-line rectangles for inline-level boxes instead of `Bounds`. Re-verified independently:
  - Rendered the exact `overflow:hidden` fixture through `PdfGenerator`, rasterized with both PDFium and MuPDF — text now renders visibly, correctly clipped to the padding edge.
  - Mutation-tested: reverting just this commit's `FragmentEmitter.cs` change makes both new `InlineBlockOverflowClipPaintTests` fail, and re-rasterizing in that state reproduces the exact empty-box bug.
  - Confirmed the fix is correctly scoped — `ClipSourceBoundsOf` only branches for `box.IsInline`, falling through to the original `BoundsOf` for ordinary block-level boxes, so no regression risk to normal `overflow:hidden` clipping.
  - Confirmed it composes correctly with round 3's `FlowAtomicBlockContentChild` routing (that path clears and re-registers a single clean rectangle equal to its own `Bounds`, so the two fixes don't diverge).
  - New `InlineBlockOverflowClipPaintTests` genuinely reads the content stream and checks the clip encloses the drawn text, not just token presence — exactly the kind of test CLAUDE.md's testing conventions call for.
  - New invariant file accurately documents the trap for future readers.
  - Build: 0 warnings, 0 errors. Tests: 11,727 passed, 0 failed, 9 skipped (independently reran myself, exact match).
  - Diff coverage: 100% on changed lines (independently reran diff-cover against origin/main).
  - Spot-checked several regenerated showcases (`baseline_alignment`, `atomic_inline_width`, `invoice`) rasterized through both renderers — all correct, no visible regressions.

Good process here: each claimed fix was checked against actual rendered output rather than trusted from the commit message, and each time a claim didn't hold up, the next commit named the real root cause and fixed it properly rather than papering over it. Merging.

PR #1113, opened 2026-09-16
Hang an automatic underline from the decorated line's own baseline

### LINKED ISSUE #1111: An automatic underline is positioned from the decorating box's font against the line's content geometry, so a line mixing font sizes is underlined in the wrong place
## Summary

`FragmentPainter.PaintDecoration` positions an automatically placed underline as:

```csharp
Keywords.Underline => rectangle.Top + ResolveAutomaticUnderlineCenterOffset(
    styleSource.ActualFont, pen.Width, g.PixelsPerPoint),
```

where the helper returns `font.TextBaselineOffset + gap + thickness / 2`.

`rectangle.Top` is **content** geometry — the union of the inline content on that line
(`DecorationContent.SpanOf`, or the box's own per-line rectangle). `TextBaselineOffset` is measured in
the **decorating box's** font, which is the right font per
[css-text-decor-3 §2.4](https://www.w3.org/TR/css-text-decor-3/#line-decoration) ("the position ... are
based on the decorating box's font").

Adding the second to the top of the first only lands on the baseline while every glyph on the line is
set in the decorating box's own font. As soon as the line carries a different size, the sum is not the
baseline, and the underline is displaced by the difference between the two ascents — in either
direction.

## Repro

```html
<!DOCTYPE html><html><body style="margin:0">
  <div style="text-decoration:underline; font-size:12px">
    Total due: <span style="font-size:32px">$1,080.00</span>
  </div>
</body></html>
```

The decorating block is 12px and the run on the line is 32px, so the underline is drawn roughly one
32px ascent *above* where it belongs — through the middle of the large glyphs rather than under them.
Swapping the two sizes displaces it the other way, leaving the line floating well below the text.

## Measured

Painting the fixtures through `RecordingGraphics` at `PixelsPerPoint = 1.0`, comparing the `DrawLine`
call against the baseline the `DrawString` call actually painted at
(`point.Y + font.TextBaselineOffset`):

| document (decorating box / run on the line) | painted baseline | underline drawn at | should be |
| --- | --- | --- | --- |
| 12pt / 12pt — control | 10.863 | 12.113 | 12.113 |
| 9pt block / 24pt run | 21.803 | **9.474** | 23.05 |
| 24pt block / 9pt run | 22.224 | **37.053** | 23.47 |

So the 9pt-block case puts the line 12.3pt above the baseline and the 24pt-block case 14.8pt below it,
against the 1.25pt the control gets right.

Chrome 141 (`--headless --print-to-pdf`, rasterized at 300dpi) places the underline 40 device pixels
below the glyph ink top in every one of those three; PeachPDF matches on the control and diverges by
+53 device pixels on the third.

## Note

`CssLineBox.BaselineY` looks like the obvious anchor but is not usable as-is: inside a table cell,
whose UA `vertical-align: middle` moves the content block after the line closed, it is short by the
centering offset (17.601 against a painted baseline of 20.014 in a `line-height: 2` cell). The word
rectangles in the fragment tree are what `DrawString` is handed, so they are the reliable source.

dchupp:
Fixes #1111

## The defect

`PaintDecoration` placed an automatic underline at:

```csharp
rectangle.Top + font.TextBaselineOffset + gap + thickness / 2
```

Those two terms come from different places. `rectangle.Top` is **content** geometry — the union of the
inline content on the line (`DecorationContent.SpanOf`, or the box's own per-line rectangle).
`TextBaselineOffset` is measured in the **decorating box's** font.

Their sum is the baseline only while every glyph on the line is set in that same font. A line carrying
any other size is underlined one ascent-difference away, in whichever direction the mismatch runs.

[css-text-decor-3 §2.5](https://www.w3.org/TR/css-text-decor-3/#text-underline-position-property)
leaves the exact position "UA-defined in this level", but requires "a single thickness and position on
each line for the decorations deriving from a single decorating box" — which a position derived from
each rectangle's own top is not, since it varies with whatever content that rectangle happened to
bound. Its note is the reasoning for consulting the line: "since line decorations can span elements
with varying font sizes and vertical alignments, the best position for a line decoration is not
necessarily the ideal position dictated by the decorating box".

```html
<div style="text-decoration:underline; font-size:12px">
  Total due: <span style="font-size:32px">$1,080.00</span>
</div>
```

Painting the fixtures through `RecordingGraphics` at `PixelsPerPoint = 1.0` and comparing the
`DrawLine` against the baseline the `DrawString` call actually painted at:

| decorating box / run on the line | painted baseline | drawn at | should be |
| --- | --- | --- | --- |
| 12pt / 12pt — control | 10.863 | 12.113 | 12.113 |
| 9pt block / 24pt run | 21.803 | **9.474** | 23.05 |
| 24pt block / 9pt run | 22.224 | **37.053** | 23.47 |

12.3pt above the baseline in one direction, 14.8pt below it in the other — on 12px text.

## The fix

Resolve the anchor from the line's own words. Layout places each word's rectangle exactly its own
(rounded) `RFont.Ascent` above the baseline, so `wordRect.Y + wordStyle.ActualFont.Ascent` is the same
number for **every** baseline-aligned word however large it is set — `AlphabeticBaselineOn` returns the
first it finds and the rest cannot disagree, so there is no averaging or tie-break.

`PaintDecoration` then applies the **decorating box's own** ascent rounding to reach the baseline the
glyphs are painted on, keeping that term in the font §2.4 names. For a single-font line the two steps
cancel back to the old expression exactly, so **every such document paints byte-for-byte as before** —
which is most of them.

### `CssLineBox.BaselineY` is the obvious anchor and is not usable

It was the first implementation. It agrees with the painted baseline everywhere except inside a table
cell, where the UA stylesheet's `vertical-align: middle` moves the content block *after* the line
closed and recorded its baseline: a `line-height: 2` cell reports **17.601** against a painted baseline
of **20.014**. Taking it would have silently moved every underline in every table cell, and no existing
test would have caught it. The word rectangles in the fragment tree are what `DrawString` is handed,
and are paint's contract besides (`docs/architecture.md` §6). Recorded in
`.claude/invariants/paint-a-layout-recorded-baseline-is-not-the-painted-one.md`.

### The `vertical-align` rule is the strongest citation here

§2.5: a UA "**must** adjust line positions to match the shifted metrics of decorating boxes shifted
with `vertical-align` values other than `baseline` … but **must not** adjust the line position or
thickness in response to **descendants** of a decorating box that are so styled."

Both halves fall out of where `IsShiftedOffTheBaseline` stops walking: a shift on the decorating box
itself has already moved these words, so the baseline follows it; a shift below it excludes that word
and leaves the line where the rest of the text is. The shift lives on the `<sup>`, not on the anonymous
box that owns the word, so it is an owner-chain walk rather than a single lookup — and stopping below
the decorating box is also what keeps a cell's own `vertical-align: middle` from being mistaken for
one.

## Tests

`UnderlineBaselineIntegrationTests`, 12 cases, all asserting against the `DrawString` call that painted
the text (its `Point` and its `Font`) rather than against the box tree the painter itself reads, so no
assertion is self-consistent with the implementation — and all relative, so none depends on which font
the host resolves.

Bracketed: the crisp "one CSS pixel below the painted baseline" equality is asserted where the line
carries a run in the decorating box's own font; where nothing on the line is, the gap is bounded
instead, because the decoration is positioned from one font and the glyphs painted from another and
the two disagree by each face's ascent rounding (under a point, against the 12–15pt this used to be).
There is also a two-document test — adding a differently-sized run to a line must not move the
underline — and a superscript contrast case, without which a painter that simply followed the last run
it met would still pass everything else.

**Mutation-checked:** reverting `Html/Core/Paint/` to `origin/main` fails 7 of them and leaves 5
passing — the single-font control, the fallback case, and the mixed-size cases where the decorating box
happens to be the larger font on the line and the old code was coincidentally near-correct.

## Verification

- Full suite, net8.0: **11,876 passed, 1 failed** —
  `AnonymousBoxDefaultingTests.AListItemCostsLittleMoreThanAPlainBlock`, which fails identically on
  clean `main` (2.62x there, 2.61x on this branch, against a 2.55x bound). Pre-existing, and not
  something a decoration-only change reaches.
- **Diff coverage 100%** on the added lines.
- No new build warnings (`main`'s 7 `CS1998`s are unchanged).
- Against Chrome 141 (`--headless --print-to-pdf`, rasterized at 300dpi, underline band measured
  against the glyph ink): the mixed-size fixture that diverged by **+53 device pixels** now diverges by
  **0**, and the single-font fixtures are unmoved.

Caveat on the numbers: this is a single-TFM local run on SDK 9.0.120 targeting net8.0, so the absolute
totals will not match CI's.

## Deliberately out of scope

- **`line-through` and `overline` still key off the rectangle.** Both are wrong for the same reason on a
  mixed-size line, and a `line-through` also sits a little below the x-height centre browsers use on a
  *single*-size line — on the order of 5% of the glyph-ink height, so a calibration rather than a
  mismatch. Either way it wants its own before/after rather than riding along here.
- **`text-decoration-thickness: auto` stays at the engine's fixed 1**, per its existing recorded choice
  in `docs/html-css-support.md`.

## Also in this change

`DecorationContent` now always collects words, since the baseline needs them too. The walk only ever
runs for a box that declares a decoration, so this costs a dictionary insert per word of decorated text
and nothing at all on an undecorated document. `collectWords` was `WantsInkFrom`'s only caller, so both
are gone.

Docs, migration note, recent-fixes entry and the invariant are in the same commit.


dchupp:
Adding a throughput/allocation measurement, since this change makes `DecorationContent` collect words on a path that previously could skip them.

Built `main` (`7c05c882`) and this branch each into their own Release console host, rendering a 400-row table (three cells per row, two of them underlined, ~10 pages) and a byte-identical copy with the decorations removed. Twelve renders per measurement after three warm-ups, `GC.GetTotalAllocatedBytes(precise: true)`:

| document | | ms/render | MB/render |
| --- | --- | --- | --- |
| decorated | `main` | 195.7 / 201.0 | 129.45 / 129.38 |
| decorated | this branch | 195.3 / 196.8 | **129.38 / 129.38** |
| undecorated | `main` | 192.0 / 184.0 | 73.34 / 73.35 |
| undecorated | this branch | 171.3 / 178.6 | **73.37 / 73.37** |

Two runs each, both shown, because the wall-clock numbers on this box move more between repeats than they do between the two builds. Allocation is the stable signal and it is unchanged to the hundredth of a megabyte in both directions — the extra word collection only happens inside a walk that already runs for a decorated box, and an undecorated document never starts it.

Same caveat as the PR body: single-TFM local run, SDK 9.0.120 targeting net8.0.

jhaygood86 (changes requested):
Solid engineering here — the core diagnosis and fix are real and well-verified. I independently reproduced the bug (rasterized the mixed-size fixture on `main`: the underline floats above the line, visibly cutting through the large digits — a dramatic, unambiguous confirmation) and confirmed the fix resolves it cleanly on both PDFium and MuPDF. I also confirmed the single-font byte-for-byte-identical claim directly (SHA-256'd the extracted content streams from both `main` and this branch — identical apart from the PDF creation-date comment), verified the `wordRect.Y + Ascent` baseline-invariance claim by reading the actual layout code, and reproduced the table-cell `CssLineBox.BaselineY` divergence myself (got 19.015 vs 22.484 on my own fixture — same phenomenon, same direction, as your reported numbers, just a different specific case). Mutation-testing the paint-layer revert confirms a real, meaningfully-partial split (7 fail / 5 pass, not all-or-nothing).

Two blocking issues before this is mergeable:

**1. Build is not warning-clean.** `dotnet build PeachPDF.slnx -t:Rebuild` reports `xUnit1026` (unused Theory parameter `because`) three times — once per TFM — from the new `Underline_TopEdgeSitsOneCssPixelBelowThePaintedBaseline` test. This contradicts the PR description's "No new build warnings" claim and violates this repo's zero-warnings merge gate (CLAUDE.md). Easy fix: reference `because` in the assertion message, or drop the unused parameter.

**2. The spec citation is inaccurate, and it's shipped in trusted files, not just the PR description.** I fetched the actual css-text-decor-3 text. §2.4 is only the `text-decoration` shorthand property definition — it says nothing about position or thickness. The quoted parenthetical `"the position … are based on the decorating box's font"` doesn't appear anywhere in the spec. The real, relevant rule is in **§2.5**: *"The exact position and thickness of line decorations is UA-defined in this level. However, for underlines and overlines the UA must use a single thickness and position on each line for the decorations deriving from a single decorating box."* — notably weaker (UA-defined) than what's claimed. More importantly, §2.5 goes on to say the opposite of how this is framed: *"since line decorations can span elements with varying font sizes... the best position for a line decoration is not necessarily the ideal position dictated by the decorating box... UA consideration of descendant content will therefore result in better typography."* — i.e. the spec's own reasoning actually favors accounting for the line's real content (what this fix does), not "measuring from the decorating box's font" (closer to what the old, buggy code did). The one part of your framing that *is* well-supported by the spec is the `vertical-align`/superscript exclusion — §2.5 explicitly says a UA must adjust for a shifted decorating box but must not follow a shifted *descendant*.

This inaccurate §2.4 citation is baked into three places, not just the PR body: the shipped code comment in the paint file, and `.claude/recent-fixes/2026-09-16-underline-hangs-from-the-lines-own-baseline.md`. Please correct all three to cite §2.5 and soften the claim to match the actual "UA-defined, but must consider descendant content" text — the fix itself doesn't need to change, just the spec framing around it.

Non-blocking, worth a quick pass while you're in there:
- The mutation-test split you report is off by one — "4 passing" should be "5 passing" (I got 7 fail / 5 pass reverting the paint layer). The extra pass is a mixed-size case where the decorating box happens to be the larger font on the line, where the old code was coincidentally near-correct.
- The PR body says "12 cases," the recent-fixes doc says "11" — actual count is 12.
- The `line-through` "sits below the x-height centre browsers use" framing is directionally right against real Chrome but a bit stronger than what I measured (~5 percentage points of the glyph-ink height, not a dramatic mismatch) — not a blocker since it's already disclosed as out of scope, just a calibration note for whenever that follow-up happens.

Nice, thorough test suite and genuinely useful invariant file otherwise — happy to re-review quickly once the two blocking items are fixed.

dchupp:
Both blocking items are fixed, and you were right on both. Pushed as `85771b9b`.

**1. `xUnit1026`.** Real, and my "no new build warnings" claim was wrong rather than unlucky: I had been grepping the build output for `warning CS`, which cannot match an analyzer code. Re-checked with `grep -oE "warning [A-Za-z]+[0-9]+" | sort | uniq -c`, which now reports only the 14 `CS1998` (7 × 2, `main`'s own). `because` is now used, in an `Assert.True` that the underline is below the baseline at all — a weaker claim than the equality beside it, but a real one, and it makes the failure message name the case.

**2. The spec citation.** You are right, and I had written it from memory rather than fetching it — the thing this repo's CLAUDE.md warns about specifically. I fetched css-text-decor-3 before changing anything and confirmed your reading:

- §2.4 is the `text-decoration` shorthand. The quoted parenthetical does not exist.
- §2.5 has the real rule, and it is *"UA-defined in this level"*, with a `must` only for *"a single thickness and position on each line for the decorations deriving from a single decorating box"*.
- The note does cut against my framing: *"since line decorations can span elements with varying font sizes and vertical alignments, the best position for a line decoration is not necessarily the ideal position dictated by the decorating box."*

Two things I found while verifying, which changed the framing further:

- **The single-position `must` is the sentence this fix actually satisfies.** A position derived from each rectangle's own top is not one position per line for the decorating box — it varies with whatever content that rectangle happened to bound. So there is a normative hook here, it just is not the one I claimed.
- **The `vertical-align` exclusion is the strongest citation in the change, not an incidental detail.** §2.5's *"must adjust line positions to match the shifted metrics of decorating boxes … but must not adjust the line position or thickness in response to descendants of a decorating box that are so styled"* is exactly the two halves of `IsShiftedOffTheBaseline`: a shift on the decorating box has already moved these words so the baseline follows it, and the walk stopping below the decorating box is the "must not" half. I have put that citation on `AlphabeticBaselineOn`, where it belongs.

Corrected in all three places — the paint-file comment, the recent-fixes entry (which now has a short "what the spec actually says, having fetched it" section), and the PR description.

**One thing to flag while I was in there:** the propagation rule this area is often cited for — *"propagated to an anonymous inline box that wraps all the in-flow inline-level children"* and *"Atomic inlines … are not decorated"* — is in §2's own introductory prose, before §2.1, not §2.4 either. `PaintPropagatedDecoration`, `DecorationContent` and the `text-decoration-line` row in `docs/html-css-support.md` all cite §2.4 for it today. I have deliberately **not** touched those here, since they pre-date this branch and fixing them is a docs pass of its own — say the word and I will open a separate PR for it rather than widen this one.

**Non-blocking, all three taken:**

- Mutation split is **7 fail / 5 pass**, not 7/4 — confirmed independently (`git checkout origin/main -- src/PeachPDF/Html/Core/Paint/`, which is where I went wrong the first time: my earlier run checked out `HEAD`, which already had the fix, and I had also counted before adding the 12th test). Corrected in both places.
- **12 cases**, not 11. The recent-fixes entry said 11 because it was written before the fallback case that closed the last uncovered line. Both now say 12.
- `line-through` framing softened to "a little below … on the order of 5% of the glyph-ink height", which matches your measurement rather than my stronger wording.

Diff coverage is still 100%; full suite still 11,876 passed with only the pre-existing `AListItemCostsLittleMoreThanAPlainBlock` allocation failure (2.62x on clean `main`, 2.61x here).

Thanks for the independent reproduction — the SHA-256 of the extracted content streams is a better check of the byte-for-byte claim than anything I ran, and I will steal it.

jhaygood86 (approved):
Both blocking issues from the last round are resolved, independently re-verified from scratch rather than re-read from the PR text:

- **Build warnings**: `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings, 0 errors across all three TFMs. Confirmed the `xUnit1026` cause is gone: `because` is now read inside the assertion's failure message.
- **Spec citation**: fetched css-text-decor-3 §2.5 directly and compared verbatim against all three quotes now in the PR — the "single thickness and position" rule, the descendant-consideration note, and the `vertical-align` adjustment rule all match the real spec text exactly. No more misattribution to §2.4.
- **Mutation test**: reverted `Html/Core/Paint/` to `origin/main`, ran `UnderlineBaselineIntegrationTests` myself — 7 fail / 5 pass, exactly matching the corrected claim.
- **Core fix**: re-rasterized the mixed-font-size fixture with both PDFium and MuPDF — underline sits correctly under both the 12px and 32px text in both renderers, no split or offset.
- Build: 0 warnings, 0 errors (reran myself). Tests: 11,877 passed, 0 failed, 9 skipped (reran myself, exact match). CI green on all three platforms.

Good iteration — thanks for tracking down the real §2.5 text and correcting the mutation-test count rather than just asserting the original numbers were close enough. Merging.

Issue #1114, opened 2026-09-16
text-decoration-style: wavy paints as solid


dchupp:
## Summary

`wavy` cascades and computes correctly, but paints as a straight line.
`Html/Core/Utils/TextDecorationStyleMapper.ToDashStyle` maps it to `RDashStyle.Solid`, because no dash
pattern can express a wave: a decoration is stroked with `RGraphics.DrawLine`, which takes only a pen
and two endpoints.

[css-text-decor-3 §2.2](https://www.w3.org/TR/css-text-decor-3/#text-decoration-style-property) defines
`wavy` as "Draw a wavy line", so this is a visible spec deviation rather than a discretionary choice —
unlike the *geometry* of `double`, which the same section does leave to the UA.

`docs/html-css-support.md` listed all five styles as supported without qualification; #1112 corrects
that row to say `wavy` paints solid, and adds
`.claude/accepted-gaps/text-decoration-style-wavy-paints-solid.md` pointing here.

## Repro

```html
<span style="text-decoration: underline wavy">wavy</span>
<span style="text-decoration: underline solid">solid</span>
```

Both render identically.

## What it would take

Build an `RGraphicsPath` of half-period Béziers with `AddBezierTo` and stroke it with
`DrawPath(RPen, RGraphicsPath)`. The one thing to settle first is units: box-geometry paths are built
in `PixelsPerPoint`-divided coordinates (see `FragmentPainter.Decorations.BuildRingPath`'s remarks and
issue #812), while the decoration painter works in layout space, so the pen width and the path need to
agree about which they are in.

`DecorationSegments.Subtract` already splits a line into segments for atomic inlines and
`text-decoration-skip-ink`, so a wave has to start at each segment's own origin rather than at the
line's, or the phase will jump across a skip.

## Chrome reference

Chrome 141, `--headless --print-to-pdf`, rasterized at 300dpi, a 16px font:

| style | rows covered |
| --- | --- |
| solid | 47–49 |
| wavy | 54–61 |

So the wave occupies roughly 2.5× the resolved thickness vertically, and sits below where the
equivalent solid underline goes rather than being centred on it.

## Also

SVG text decorations go through the same mapper and are a single stroke for every style, including
`double` — so whatever shape the fix takes, `Svg/SvgRenderer` is a second caller to consider.

PR #1115, opened 2026-09-16
Draw text-decoration-style: double as two strokes

dchupp:
## The defect

`TextDecorationStyleMapper` maps every `text-decoration-style` to an `RDashStyle`, and `double` has no
dash pattern that could express it, so it fell through to `RDashStyle.Solid` and painted **one** line —
output byte-identical to `solid`.

`docs/html-css-support.md` listed all five styles as supported without qualification, so this was a
documented capability that silently did nothing.

```html
<span style="text-decoration: underline double">1,080.00</span>
<span style="text-decoration: underline solid">1,080.00</span>
```

Both render identically. Found comparing report-shaped CSS against Chrome: the accounting convention of
a double rule under a grand total came out with nothing to distinguish it from the subtotal above.

## The fix

`double` is not a pen — it is two strokes of the pen the other styles already use. So the mapper is left
alone (it still supplies the pen each stroke is drawn with) and the second `DrawLine` is added where the
segment is stroked, `FragmentPainter.StrokeDecorationSegment`. That keeps it **downstream of
`DecorationSegments.Subtract`**, so a double line breaks around an atomic inline and skips glyph ink
exactly as a single one does — no second code path to keep in step with those two rules.

### Which way the pair grows — measured, not argued

The first revision argued this from each line's "anchoring edge" and **got `line-through` wrong**.
Chrome 141, rasterized at 300dpi, 16px font:

| | solid | double |
| --- | --- | --- |
| underline | rows 172–174 | 172–174 + 181–183 |
| overline | 122–124 | **112–115** + 122–124 |
| line-through | 153–155 | 153–155 + **159–161** |

Chrome's rule is uniform: keep the first stroke exactly where the single one sits, and add the second
**below** for an underline *and* a line-through, **above** for an overline. It does not straddle. That is
what this now does. The underline direction also happens to be the only one its own anchor allows — its
top edge is what `ResolveAutomaticUnderlineCenterOffset` keeps below the alphabetic baseline — but that
argument does not decide line-through, and inventing one for it produced a straddle no browser draws.

### Geometry: a deliberate deviation, not a free choice

Two strokes of the resolved thickness with a gap of the same thickness, so the pair spans 3×.

[css-text-decor-3 §2.2](https://www.w3.org/TR/css-text-decor-3/#text-decoration-style-property) does not
define the styles itself — their "values have the same meaning as for the `border-style` properties", and
[css-backgrounds-3](https://www.w3.org/TR/css-backgrounds-3/#border-style) defines `double` so that "the
sum of the two lines and the space between them equals the value of `border-width`".

This **deviates** from that cross-reference. `border-width` is a total; `text-decoration-thickness` is a
stroke (`from-font` reads the font's own `underlineThickness`), and dividing it three ways would render
the initial 1px double underline as two ⅓px hairlines — fainter than the `solid` underline it is meant to
be a heavier version of. Chrome agrees, per the table above. Tracked as **#1121** with
`.claude/accepted-gaps/double-decoration-thickness-is-per-stroke-not-a-total.md`, per CLAUDE.md's rule
for a deliberate deviation.

### A double overline can leave the page, and that is disclosed

A double overline grows upward, making it the one decoration that reaches above its own box. Flush
against the top of a page there is no room: on a zero-margin Letter page the clip runs y 0..792 and the
strokes are emitted at **794.0** and **792.0**, so the upper one is 2pt off-page. With
`body { margin: 20pt 0 0 0 }` they are at 774 and 772, both well inside.

Disclosed rather than worked around. Repositioning would put an overline somewhere it was not asked to
be, silently, on the documents where an author can see it least — and nothing is lost against a browser:
**Chrome 141 loses the overline entirely on the same markup**, both strokes *and* the single-stroke
`solid` case, because it also positions it above the content and there is no page there. PeachPDF still
draws the lower stroke, so it is the more forgiving of the two and `double` is no worse off than `solid`
already was. See `.claude/accepted-gaps/double-overline-flush-to-a-page-top-loses-its-upper-stroke.md`
and the `docs/**` row.

## Tests

`TextDecorationDoubleStyleTests`, 8 cases, all stated as a difference against the equivalent `solid`
document rather than against absolute numbers, so none depends on the host's font: two strokes where
`solid` draws one, same pen and same x-extent, the first stroke unmoved, the separation tracking the
resolved thickness (1pt and 4pt), and each growth direction. The contrast half is a theory over
`dotted`/`dashed`/`solid`/`wavy` asserting each is still **one** stroke with the right dash style —
without it, a change that simply doubled every decoration would pass everything else.

`TextDecorationDoublePdfClipTests`, 3 new cases, added because the review found a real defect that no
draw-call test could see: **`TestRecordingGraphics` never applies clipping**. These generate a real PDF
via `PdfGenerator` and read the stroke coordinates back out of the content stream against the page's own
clip — both strokes inside it with headroom, the upper one outside without, and a double *underline*
unaffected either way as the control that keeps it about the overline's direction rather than about
doubling. (For the next such test: the page clip is a `W* n` *path*, not a `re W n` rectangle, so its
bounds come from the preceding moveto/lineto points — my first attempt failed on exactly that.)

**Mutation-checked:** reverting `FragmentPainter.Decorations.cs` to `origin/main` fails the 4 `double`
cases and leaves the 4 contrast cases passing.

## Verification

- Full suite, net8.0, run twice: **11,872 passed, 1 failed** —
  `AnonymousBoxDefaultingTests.AListItemCostsLittleMoreThanAPlainBlock`, which fails identically on clean
  `main` (2.62x there against a 2.55x bound). Pre-existing.
- **Diff coverage 100%.**
- `dotnet build PeachPDF.TestHarness -t:Rebuild` adds no warning to `main`'s baseline.
- Showcase rendered and inspected as a raster, not just generated — all three growth directions and the
  grand-total case read correctly.

Caveat: single-TFM local run, SDK 9.0.120 targeting net8.0, so absolute totals will not match CI's.

### One flake worth flagging

`GraphicsAdapterInkCrossingsTests.EachDescenderInARun_ReportsInkAtItsOwnPenPosition` failed **once**,
only under `--collect:"XPlat Code Coverage"`, with "both descenders should be found, got 1". It passes in
isolation and in two subsequent full runs, and nothing here reaches `GetInkCrossings`. It looks like the
process-wide `FontFactory` cache hazard CLAUDE.md warns about, surfacing when coverage instrumentation
changes how parallel collections interleave. Raising it rather than leaving it in a log.

## Out of scope

- **`wavy` still paints solid** — #1114, with
  `.claude/accepted-gaps/text-decoration-style-wavy-paints-solid.md` and a corrected `docs/**` row that
  now says so rather than listing it as supported. It needs a stroked `RGraphicsPath`, which has a
  coordinate-space question (`PixelsPerPoint`-divided path against a layout-space pen width) that
  `double` does not.
- **SVG text decorations are unchanged** and remain a single stroke for every style, `double` included.
  They share the mapper but not the painter; noted in the mapper's comment and the docs row.

Docs, showcase, migration note, recent-fixes entry and the accepted-gap note are in the same commit.


dchupp:
The first CI run failed on **net11.0/Windows only** — `TextDecorationSkipInkTests.Underline_All_SkipsTheSameWayAuto_Does`, expected 2 segments and got 1 — with net8.0 and net10.0 green on the same runner. I don't believe it is this change, and here is the reasoning rather than an assertion:

- `StrokeDecorationSegment` runs **after** `DecorationSegments.Subtract` has decided how many segments there are. For every style but `double` it makes the identical `DrawLine` call the old code made, so the stroke count is exactly the segment count, before and after.
- That test renders `skip-ink: auto` and `skip-ink: all` as two documents and compares them to each other. `SkipsInk` treats the two values identically and this change doesn't touch it, so there is no path here that could make those two disagree *with each other in particular* — which is the shape the failure took (`auto` found two descenders, `all` found one).
- #1113 is a same-base branch touching the same painter file, and its net11.0/Windows job is green.

I hit the same symptom once locally, on a different test — `GraphicsAdapterInkCrossingsTests.EachDescenderInARun_ReportsInkAtItsOwnPenPosition`, "both descenders should be found, got 1" — only under `--collect:"XPlat Code Coverage"`, passing in isolation and in two subsequent full runs. Same shape: a descender's ink not found.

My guess is the process-wide `FontFactory` cache CLAUDE.md warns about — two documents rendered in the same process resolving the same family to different cached state depending on what else is running in parallel, with ink measurement being the one thing that reads real outlines. I have no way to re-run the job without admin rights, so I have force-pushed (the recent-fixes entry now records both sightings) to get a fresh run. If it goes green, that is the evidence; if it fails again in the same place, I am wrong and I will chase it properly.

Either way it looks worth a dedicated look on its own — a test that is 11/12 reliable is a test whose failures get explained away.

jhaygood86 (changes requested):
Solid fundamentals — segment-subtraction integration is correctly downstream (confirmed a skip-ink-broken double underline produces 4 DrawLine ops, 2 x-segments × 2 strokes, not a separate path), the mutation-test split is exact (4 double-specific tests fail, 4 contrast tests pass, matching the PR's own numbers precisely this time), build is genuinely warning-free (verified both the full-solution and TestHarness-only claims independently), and diff coverage is genuinely 100%. Underline and line-through growth directions are visually correct on both PDFium and MuPDF.

Two blocking issues:

**1. Same spec-citation problem as #1113.** I fetched css-text-decor-3 §2.2 directly. The actual text is: *"This property specifies the style of the line(s) drawn for text decoration... Values have the same meaning as for the border-style properties [CSS-BACKGROUNDS-3]. wavy indicates a wavy line."* The PR (description, both new `.claude/` files, the mapper's doc comment, the new XML doc in `FragmentPainter.Decorations.cs`, and `docs/html-css-support.md`) repeatedly quotes this as *"Draw a double line"* and says it "leaves the rest to the UA" — that phrase isn't in the spec. More importantly, the real text directly cross-references border-style's `double` definition (fetched from css-backgrounds-3: *"the sum of the lines and the space must equal border-width"*), which the PR's reasoning does discuss and reasonably decline to follow (thickness-as-stroke vs. thickness-as-total is a defensible call) — but frames as declining an optional analogy under free UA discretion, when the spec actually points there. Please correct the citation everywhere it appears to reflect that this is a deliberate deviation from what the spec references, not filling a silence. (The `wavy` gap file/issue #1114 has a milder version of the same misquote — "Draw a wavy line" vs. the real "wavy indicates a wavy line" — worth fixing at the same time.)

**2. Real bug found by rendering: `overline: double`'s second stroke gets clipped away in a realistic zero-headroom layout.** Rendered the PR's own test structure (`<body style='margin:0'>` immediately followed by decorated text — exactly what `TextDecorationDoubleStyleTests.BuildAndLayout` uses) through `PdfGenerator` and rasterized with both PDFium and MuPDF: the upward-grown second stroke is almost entirely clipped by the page's own content-clip rectangle, leaving only a ~2px sliver. Confirmed in the content stream — the outer stroke's Y coordinate falls above the page's clip top edge, the inner stroke has only 0.08pt of headroom. Adding top margin makes both strokes fully visible in both renderers, confirming it's a real headroom/clip interaction, not a rasterizer artifact. The PR's own tests don't catch this because they only assert `DrawLine` arguments through `TestRecordingGraphics`, which never applies real PDF clipping — any real document with an `overline: double` heading flush against a page's top margin will silently lose the second stroke. This needs either a fix (constrain/reposition the second stroke so it can't escape the box's own available space) or an explicit disclosed limitation with a regression test that actually renders and rasterizes, not just records draw calls.

Non-blocking: the milder wavy-quote issue noted above, and a coverage-instrumentation-triggered flake I saw once in `TextDecorationSkipInkTests` (passed cleanly on immediate rerun, consistent with this repo's known process-wide font-cache flakiness pattern, not obviously related to this change).

Nice geometry reasoning otherwise (the border-style-double rejection, the per-decoration-type growth-direction logic) — happy to re-review once the clip issue and citations are addressed.

dchupp:
Follow-up on the CI failure: the re-push flipped it, which I think settles it.

- **Run 1** — windows-latest / net11.0 failed `Underline_All_SkipsTheSameWayAuto_Does` (expected 2, actual 1); ubuntu and macos green.
- **Run 2** — same branch, only a markdown file changed between them. **Windows now passes**; **ubuntu-latest / net11.0** fails a *different* test, `Underline_Auto_GapsLandOnTheDescendersThemselves` (expected 2, actual 0).

The failure moved platform and moved test, which is a race rather than something this change does. Both times it is net11.0 only, with net8.0 and net10.0 green in the same job on the same runner.

I have written it up as #1119 with all three sightings (including the one I hit locally on net8.0 under coverage instrumentation) and what I would do to separate "`FontFactory` cache race" from "net11 runtime difference" — the short version being that net11.0 being 3 for 3 argues against pure xUnit parallelism, since net8.0/net10.0 should then show it at a similar rate and have not.

Happy to hold this PR until that is understood if you would rather not merge against an intermittent red. The change itself is unaffected either way: `StrokeDecorationSegment` runs after `DecorationSegments.Subtract` has decided the segments, and for every style but `double` it makes the identical `DrawLine` call the old code made.

dchupp:
Both blocking items addressed, plus a third thing your review turned up that I had got wrong. Pushed as `3ee397ed`.

## 1. The spec citation — and a direction bug it was hiding

You are right that §2.2 defines the styles by cross-reference to `border-style`, and that "Draw a double line" is not in the spec. Corrected in all six places (description, both `.claude/` files, the mapper comment, the new XML doc, `docs/**`), and the framing is now "this **deviates** from what §2.2 points at" rather than "fills a silence" — with `.claude/accepted-gaps/double-decoration-thickness-is-per-stroke-not-a-total.md` and **#1121** filed to track it, per CLAUDE.md's rule for a deliberate deviation.

Correcting it made me go and *measure* the growth directions instead of arguing them from "anchoring edges", and **my line-through was wrong**. Chrome 141 at 300dpi, 16px:

| | solid | double |
| --- | --- | --- |
| underline | rows 172–174 | 172–174 + 181–183 |
| overline | 122–124 | **112–115** + 122–124 |
| line-through | 153–155 | 153–155 + **159–161** |

Chrome's rule is uniform — keep the first stroke, add the second **below** for underline *and* line-through, **above** for overline. It does not straddle. My "a line-through has no anchoring edge so it straddles" was a story I told myself; no browser draws it. Fixed, and the test renamed to `Double_GrowsUpwardForAnOverline_AndDownwardForALineThrough`. The underline direction survives, but as a measurement that happens to agree with the anchor argument rather than as its consequence.

## 2. The overline clip — reproduced, and disclosed rather than fixed

Reproduced exactly: on a zero-margin Letter page the clip runs y 0..792 and the two strokes are emitted at **794.0** and **792.0**, so the upper one is 2pt off the page. With `body { margin: 20pt 0 0 0 }` they are at 774 and 772, both well inside.

I have taken the "explicit disclosed limitation with a test that actually renders" branch rather than repositioning, for two reasons — and the second is a measurement I would like you to check, because it changes the severity:

- Constraining the second stroke would put an overline somewhere it was not asked to be, silently, on exactly the documents where an author can see it least.
- **Chrome loses the overline entirely on the same markup — both strokes, and the single-stroke `solid` case too.** Rendered `overline solid` and `overline double` at `margin: 0` through both engines at 300dpi: Chrome's output is pixel-identical between the two and shows no overline at all, while PeachPDF draws the lower stroke in both. So on this input PeachPDF is the *more* forgiving of the two, and `double` is no worse off than `solid` already was.

`TextDecorationDoublePdfClipTests` (3 cases) generates a real PDF via `PdfGenerator` and reads the stroke coordinates back out of the content stream against the page's own clip: both strokes inside with headroom, the upper one outside without, and a double *underline* unaffected either way as the control. Disclosed in the `docs/**` row and in `.claude/accepted-gaps/double-overline-flush-to-a-page-top-loses-its-upper-stroke.md`.

Your underlying point is the transferable one and I have written it into the recent-fixes entry: **`TestRecordingGraphics` never applies clipping**, so no draw-call test in this area can see a stroke that leaves the page. Two things worth knowing for the next such test — the page clip is a `W* n` *path*, not a `re W n` rectangle, so its bounds come from the preceding moveto/lineto points; my first attempt at the test failed for exactly that reason.

## 3. Non-blocking, both taken

- The `wavy` misquote is fixed in the gap file and in #1114 — the real text is "`wavy` indicates a wavy line", which is the one style §2.2 defines in its own words rather than by cross-reference. Worth noting that makes `wavy` the *less* ambiguous deviation of the two.
- The skip-ink flake: I saw it twice more in CI on this PR, and the second sighting is what settles it — the re-push flipped which platform failed and which test. Written up as **#1119** with all three sightings. Every one was net11.0.

## Verification

Full suite 11,875 passed, 1 failed (`AListItemCostsLittleMoreThanAPlainBlock`, identical on clean `main`). Diff coverage 100%. `dotnet build PeachPDF.TestHarness -t:Rebuild` adds nothing to `main`'s baseline — and I am now grepping `warning [A-Za-z]+[0-9]+` rather than `warning CS`, which is how the `xUnit1026` on #1113 got past me. Showcase re-rendered and inspected: the line-through caption was wrong too, and is fixed.

jhaygood86 (approved):
Both blocking issues from the last round are resolved — and this round found genuine improvements beyond just fixing them:

- **Spec citation**: fetched both css-text-decor-3 §2.2 and css-backgrounds-3's border-style `double` definition directly — the PR's quotes are accurate renderings of the real text, and the framing now correctly presents the thickness-as-stroke-not-total choice as a disclosed deviation from what the spec cross-references, not free discretion.
- **Growth-direction correctness**: independently rendered real Chrome at high resolution with pixel-gridded overlays for all three decoration types — confirmed line-through does NOT straddle (it grows down, same as underline), matching the PR's corrected logic exactly. Good catch by the author finding this themselves via more careful Chrome measurement — the original straddle logic I hadn't flagged was actually wrong.
- **The clipping gap**: confirmed it's honestly disclosed everywhere (accepted-gap file, docs row, code comment, and the new `TextDecorationDoublePdfClipTests` actively pin the current clipped behavior as a regression net rather than claiming it's fixed). Independently verified the mitigating claim too — printed real Chrome output flush to a page top with zero margin: the overline vanishes completely in both the solid and double cases, worse than PeachPDF's behavior of still drawing the lower stroke. Confirmed `TextDecorationDoublePdfClipTests` does genuine content-stream clip-path parsing (extracts the `W* n` path's actual points, not a naive rectangle check) and mutation-tested as load-bearing (all 3 fail when the fix is reverted).
- Mutation test: reverted the paint file myself, got 7 fail / 4 pass on the combined test files — double-specific and clip tests fail, contrast cases pass, matching the PR's claim.
- Build: 0 warnings, 0 errors (reran myself). Tests: 11,876 passed, 0 failed, 9 skipped (reran myself, exact match). Diff coverage 100%. CI green on all three platforms.

One optional, non-blocking suggestion: the new `double-overline-flush-to-a-page-top-loses-its-upper-stroke.md` gap file doesn't cite a tracking issue, unlike the other two gap files this PR adds (#1114, #1121). It's arguably a page-geometry/clipping limitation rather than a spec deviation in the strict sense CLAUDE.md's issue-tracking rule targets, and the file itself makes a good case that PeachPDF's behavior is already more forgiving than Chrome's on the same markup — so I won't block on it, but filing one for consistency would be a nice follow-up.

Merging.

jhaygood86:
Approved, but this now has a merge conflict — PR #1113 (also touching `FragmentPainter.Decorations.cs`) just merged ahead of it. Needs a rebase onto current `main` before it can merge; I'll merge automatically once that's done and CI is green again.

dchupp:
Rebased onto current `main` (now that #1113 has landed) and force-pushed as `42006479` — the conflict is cleared and this is mergeable again.

What the rebase touched, so the force-push is not a silent rewrite:

- **`docs/html-css-support.md`** was the only real conflict: #1113 and this PR edited adjacent rows of the same table. Resolved by keeping both — `main`'s `text-decoration-thickness` row with #1113's baseline sentence, and this PR's rewritten `text-decoration-style` row.
- **`FragmentPainter.Decorations.cs`** auto-merged, but one thing did not survive semantically: #1113 renamed `ResolveAutomaticUnderlineCenterOffset` to `ResolveAutomaticUnderlineClearance`, and the `<see cref="..."/>` in `StrokeDecorationSegment`'s remarks still pointed at the old name. Git had no reason to flag it; the build did, as `CS1574`. Updated.

Re-verified on the rebase rather than assuming it carried:

- Full suite, net8.0: **11,943 passed**, 1 failed — `AnonymousBoxDefaultingTests.AListItemCostsLittleMoreThanAPlainBlock`, the same allocation bound that fails on clean `main` here. Worth noting you measured 0 failures on your own machine for #1113, so that one looks specific to this box rather than to any branch.
- Warnings: only `main`'s own 14 `CS1998` (7 × 2), no `CS1574`.

No changes to the substance since the last round — the two blocking items from your review (the `border-style` cross-reference framing plus #1121, and the overline clip disclosure with `TextDecorationDoublePdfClipTests`) are as described in the previous comment, along with the `line-through` direction correction that reviewing the citation turned up.

Issue #1124, opened 2026-09-16
A double overline flush against the top of a page loses its upper stroke to the page clip


dchupp:
## Summary

Tracking issue for `.claude/accepted-gaps/double-overline-flush-to-a-page-top-loses-its-upper-stroke.md`,
added in #1115, so it carries a `#NNN` reference like the two gap files beside it (#1114, #1121).

`text-decoration-style: double` grows an overline's pair **upward**, which is what browsers do. That
makes it the one decoration that reaches above the box it belongs to, and with no headroom above the
text the upper stroke lands outside the page clip and is not drawn.

Measured, Letter page, zero page margin, `body { margin: 0 }`, 16px text: the page clip runs y 0..792
in PDF user space, and the two strokes are emitted at **794.0** and **792.0** — the upper one 2pt off the
page. With `body { margin: 20pt 0 0 0 }` they are at 774 and 772, both well inside.

## Why it was left as a disclosed limitation rather than fixed

- Repositioning the pair so it fits would put an overline somewhere it was not asked to be, silently,
  on exactly the documents where an author can see it least.
- **Nothing is lost relative to a browser.** Chrome 141 given the same markup loses the overline
  *entirely* — both the double and the single-stroke `solid` case — because it also positions it above
  the content and there is no page there. PeachPDF still draws the lower stroke, so it is the more
  forgiving of the two, and `double` is no worse off than `solid` already was.

The painter emits both strokes either way; the page clip is what removes one.

## Note on classification

This is a page-geometry/clipping limitation rather than a spec deviation in the strict sense — nothing
in css-text-decor-3 says where a decoration goes when there is no page left to draw it on. Filed for
consistency with the neighbouring gap files rather than because a spec rule is being broken.

## Current behaviour is pinned, not merely described

`TextDecorationDoublePdfClipTests` generates a real PDF and reads the stroke coordinates back out of the
content stream against the page's own clip path: both strokes inside with headroom, the upper one
outside without, and a double *underline* unaffected either way as the control. So a change that alters
this — deliberately or not — will fail rather than pass quietly.

Worth knowing for anyone who does revisit it: `TestRecordingGraphics` never applies clipping, so no
draw-call test in this area can see a stroke that leaves the page, and the page clip is a `W* n` *path*
rather than a `re W n` rectangle, so its bounds come from the preceding moveto/lineto points.

Issue #1121, opened 2026-09-16
text-decoration-style: double does not divide text-decoration-thickness three ways, as its border-style cross-reference would require


dchupp:
## Summary

Filed to track a deliberate deviation introduced by #1115, per CLAUDE.md's rule that a spec deviation
left in place is tracked as an issue rather than only recorded as prose.

[css-text-decor-3 §2.2](https://www.w3.org/TR/css-text-decor-3/#text-decoration-style-property) does not
define the decoration styles itself. It says their "values have the same meaning as for the
`border-style` properties", and
[css-backgrounds-3](https://www.w3.org/TR/css-backgrounds-3/#border-style) defines `double` as: "Two
lines ... **the sum of the two lines and the space between them equals the value of `border-width`**".

Read literally through that cross-reference, a `double` decoration's two strokes and their gap should
sum to the resolved `text-decoration-thickness`. #1115 does not do that: it draws each stroke at the
resolved thickness with a gap of the same thickness, so the pair spans **three times** it.

## Why the deviation

`border-width` is a **total** — the whole border occupies exactly that much. `text-decoration-thickness`
is a **stroke** width: `from-font` reads the font's own `underlineThickness`, and `auto` is the engine's
single-line thickness, neither of which describes a three-band total. Dividing it three ways would make
the initial `text-decoration-thickness: auto` (1px) double underline two ⅓px hairlines with a ⅓px gap —
thinner and fainter than the `solid` underline it is meant to be a heavier, more emphatic version of.

No browser does it either. Chrome 141, rasterized at 300dpi with a 16px font, keeps its first stroke
exactly where the solid underline sits (rows 47–49 for both) and adds a second at rows 56–58 — a pair
spanning a little under four times a single line, i.e. it also treats the thickness as per-stroke and
grows the total.

So the deviation is deliberate and matches observable browser behaviour, but it is a deviation from what
§2.2 points at rather than a choice in a space the spec leaves open, and #1115's first draft framed it
as the latter.

## What closing this would look like

Either a spec/browser-behaviour reconciliation that says the cross-reference is not meant to carry the
`border-width` sum rule across (worth raising with the CSSWG if so, since the text is unqualified), or a
`text-decoration-thickness`-aware total mode. Recorded meanwhile in
`.claude/accepted-gaps/double-decoration-thickness-is-per-stroke-not-a-total.md`.

PR #1125, opened 2026-09-16
Cite a tracking issue from the overline-clip gap note

dchupp:
Taking up the optional follow-up from #1115's review.

`.claude/accepted-gaps/double-overline-flush-to-a-page-top-loses-its-upper-stroke.md` was the one of
the three gap files that PR added without a `#NNN` reference, while its two neighbours carry #1114 and
#1121. Filed #1124 and cited it.

I have kept your own reasoning about why it is the odd one out, in the note itself rather than only in
the review thread — it is a page-geometry limitation rather than a spec deviation in the strict sense
CLAUDE.md's issue-tracking rule targets, since nothing in css-text-decor-3 says where a decoration goes
when there is no page left to draw it on. A future reader deciding whether to relitigate it should get
that distinction along with the reference, otherwise the issue number implies a spec violation that
isn't there.

The issue itself carries the measurement (strokes emitted at 794.0 and 792.0 against a clip top of
792.0 on a zero-margin Letter page), the Chrome comparison that makes this the more forgiving of the
two behaviours, and a pointer to `TextDecorationDoublePdfClipTests` as the thing that pins it.

Docs-only: one file, no code, no tests, and nothing in `docs/**` or `README.md`.

Fixes nothing and changes no behaviour, so there is no migration note or recent-fixes entry — happy to
add one if you would rather every change carry them.


jhaygood86 (approved):
Exactly what it says — verified the citation is accurate (issue #1124 exists, open, matches the gap file's content) and the nuance about this being a page-geometry limitation rather than a strict spec deviation is preserved. Thanks for following up on the non-blocking suggestion. Build clean, CI green. Merging.

PR #1126, opened 2026-09-16
Render border styles like browsers

ChrisVanDijk:
## Summary

- render dotted and dashed borders with browser-style fitted spacing, round dots, and predictable phase
- draw `double`, `groove`, `ridge`, `inset`, and `outset` borders with correct bands, bevel shading, and corner joins
- support rounded mixed-style borders through shared width-aware corner transitions, including curved `groove`/`ridge` bands
- phase patterned rounded sides against the complete closed contour so width-dependent corner dots match Chromium
- expand the border-style showcase and paint-level regression coverage

## Validation

- full solution rebuild succeeds with zero warnings
- 80 focused border paint tests pass on `net8.0`
- changed production lines have 100% diff coverage
- border showcase output compared in Chrome, PDFium, and MuPDF


jhaygood86 (approved):
Large, entirely paint-affecting change, reviewed in two parallel passes covering (1) dotted/dashed rendering, the \`XPen\`/dash-pattern spec fix, the \`RAdapter.GetPen\` pen-cache leak, and mitred-corner compositing, and (2) \`double\`/\`groove\`/\`ridge\`/\`inset\`/\`outset\` bevel shading and all the rounded-border work. Both came back clean, with real independent verification rather than trusting the commit messages:

**Dotted/dashed + infrastructure:**
- Confirmed the PDF 32000-1 §8.4.3.6 dash-array constraint (nonnegative, not all zero) — the old code was stricter than spec and made a round dot inexpressible; the fix matches the spec exactly.
- Rasterized dotted/dashed borders with both PDFium and MuPDF — round dots, even spacing, flush at every corner.
- Verified the period-fitting "smaller count sometimes wins" logic by hand-computing both candidate gaps for two constructed edge cases, confirming the code picks whichever is closer to ideal in both directions (not just always rounding one way), and mutation-tested it.
- The \`GetPen\` cache-leak fix was checked most carefully as the highest-risk item: confirmed the bug is real (reverting it makes a solid stroke inherit a round cap from a same-colored prior dotted stroke), confirmed the reset covers every property on \`RPen\`'s interface field-for-field against \`XPen\`'s own constructor defaults, and checked all 14 \`GetPen\` call sites for reliance on stale state — including the one legitimate case (a marker's hollow circle not setting \`Width\`) that depends on the reset's default, which is correctly documented in a new invariant file.
- The mitred-corner seam fix was verified with actual before/after pixel evidence — mutating the ring-fill path back to per-edge quads reproduces a measurably distinct blended seam color along the corner diagonal in both renderers; the real fix eliminates it.

**double/groove/ridge/inset/outset + rounded borders:**
- Fetched CSS 2.1 §8.5.3 directly — it only requires the two \`double\` bands plus gap to sum to the border width, not exact thirds; exact thirds is Chrome's convention (correctly implemented, though the commit message's citation slightly overstates what the spec itself mandates — non-blocking).
- Rasterized \`double\`/\`groove\`/\`ridge\`/\`inset\`/\`outset\` on square borders with both renderers — clean bands, correct mitred corners, visible carved/raised bevel.
- The "black-on-black inset" claim was verified by direct pixel measurement (not eyeballing) and independently recomputed in a standalone reimplementation of the bevel-color formula — matches to within a rounding unit.
- Rounded \`double\` border pill-overlap bug confirmed fixed — no stub artifact at pill end caps in either renderer.
- Mixed-style rounded borders (four different styles on one box) sanity-checked — no crash, no color bleed, clean per-corner transitions.
- Rounded dotted-border whole-perimeter phase fitting confirmed on a circle — even spacing, no seam.
- Traced the accepted-gap file history through all 7 commits and found the rounded groove/ridge gap was actually closed by a later commit in this same PR (not still open as the earlier commit message implied) — confirmed by rendering it directly and by \`docs/html-css-support.md\` now correctly describing it as implemented.
- Mutation-tested against the full 80-case \`BorderStylePaintIntegrationTests\` file: 51/80 fail when reverted, all specific and on-target; 80/80 pass restored.

Both reviewers independently ran the full build and test suite with matching numbers, which I reran myself: 0 warnings, 0 errors, 12,006 passed / 0 failed / 9 skipped. CI green on all three platforms.

Substantial, well-verified work — nice that the remaining rounded groove/ridge gap got closed within the same PR rather than left as a stub. Merging.

PR #1129, opened 2026-09-16
Fix intermittent shared font cursor corruption

### LINKED ISSUE #1119: text-decoration-skip-ink tests are intermittently failing on net11.0 only: a descender's ink is sometimes not found
## Summary

Three separate sightings of the same symptom — a descender that skip-ink should break the underline
around is not found — across two unrelated PR branches. **Every one was on net11.0**, with net8.0 and
net10.0 green in the same job, on the same runner, with the same fonts.

| # | where | test | result |
| --- | --- | --- | --- |
| 1 | local, net8.0, only under `--collect:"XPlat Code Coverage"` | `GraphicsAdapterInkCrossingsTests.EachDescenderInARun_ReportsInkAtItsOwnPenPosition` | "both descenders should be found, got 1" |
| 2 | CI, **windows-latest / net11.0** | `TextDecorationSkipInkTests.Underline_All_SkipsTheSameWayAuto_Does` | expected 2, actual 1 |
| 3 | CI, **ubuntu-latest / net11.0** | `TextDecorationSkipInkTests.Underline_Auto_GapsLandOnTheDescendersThemselves` | expected 2, actual 0 |

Sightings 2 and 3 are two runs of the same PR (#1115) — the second run is the **re-push of the first**,
with only a markdown file changed between them. Windows failed and Ubuntu passed; then Windows passed
and Ubuntu failed, in a different test. The failure moves between platforms and between tests, which is
what makes it look like a race rather than a platform behaviour.

## Why it does not look like either PR

Both branches touch `FragmentPainter.Decorations.cs`, so that is the obvious suspect, but:

- #1115 only changes how an already-decided segment is *stroked*; it runs after
  `DecorationSegments.Subtract` has decided how many segments there are, and for every style but
  `double` it makes the identical `DrawLine` call. It cannot change a segment count.
- Sighting 2 is `auto` vs `all` disagreeing **with each other**. `SkipsInk` treats those two values
  identically, and neither PR touches it — so nothing in either change could move one and not the other.
- Sighting 1 was on net8.0 on a branch (#1113) whose CI is green on net11.0 across all three platforms.

## A guess, and what would confirm it

`GetInkCrossings` is the only thing in this area that decodes real outlines, and `FontFactory` caches
resolved font data in process-wide `static readonly` dictionaries — which CLAUDE.md already flags as an
order-dependent-flakiness hazard with xUnit's parallel collections. Every one of these tests renders
*two or more* documents and compares them, so a cache populated differently between the two renders
would produce exactly this: one render finds the descenders, the other finds fewer or none.

What would separate "cache race" from "net11 runtime difference":

- run `TextDecorationSkipInkTests` with `-parallel none` on net11.0 in a loop and see whether it ever
  fails;
- log the resolved `FaceKey` and the glyph count `GetInkCrossings` walked on each of the two renders
  inside the failing test, so a failure says which render came up short rather than only that they
  differed;
- if it only reproduces under parallelism, the fix is probably a per-test font isolation rather than
  anything in the decoration code.

That net11.0 is 3 for 3 argues against a pure race, though — if it were only xUnit parallelism, net8.0
and net10.0 should show it at a similar rate, and across these runs they have not.

## Impact

It is a false red on unrelated PRs, and the shape of it — "the feature under test silently did less
than it should" — is the shape that gets explained away. Filing it so the next person who sees a
skip-ink test go red on net11.0 has the three data points rather than one.

ChrisVanDijk:
## Summary

- synchronize all OpenType cursor readers on the shared `OpenTypeFontface.SyncRoot` monitor
- prevent GSUB/GPOS shaping reads from racing glyph-outline decoding used by `text-decoration-skip-ink`
- add a deterministic regression test proving shaping reads block on the outline cursor monitor

Fixes #1119

## Testing

- `dotnet test src/PeachPDF.Tests/PeachPDF.Tests.csproj -f net10.0 --filter "FullyQualifiedName~GlyphOutlineDecoderTests|FullyQualifiedName~GraphicsAdapterInkCrossingsTests|FullyQualifiedName~TextDecorationSkipInkTests" --no-restore` (40 passed)
- `dotnet test src/PeachPDF.Tests/PeachPDF.Tests.csproj -f net10.0 --no-restore` (12,007 passed, 9 skipped)


jhaygood86 (approved):
This resolves a genuine, well-diagnosed race condition — and it's the likely root cause behind the recurring font-decoding flakiness that's come up repeatedly this session (disclosed but unexplained in PR #1094's macOS-only flake and PR #1115's ink-crossing flakes), which is documented in issue #1119.

Verified the mechanism directly: `OpenTypeFontface` has one mutable shared read cursor, guarded by a dedicated `SyncRoot` lock. Before this fix, two disjoint lock domains protected the same cursor — `GlyphOutlineDecoder`/`ColrTable` correctly locked on `SyncRoot`, but GSUB/GPOS/GDEF/MATH's per-lookup readers locked on `lock(face)`/`lock(_face)` instead — the object's own default monitor, unrelated to the `SyncRoot` field. `lock(obj)` and `lock(obj.SomeField)` are different monitors, so shaping reads and outline decoding could run concurrently, both mutating the shared cursor position — exactly the silent, non-throwing corruption that shows up as skip-ink missing a descender it should have found. Traced this to commit `8152c8b2` three days prior, which introduced `SyncRoot` and wired it into the *cached* lookup-getter wrappers and the outline/COLR decoders, but left the pre-existing `lock(_face)` in the actual per-lookup `Read*Lookup`/`ReadLookupHeader` methods and in `GetActiveLookupIndices` (the widest, most frequently hit critical section, by its own comment) — exactly the leftover second lock domain, and consistent with the flake's timing.

Completeness: grepped every `lock(` site in `src/PeachPDF/Fonts` — all now correctly target `SyncRoot`, no bare `lock(face)`/`lock(_face)` remains. The few reader types that touch the cursor without their own lock (`CoverageTable`, `ClassDefTable`, `CpalTable`) are only ever constructed from call sites that already hold the lock, or during the one-time single-threaded load-time parse before concurrent access begins.

No deadlock risk — single lock domain, default `object` monitor (reentrant), no second differently-ordered lock ever acquired while held. Lock scope is unchanged in granularity from before (still per-lookup/per-decode, not per-render), so no new perf regression.

The new regression test is genuinely deterministic, not a probabilistic race-hunt: it holds the lock on the main thread, starts a reader thread, and proves via a `ManualResetEventSlim` + timed `Join` that the reader is provably blocked while the lock is held. Mutation-tested this myself — reverting just the lock target (`SyncRoot` → `face`) fails the new test 4/4 times, immediately and identically, exactly as expected for a construction-based deterministic test.

Race theory holds up against issue #1119's evidence — a scheduling-dependent race is consistent with the failure moving between platforms and between tests across identical-code reruns, and needing coverage-instrumentation overhead to manifest in one sighting (widened timing window), which a genuine platform difference wouldn't produce.

Build: 0 warnings, 0 errors (reran myself). Tests: 12,007 passed, 0 failed, 9 skipped (reran myself, exact match). CI green on all three platforms.

Great diagnostic work tracking this down to a genuinely subtle two-lock-domains bug rather than dismissing it as environmental flakiness. Merging.

PR #1135, opened 2026-09-16
Fix top and bottom replaced line sizing

ChrisVanDijk:
## Summary
- size top- and bottom-aligned replaced elements by their margin-box height exactly once
- keep edge-aligned atomic heights separate from baseline extents so line sizing is content-order independent
- use the same effective `vertical-align` resolution for sizing and placement, including `::first-line`
- add isolated margin/order/first-line regressions and an Acid2 fragment-geometry regression

## Validation
- `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0` — 12,059 passed, 9 platform skips
- `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings, 0 errors
- 100% diff coverage
- Acid2 rasterized with PDFium and MuPDF; the red/yellow eye-row strip is gone in both


jhaygood86 (approved):
Real spec-conformance fix with a genuine visible regression, thoroughly verified rather than trusted at face value.

**Spec accuracy**: fetched CSS2.1/CSS2 §10.8/§10.8.1 directly. Confirmed the two claims this fix rests on: (1) for a replaced/atomic element, "the box used for alignment is the margin box" — supports the margin-edge placement fix; (2) top/bottom-aligned boxes are a separate category from baseline-relative values and the spec's own stated goal is to "minimize the line box height" — confirms the diagnosis that treating a top/bottom-aligned image as baseline-aligned-then-re-aligned was a genuine double-count, not just a convenient bugfix narrative. Also verified `IncludeEdgeAlignedAtomicHeights`'s sequential growth computes `max(baseline, topHeight, bottomHeight)` regardless of accumulation order — the actual spec-mandated minimum.

**Acid2 fix — independently confirmed, dual-renderer, genuine before/after.** Swapped in `origin/main`'s pre-fix `CssLayoutEngine.cs`/`CssLineBox.cs`, rebuilt, rendered the real Acid2 fixture, rasterized with both PDFium and MuPDF — the red/yellow strip above the eyes is clearly there. Restored the PR's files, rebuilt, re-rendered: strip is completely gone in both renderers, eye row sits flush.

**`::first-line` double-count risk** (the exact thing the author's own iteration caught in an earlier WIP): traced `ResolvedFirstLineStyle`'s resolution timing (happens during cascade/parse, strictly before layout) and confirmed `EffectiveVerticalAlignOf` is literally the same function called from both the open-line sizing pass and the closed-line placement pass — structurally forecloses the sizing/placement disagreement that caused the earlier bug, not just something the new test happens to pass.

**Source-order independence**: verified with an independently-constructed variant (three different image/text height ratios, not just the PR's own pinned case) — all produce identical results regardless of ordering.

**Mixed top-and-bottom on one line**: not covered by the PR's own tests, so I had this checked directly — a 25pt top-aligned image + 15pt bottom-aligned image + ~18pt baseline text on one line correctly resolves to a 25pt line with both images flush to their respective edges. Works, but worth adding as a permanent regression test (non-blocking).

**Interaction with #1103**: full `BaselineAlignmentLayoutIntegrationTests` (22 tests, both #1103's original coverage and this PR's additions) passes clean. Spot-checked inline-block `vertical-align:top` with margin — confirmed unaffected (still uses border-box, a pre-existing gap this PR doesn't touch, now just more visible by contrast since images are correct). Worth a follow-up issue per this repo's convention, not this PR's job.

**Mutation test**: reverting the fix (back to unconditional baseline-extent union) fails all 8 new tests with exactly the predicted symptom — the Acid2 test off by +6.18pt, matching the PR's own cited 6.175pt almost exactly.

Build: 0 warnings, 0 errors (reran myself). Tests: 12,066 passed, 0 failed, 9 skipped (reran myself, exact match). CI green on all three platforms.

Thorough work, and good that the recent-fixes note is honest about the WIP iterations rather than presenting this as clean on the first pass. Merging.

Issue #1105, opened 2026-09-15
An atomic inline-level box is never moved onto a line of its own when it does not fit


ChrisVanDijk:
## Spec rule

CSS 2.1 [§9.4.2](https://www.w3.org/TR/CSS21/visuren.html#inline-formatting) makes an atomic
inline-level box one unbreakable unit on the line it sits on. When such a box does not fit in what is
left of the current line box, the line is closed and the box begins the next one — which is exactly
what a line break exists for.

## Current behaviour

`CssLayoutEngine.FlowAtomicBlockContentChild` places the box at `coordinates.CurrentX`
unconditionally. It never asks whether the box fits, so a box that does not fit overhangs the
containing block's right edge instead of moving down. A row of fixed-width `inline-block` cards
therefore stays on one line and the last card runs off the page, where a browser wraps it.

The wrap decision lives inside `FlowBox`'s per-word loop (the `wrapping` branch that restores
`MaxBottom`, applies `line-clamp`, records `PrecedesForcedBreak`, re-measures a `::first-line` tail,
re-intersects floats and opens the next `CssLineBox`). None of it is reachable for a box that
contributes no word to the parent's line, so honouring the break means lifting that sequence into a
helper both paths call.

## The box-sizing bug coupled to it

The same method sizes a declared `width` as the whole border box, rather than adding the box's own
padding and border to it under the default `box-sizing: content-box`. Measured on the
`cascade_layers` showcase's own `.card { width: 150px; padding: 16px }`: Chrome paints 182px,
PeachPDF paints 150px.

Correcting that in isolation makes the rendering worse, which is why the two belong in one change:
with the cards at their correct 182px, three fit across the page and Chrome wraps the fourth onto a
second row, while PeachPDF — having no wrap — runs it off the page. The short measurement is
therefore deliberately kept for boxes holding block-level content, and only for those; a box routed
to the same path because its own inline content wraps is sized correctly, since otherwise one
declaration would paint two different widths depending on whether its text happened to fit.

Recorded in `.claude/accepted-gaps/an-atomic-inline-does-not-wrap-onto-a-line-of-its-own.md`.

ChrisVanDijk:
## Status after #1144

PR #1144 fixed the independent-formatting-context inline-block path used by the cascade_layers cards:

- mixed block/inline content inside the inline-block is normalized and paints
- declared content-box width includes padding and borders
- the whole margin box is preflighted and moved to the next line when needed
[...]

PR #1144, opened 2026-09-16
Fix inline-block mixed content rendering

ChrisVanDijk:
## Summary

- normalize mixed block/inline content inside inline-block formatting contexts so trailing card text renders
- size and wrap fixed-width atomic inline-blocks as whole margin boxes
- preserve bottom margins between wrapped inline-block rows
- add layout and paint regression coverage and update support documentation

Related to #1105.

## Verification

- dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0 --no-restore (12,133 passed; 9 skipped)
- dotnet build PeachPDF.slnx -t:Rebuild -m:1 (0 warnings; 0 errors)


jhaygood86 (changes requested):
Both headline fixes are real, well-designed, and independently verified — but there's a genuine, dual-renderer-confirmed paint-leak regression in the line-clamp interaction that needs to be addressed first.

**The two main fixes check out cleanly:**
- Fetched CSS 2.1 §9.2.1.1, §9.4.2, and CSS Display 3 §2.3 directly — all three claims (anonymous block wrapper for mixed children, atomic inline-level box as one unbreakable unit, inline-block as an independent formatting context) are accurately represented.
- **DOM-normalization fix**: reproduced the exact repro on `origin/main` via the real CLI — the description text box genuinely comes out `w: 0` and never paints, confirmed with both PDFium and MuPDF. Confirmed fixed on this branch. Also traced a nested inline-block-in-inline-block case (not in the PR's own tests) — both levels normalize correctly, no infinite recursion or double-processing.
- **Flex-regression-avoidance claim**: constructed an `inline-flex` containing two block `<div>` items myself — layout traces are byte-identical between `main` and this branch, confirming the new traversal genuinely stays opaque to flex formatting contexts as claimed.
- **cascade_layers showcase**: rendered the real showcase on both branches — `main` shows only titles, narrow shrink-to-fit cards, and the 5th card missing entirely (run off-page); this branch shows all 5 descriptions, correct ~182px width, clean three-then-two wrapping. Both renderers agree pixel-for-pixel.
- **box-sizing:border-box interaction**: verified correct by construction (`ActualBoxSizeIncludedWidth` is 0 under border-box) and confirmed empirically with a dedicated fixture.
- **Margin-bottom fix + #1135 interaction**: verified independently with an asymmetric-margin fixture (not the PR's own symmetric 12pt case), and confirmed the two mechanisms are genuinely orthogonal by reading both — #1135's logic only ever applies to `CssRect` subclasses for replaced elements; an inline-block never produces an `IsImage` word and never touches that code path.
- **`OpenNextLine` extraction**: traced both call sites line-by-line against the deleted inline code — nothing dropped, the `followsForcedBreak: false` at the atomic-box call site is correct (an atomic wrap is never triggered by a `<br>`).

**Blocking: line-clamp + atomic-inline-block-wrap leaks the clamped box's own border into the painted page.** The PR's own test (`AtomicInlineWrapAtLineClamp_StopsBeforeLayingOutTheHiddenBox`) only asserts no *text* leaks (`DrawStringCalls`), and that part is genuinely fixed — mutation-tested and confirmed. But building a taller hidden card (nested `<div>`/`<p>`, not the PR's single `<b>`) and rasterizing with both PDFium and MuPDF shows a 1px sliver of the clamped-away card's border still painting at the top of the page. Root cause: `PrepareAtomicBlockContentChild` unconditionally sets `b.Location`/`ActualRight`/`ActualBottom` *before* the line-clamp check runs; when `TryApplyLineClamp` returns true and the caller does an early `return`, nothing rolls that state back, and `FragmentEmitter.ChildrenOf` walks `box.Boxes` (the DOM tree) rather than only confirmed `Line.Rectangles` entries, so it still emits a fragment for the abandoned box using its stale geometry. This directly contradicts the PR's own stated design intent ("line-clamp can stop here without first emitting content that must then be undone") — the text doesn't leak, but the box's own border/background does, which is exactly the kind of defect a `DrawStringCalls`-only assertion can't catch (this repo's own testing conventions call this out specifically for paint changes).

To resolve: either roll back `b.Location`/`ActualRight`/`ActualBottom` on the clamp-abort path, or add a guard in `FragmentEmitter.ChildrenOf` (or wherever fragments are emitted) confirming a box was actually flowed before emitting its fragment.

**Non-blocking**: diff coverage is 91% overall (clears the gate) but `CssLayoutEngine.cs`'s new wrap logic has three uncovered branches — right-float intersection, RTL text-indent, and left-float-after-wrap — none exercised by an atomic-inline-block wrap test. Worth a fast-follow given this dense code area's track record today.

Full suite: 12,133 passed, 0 failed (reran myself, exact match), and the adjacent #1094/#1099/#1103/#1135 test files all still pass (63/63). Solid work on the two main fixes — just needs the clamp-abort path cleaned up.

jhaygood86 (approved):
Verified the fix for the previously-flagged paint-leak, independently re-checked rather than trusted.

**Fix mechanism**: a two-part fix, and the split matters. `PrepareAtomicBlockContentChild`'s side effects (writing `b.Location`/`ActualRight`/`ActualBottom`) are now separated from the fit-check itself via a new side-effect-free `ResolveAtomicInlineBlockWidth` — geometry is only committed once the box is actually confirmed to flow. But that alone isn't sufficient, since the intrinsic-width measurement still recursively lays out descendants and an untouched box's default `(0,0)` location with non-zero border/padding is enough for `FragmentEmitter`'s default-bounds fallback to manufacture a fragment anyway — so a generation-scoped `SuppressFragmentEmissionForCurrentLayout()` flag (following this codebase's existing `_xGeneration`-style convention) closes that second gap, and `FragmentEmitter.BuildDraft` now returns null before recursing into a suppressed box's children, correctly excluding the whole rejected subtree.

**Leak confirmed gone**: reproduced the exact prior repro (taller nested `<div>`/`<p>` hidden card) end-to-end through `PdfGenerator`, rasterized with both PDFium and MuPDF — full-page inspection, no border/background sliver anywhere.

**No regression in the normal path**: regenerated all showcases including `cascade_layers`, rasterized with both renderers — five cards, three-plus-two rows, correctly sized, descriptions painting, identical between renderers.

**Test coverage gap closed**: the updated test now asserts on painted geometry directly (`FragmentEmissionSuppressedForCurrentLayout`, null hosting-line-boxes, and no matching `DrawRectCall`/`DrawPathCall`/etc. for the clamped box's colors) rather than only `DrawStringCalls` — exactly what was missing before.

**Mutation test**: reverting just this commit reproduces the leak identically in both renderers. Restored cleanly.

**One related, pre-existing gap found during exploration** (not a regression, not in this PR's scope): when a line-clamp is triggered by ordinary text wrapping *before* layout ever reaches the atomic inline-block's own fit-check, the suppression flag is never set, since control returns from a different `TryApplyLineClamp` call site earlier in the loop. Confirmed via mutation testing that this exact variant already reproduces identically on pre-PR `main` — a broader, separate instance of the same defect class this PR never claimed to fix. Worth its own tracked issue/accepted-gap, since fixing it properly means auditing every `TryApplyLineClamp` call site, not a quick addition here.

Build: 0 warnings, 0 errors (reran myself). Tests: 12,133 passed, 0 failed, 9 skipped (reran myself, exact match). Diff coverage 93%, well above the gate. CI green on all three platforms.

Good, principled fix — the suppression-flag approach is well-matched to how this codebase already tracks per-generation state elsewhere. Merging.

PR #1148, opened 2026-09-16
Fit a dotted/dashed outline to the ring's outer edge

ChrisVanDijk:
`OutlineDrawHandler.DrawDottedOrDashedLine` spanned each side's stroke from `rect.Left - mid` to `rect.Right + mid`, where `mid = outline-offset + outline-width / 2` — the point at which the band's *centreline* crosses the corner's mitre diagonal. `StyledStrokeFitting` measures whatever span it is handed, so every outline side was fitted to an edge **one whole outline-width short**.

Border does not have this problem: `BordersDrawHandler` spans `rect.Left..rect.Right`, its outer edge, corner squares included.

### Why the outer edge is the right length

Chrome has no separate outline painter for the non-`auto` styles. `PaintSingleRectOutline` hands its **border** painter a rectangle inflated by `outline-offset + outline-width`, with uniform per-side widths, and the ordinary border path draws it. An outline is therefore a border on the inflated rectangle, and its span has to be that rectangle's full side — exactly as `BordersDrawHandler` already does. `mid` stays, but only as the across-axis coordinate; only the along-axis span changed.

### What it looked like

The short span both picked the wrong pattern and misplaced it:

- **Wrong spacing.** A 120pt box with a 6pt dotted outline came out at period 12 over a 126pt span, where the equivalent border gives period 11.455 over 131.73pt.
- **Wrong phase.** The first and last mark sat half a width in from each corner, so the top and left edges' first dots landed half a width apart diagonally — every corner painted as a figure-8 blob — and `dashed` had a notch cut out of its corner L.

Verified by rasterizing the same page before and after through both PDFium and MuPDF, per the repo's two-renderer paint-verification convention. The two renderers agree, and the fixed corners are structurally identical to the equivalent border's.

### Tests

- **`OutlineStyleDottedOrDashed_MatchesTheEquivalentBorder`** (new) pins this against border rather than re-deriving Chrome's numbers a second time: an `outline-offset` of `-outline-width` lands the ring exactly where an equal-width border sits, so the two must emit the same four `DrawLine` calls, dash arrays included. Border's own dotted/dashed rendering was measured against Chrome in #1126, which makes it the oracle here. Verified it fails on the pre-fix code for both styles.
- **`OutlineStyleDottedOrDashed_LinesReachTheRingsOuterCorner_NoGap`** is what let this through #1126 — it asserted the old `mid` span as if it were the invariant, geometry read off the implementation rather than off the reference renderer. Renamed to `OutlineStyleDottedOrDashed_LinesSpanTheRingsOuterEdge` and now asserts the outer-edge span.

Full suite: 12135 passed, 0 failed (net8.0). `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings. Every changed line is covered.

### Docs

`docs/html-css-support.md` now states the outer-rectangle model explicitly. The #1126 migration note is still unreleased, so the outline paragraph in it was corrected in place rather than adding a second note contradicting the same baseline — relative to the last tag, outline dotted/dashed changed once, not twice.


jhaygood86 (approved):
Verified independently rather than trusted at face value — a good, honest follow-up to #1126.

**Geometric claim**: reconstructed the PR's own fixture and ran it through `StyledStrokeFitting.Fit` by hand — reproduced the buggy period-12-over-126pt and corrected period-11.455-over-131.73pt figures almost to the digit.

**Visual defects — confirmed before/after, both renderers, both styles**: physically reverted and restored `OutlineDrawHandler.cs` to render each state. Dotted before: two overlapping offset dots form exactly the described figure-8 blob at the corner, identically in PDFium and MuPDF. Dotted after: one clean dot. Dashed before: a visible notch where the corner L doesn't sit flush. Dashed after: clean flush L. All four confirmed in both renderers.

**The self-referential old-test claim**: confirmed precisely — the removed test literally asserted `rect.Left - mid` (the buggy implementation's own geometry) as its expected value, not a reference/oracle value. A real "test the implementation against itself" trap, worth remembering.

**New equivalent-border reference test**: mutation-tested by reverting just the production fix — both the new oracle test and the renamed span test fail with exact numbers matching the bug (period 12 vs 11.455; span short by exactly `mid`). Restoring the fix passes both again. Genuine oracle, not a rubber stamp.

**Scope containment**: confirmed the diff touches only `OutlineDrawHandler.cs` — border's own dotted/dashed rendering (already fixed in #1126) is untouched, its adjacent test suite (86 tests) passes unchanged, and solid/double outlines produce byte-identical content streams before and after (once volatile fields like creation timestamp are normalized).

**Migration-notes handling**: confirmed the note being edited in place was itself created by #1126, entirely after the `v0.9.18` tag (verified via `git merge-base --is-ancestor`), so it's genuinely unreleased and the PR's reasoning (correct the note rather than add a contradicting second one) is sound on the facts — worth noting CLAUDE.md's "exactly one new file" rule is literally scoped to *recording new* gaps/fixes to avoid cross-branch conflicts, so editing an unreleased note in place is a reasonable judgment call outside that rule's literal wording, not a violation of it.

Build: 0 warnings, 0 errors (reran myself). Tests: 12,135 passed, 0 failed, 9 skipped (reran myself, exact match). Diff coverage 100% on the changed file. CI green on all three platforms.

Good catch, and good discipline calling out that the original test was checking the implementation against itself rather than a reference. Merging.

Issue #856, opened 2026-08-28
Outline stroke width (pen.Width) ignores non-default PixelsPerInch


jhaygood86:
## Summary

\`OutlineDrawHandler.GetPen\` has the same defect as issue #851 (\`BordersDrawHandler.GetPen\`): it sets \`p.Width = width\` directly from a raw, un-divided layout-space (\`PixelsPerInch\`-inflated) value, with no division by \`g.PixelsPerPoint\`. At the library's default \`PixelsPerInch = 72\` (\`PixelsPerPoint = 1.0\`) this is invisible; at any other value, an \`outline\` renders visibly thicker than declared.

## Where

\`src/PeachPDF/Html/Core/Handlers/OutlineDrawHandler.cs\`, \`GetPen\` (around lines 247-259).

## Fix sketch

Same as #851: divide \`width\` by \`g.PixelsPerPoint\` before assigning it to \`RPen.Width\`.

## Context

Found while fixing #851 and deliberately left out of that PR's scope (a different draw handler, `outline` rather than `border`).

PR #1152, opened 2026-09-16
Ignore outline-width when outline-style is auto

ChrisVanDijk:
`outline: 8px auto` painted an 8px ring. CSS Basic User Interface 4 §4 says it must not:

> The `auto` value permits the user agent to render a custom outline style […] **The `outline-width` property is ignored when `outline-style` is `auto`.** User agents may treat `auto` as `solid`.

The two sentences do different jobs, and the old code conflated them. *"May treat `auto` as `solid`"* is a permission about the **style** — it is what lets this renderer keep painting four mitred solid quads instead of inventing a platform focus ring. *"The `outline-width` property is ignored"* is normative and unconditional, and holds whichever style the UA picks. Treating `auto` as `solid` therefore means solid **at the UA's own width**, not at the author's.

### The UA width, measured rather than recalled

CSS-UI-4 leaves the width undefined once it has said the author's does not apply, so it was read off Chrome's own rasterization (headless, both device scale factors):

| declared | ring |
| --- | --- |
| `8px auto`, `1px auto`, `20px auto`, `0 auto`, bare `outline-style: auto` | 2 CSS px — every one |
| `8px solid` | 8px |

Two things that are easy to get backwards, both pinned by tests: the width is ignored **including a declared zero** (Chrome paints `outline: 0 auto`), and the 2 is **CSS** px rather than device px — it renders 4 device px at a 2× device scale factor.

### Placement

Chrome also **centres** the `auto` ring on the rectangle `outline-offset` inflates the border box to, where every other style sits wholly outside that rectangle:

| case | box top edge | ring rows |
| --- | --- | --- |
| `8px auto`, offset 0 | 60 | 59–60 — one px either side of the edge |
| `8px auto`, offset 6 | 120 | 113–114 — centred on 114 |
| `8px auto`, offset −6 | 180 | 185–186 — centred on 186 |
| `8px solid`, offset 6 | 240 | 226–233 — wholly outside |

Centring needed no new geometry. `DrawRing` already takes the band's two reach distances, so pulling the *offset* back half a width before anything else runs turns the ordinary outward-facing band into a centred one, and every style arm, mitre and corner below inherits it untouched.

### Two ordering traps

- The `if (width <= 0) return;` guard had to move **inside** the non-`auto` arm, or `outline: 0 auto` would bail before the UA width was ever substituted.
- The UA width is a CSS length like any other, so it carries the `PointsPerPx * PixelsPerPoint` inflation that #856 applied to the cascaded width. A test pins the ring at both device scales.

### Verification

Rendered the same page through PeachPDF and Chrome and compared at 1 CSS px = 1 raster px: all four `auto` variants plus an `8px solid` and a `2px solid` control agree on both thickness and position. PDFium and MuPDF agree on the measurement, per the repo's two-renderer paint-verification convention.

Full suite 12172 passed, 0 failed (net8.0). `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings. Every changed line is covered.

### Scope

Chrome's focus ring has slightly rounded corners and this one stays square. That is the pre-existing, spec-permitted choice that an outline never follows `border-radius`, already documented — not a new gap; CSS-UI-4 says a UA *may* round an outline, never that it must.

The outline showcase's `auto` swatch was labelled "renders as solid", which is now misleading — it is a three-up showing `8px auto`, `1px auto` and `8px solid` so the ignored width is visible on the page.

### Migration

This **is** a released behaviour change, so it carries a migration note. `outline` support shipped in
v0.9.10, and `git show v0.9.18:src/PeachPDF/Html/Core/Handlers/OutlineDrawHandler.cs` shows the
released code reading `box.ActualOutlineWidth` for `auto`, with the released docs promising it
"renders identically to `solid`". A document that used `outline-style: auto` to get an outline of a
particular thickness will now show a thin 2px ring; `outline-style: solid` reproduces the old
rendering exactly and is the documented migration.

An invariant note goes with it. The first pass here skipped the migration note, reasoning from a
still-pending file in `.claude/migration-notes/` dated before outline support that no release could
have happened since. That inference is invalid — a pending note means only that nobody drained the
folder — and the habit it replaces is to date a behaviour change against the tag
(`git merge-base --is-ancestor <commit> <tag>`, `git show <tag>:<path>`). The same check run against
#1126 confirms the opposite for it: not an ancestor of v0.9.18, hence still unreleased, which is why
its note was amended in place rather than joined by a second one.


jhaygood86 (changes requested):
The technical work here is excellent, and I want to say that clearly before the one blocking issue: the spec reading holds up (fetched CSS-UI-4 §4 directly — the two sentences really are separate, unconditional rules, not a single conflated one), and the empirical measurements check out independently. I re-measured real Chrome myself (headless, both device scale factors, pixel-level analysis in continuous coordinates, not just a screenshot glance): all `auto` variants painted exactly 2 CSS px regardless of declared width, centered on the offset rectangle (`[-1,+1]` at offset 0, `[-7,-5]` at offset 6px), versus `solid` sitting wholly outside it — an exact match to this PR's own table once accounting for continuous-vs-discretized pixel boundaries. PeachPDF's rasterization (both PDFium and MuPDF, byte-identical) matches those independently-gathered Chrome numbers analytically, corners stay clean under the pulled-back offset, and `outline: 0 auto` genuinely paints a visible ring in both Chrome and the fix, not blank. Mutation-testing confirms the tests are load-bearing (8 of 10 fail with precisely the old symptom when the fix is reverted).

**Blocking: the "no migration note needed" claim is factually incorrect, and a note is required.** The PR states outline support "landed after the oldest still-pending note, so `auto` has never appeared in a release." I verified this directly: commit `80fe080a` ("Add support for outline / outline-* properties (#698)"), which includes `outline-style: auto` rendering as solid **at the declared width**, is an ancestor of tag `v0.9.18` — confirmed with `git merge-base --is-ancestor 80fe080a v0.9.18`. The released `docs/html-css-support.md` at that tag explicitly documented `outline-style: auto` as rendering "identically to `solid`" — the old, declared-width, non-centered behavior this PR changes. This is exactly the kind of visible, released behavior change CLAUDE.md's migration-notes convention exists for, and the convention was correctly followed for a related outline/border change earlier today (`.claude/migration-notes/2026-09-16-border-and-outline-styles-are-drawn-differently.md`) — this PR should get the same treatment: a note describing that a declared `outline-width` is no longer respected under `auto`, and the ring is now a fixed ~2px centered on the offset rectangle rather than sitting wholly outside it at the author's width.

Non-blocking: the code comments and recent-fix file attribute the CSS-px-to-position-space conversion formula to issue #856, but #856 is actually about a different bug (`pen.Width` bypassing position scaling, needing division) — the formula itself is correct and matches this repo's established length-inflation convention elsewhere, just mis-cited. Worth fixing in the same pass.

Add the migration note and fix the citation, and this is ready to go — the underlying fix is solid and thoroughly verified.

ChrisVanDijk:
Migration is in the PR (‎.claude/migration-notes/2026-09-16-outline-style-auto-ignores-outline-width.md‎), citation updated.

jhaygood86 (approved):
The missing migration note has been added — checked it directly: correctly identifies "up to and including v0.9.18" as when the old declared-width, non-centered behavior shipped (verified via git-tag ancestry in the last round), gives a clear before/after with the exact CSS-UI-4 §4 quote, and includes concrete migration guidance (use `outline-style: solid` to keep the old rendering). This is exactly what was needed.

Also noticed a new invariant file was added in the same commit documenting the "date behavior changes against release tags, not the notes folders" lesson from this review — including the exact outline-auto mistake as its own cautionary example — and then removed again in a follow-up commit. Not a concern either way: the substantive ask (the migration note) is satisfied regardless, and whether that lesson belongs as a formal code invariant or elsewhere is a reasonable judgment call.

The underlying technical fix is unchanged from the version already thoroughly verified last round (CSS-UI-4 spec reading confirmed accurate, Chrome's 2px/centered `auto`-outline behavior independently re-measured and matched, dual-renderer rasterization consistent, mutation-tested). Reran build and full suite myself: 0 warnings, 0 errors; 12,172 passed, 0 failed, 9 skipped — exact match. CI green on all three platforms.

Merging.

Issue #1157, opened 2026-09-17
A width:100% collapsed table with many long headers lays out ~19pt wider than Chrome's, overflowing its container


dchupp:
## Summary

A `width: 100%` table with `border-collapse: collapse` and a dozen columns whose headers are long
enough to drive the column minimums comes out about **19pt wider** than Chrome's, with the excess
accumulating left to right. On a page where Chrome has 20pt to spare, PeachPDF has 1.6pt — and once the
container is even slightly narrower than the fixture below, the last column runs past the page edge and
its header is clipped.

## Repro

```html
<!DOCTYPE html><html><head><meta charset="utf-8"><style>
@page { size: A4; margin: 10pt }
body { font-family: Arial, Helvetica, sans-serif; font-size: 12px; margin: 0 }
table { width: 100%; border-collapse: collapse; margin-bottom: 15px; }
.data-table th, .data-table td { border: 1px solid #000; padding: 6px; text-align: left; font-size: 10px; }
.data-table th { background-color: #f2f2f2; font-weight: bold; }
</style></head><body>
<table class="data-table"><thead><tr>
<th>C (Carbon)</th><th>Si (Silicon)</th><th>Mn (Manganese)</th><th>Cr (Chromium)</th>
<th>Ni (Nickel)</th><th>Mo (Molybdenum)</th><th>S (Sulphur)</th><th>P (Phosphorous)</th>
<th>Cu (Copper)</th><th>N (Nitrogen)</th><th>PREN</th><th>Ferrite Content</th>
</tr></thead><tbody><tr>
<td>0.01</td><td>0.01</td><td>0.01</td><td>0.01</td><td>0.01</td><td>0.01</td>
<td>0.01</td><td>0.01</td><td>0.01</td><td>0.01</td><td>0.01</td><td>0.01</td>
</tr></tbody></table>
</body></html>
```

| | page | rightmost text | spare |
| --- | --- | --- | --- |
| Chrome 141 | 595.0pt | 574.6pt | **20.3pt** |
| PeachPDF `0336f3db` | 595.3pt | 593.7pt | **1.6pt** |

## Where the width goes

Each column's first word, `pdftotext -bbox`, in points:

| column | Chrome | PeachPDF | cumulative drift |
| --- | --- | --- | --- |
| C | 15.0 | 14.9 | −0.1 |
| Si | 57.7 | 62.6 | +4.9 |
| Mn | 99.1 | 110.4 | +11.2 |
[...]

Issue #1154, opened 2026-09-17
A flex item sizes itself a hair too narrow for its own text, so a heading wraps where Chrome keeps it on one line


dchupp:
## Summary

A block inside a `display: flex` container is given an automatic width marginally smaller than the
width its own text needs, so the last word wraps. The same markup outside a flex container does not
wrap, and Chrome never wraps it.

The discrepancy is small and **non-monotonic in font size**, which is what makes it look like a
measurement/rounding mismatch between the intrinsic-width walk and real line layout rather than a
missing contribution: some sizes are fine, some are not, and they interleave.

## Repro

```html
<!DOCTYPE html><html><head><meta charset="utf-8"><style>
@page { size: A4; margin: 20pt }
body { font-family: Arial, Helvetica, sans-serif; margin: 0 }
.row { display: flex; }
h1 { margin: 0; font-size: 32px; }
</style></head><body>
<div class="row">
  <div><h1>JOB SHEET</h1></div>
</div>
</body></html>
```

`pdftotext -layout`:

```
Chrome 141      PeachPDF
----------      --------
JOB SHEET       JOB
                SHEET
```

The flex container has the full body width (~555pt) and the heading needs ~142pt, so there is no
shortage of room.

## What it depends on

| variation | Chrome | PeachPDF |
| --- | --- | --- |
| as above | one line | **wraps** |
| remove `display: flex` | one line | one line |
| remove the sibling `<div>` beside the `<h1>` | one line | **wraps** |
| add a second flex item beside it | one line | **wraps** |

So it is the flex item's own sizing: neither a sibling's contribution nor `justify-content` is involved,
and the identical markup as plain blocks is correct.

## The non-monotonic part

Sweeping only the `h1`'s `font-size` on that same fixture, PeachPDF:

| px | 28 | 30 | 31 | 32 | 33 | 34 | 36 | 48 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| | **wrap** | ok | ok | **wrap** | **wrap** | ok | ok | ok |

Chrome keeps it on one line at every one of those sizes. At 48px and 64px PeachPDF's laid-out text ends
within 0.4pt of Chrome's (229.9 vs 230.3, 297.9 vs 298.3), so the text measurement itself is right —
it is the width the item is *given* that is occasionally a hair short.

Measured on the original fixture at 32px: Chrome lays `JOB SHEET` from x=20.2 to x=162.3 (142.1pt).
[...]

PR #1158, opened 2026-09-17
Add three diagnostic scripts to tools/

dchupp:
Three scripts I kept rewriting while chasing rendering differences against Chrome for #1154 and #1157. Each exists because the obvious way to ask its question gives a misleading answer, and each of those wrong answers cost me real time on those two issues.

Nothing here ships. They need PyMuPDF, which `CONTRIBUTING.md` already expects for the two-renderer rasterization checks, and each is standalone with a `--help`.

## `pdf-text-state.py`

A PDF reader hands you an **effective** font size: whatever `Tf` set, after the CTM and the text matrix. So a page scaled by a `cm` and a genuinely mis-sized font are the *same number* in `span["size"]`.

That distinction is the whole question when a document comes out the wrong size, and I got it wrong first: I spent a while certain #1157's document was being scaled by a page transform, because every font size in it was uniformly 0.9486× its CSS value. It wasn't — the `Tf` sizes themselves were written scaled, by a host's shrink-to-fit second pass, which is a completely different thing to look for. One run of this would have said so:

```
  page 1  MediaBox=(0.0, 0.0, 595.44, 841.68)
    non-identity cm scales:
      depth 5: x=187.5 y=45   at (24.663, 778.657)     <- an image placement, not a page scale
    Tf sizes set in the content stream:
      /F0: [6.403, 6.75, 7.114, 7.826, 8.537, 9.96]    <- 10px asked for 7.5
```

It also reports a form XObject's own `/Matrix`, which is invisible in the page's own stream.

## `reduce-repro.py`

A real page that renders wrong is tens of kilobytes of markup plus a stylesheet nobody wrote for the purpose, and the cause is a handful of elements or one rule. Deleting by hand is slow and breaks the structure, which changes the layout and loses the repro.

This keeps subsets — of the children at a chosen nesting level, or of the top-level CSS rules — always emitting well-formed HTML, and asks a command of your choosing whether the result still reproduces:

```bash
python3 tools/reduce-repro.py page.html --css \
    --test 'peachpdf {} -o /tmp/o.pdf && ! pdftotext /tmp/o.pdf - | grep -q "ONE LINE"' \
    --out reduced.html
```

The reduction is delta-debugging rather than a plain halving, so it still finds a trigger that needs two elements together. It is what took #1157 from a 374 KB document to the twelve `<th>`s that actually matter.

## `compare-with-chrome.py`

Renders one file with both engines and pairs the words, so the two differences a text comparison cannot see — the characters are identical either way — become numbers:

```
  word                              chrome (pg,x,y)          peachpdf (pg,x,y)      dx      dy
  SHEET                             (1, 76.3, 20.3)            (1, 20.0, 47.8)   -56.3   +27.5
```

That is #1154's wrap. A wrap reads as a large negative `dx` with a positive `dy`; a sizing difference reads as drift that accumulates along the line or down the page.

It shells out to the CLI using `docs/cli.md`'s grammar (positional input, `-o OUTPUT`), overridable with `--peachpdf` so it can drive `dotnet run --project src/PeachPDF.Cli` directly.

## Verification

- All three run against real output here; the transcripts above are their actual output, not illustrations.
- `reduce-repro.py`'s bisect is exercised both ways on a fixture with a known single trigger — it reduces 5 elements to the 1 and 4 CSS rules to the 1, in both cases the right one.
- No code, tests or `docs/**` are touched, so there is nothing for the suite or the coverage gate to say. `tools/README.md` is new and indexes the existing `Update-HyphenationPatterns.ps1` alongside them.

Happy to drop any of the three, or move them somewhere other than `tools/` — that directory had no README and one PowerShell script, so I took it as the home for maintenance scripts rather than assuming.


jhaygood86 (approved):
Reviewed against this repo's CLAUDE.md requirements: this is a standalone, non-shipping addition to `tools/` (three diagnostic Python scripts + a README indexing them alongside the existing PowerShell script), so the C# build-warnings-zero, 90%-diff-coverage, and TestHarness-showcase requirements don't apply (no library code or user-facing feature is touched), and `docs/**`/README.md correctly aren't changed. Each script is standalone with `--help`, documents its own rationale, and the PR body includes verified real output. CI (macos/ubuntu/windows) is green.

---

PR #1159, opened 2026-09-17
Fix auto table columns overflowing a width-constrained table

### LINKED ISSUE #1157: A width:100% collapsed table with many long headers lays out ~19pt wider than Chrome's, overflowing its container

jhaygood86:
## Summary

- `DetermineMissingColumnWidths`'s auto-column distribution was a left-to-right greedy pass that granted a column its full max-content (un-wrapped) width outright once a running average happened to exceed it, without ever checking the grant against the table's actual available width — some columns kept their entire content on one line while siblings were squeezed toward bare minimum, and the sum could exceed the table's declared width, overflowing its container/page (the reporter's real-world case: a clipped header column that triggered a downstream shrink-to-fit rescale of an entire multi-page document).
- Replaced it with an order-independent three-way split against each auto column's own min/max content width: give every column its minimum on a genuine deficit, its maximum on a genuine surplus, and otherwise interpolate linearly between the two with a single global scale factor so every column lands within its own content bounds and the columns sum exactly to the available width (CSS 2.1 §17.5.2.2).
- Guarded the new interpolation math against vertical (`writing-mode: vertical-rl`/`-lr`) tables, whose columns report an infinite max-content width (no content measurement exists for that axis yet) and would otherwise divide-and-multiply through `Infinity` into `NaN` — falls back to an even split there instead, matching prior behavior. This edge case was caught by a post-change review pass before landing, not by the original fix.

## Test plan

- [x] New unit tests in `CssLayoutEngineTableTests.cs` covering the interpolation branch (fails pre-fix with the table overflowing its declared width, passes post-fix) and the deficit branch (multiple simultaneously-auto columns each landing at their own distinct content minimum)
- [x] New regression test in `TableWritingModeIntegrationTests.cs` for the vertical-table infinity guard (confirmed to throw without the guard, passes with it)
- [x] Full `PeachPDF.Tests` suite on net8.0: 12,213 passed, 0 failed (one pre-existing, environment-dependent memory-benchmark failure unrelated to this change, confirmed to fail identically on unmodified `main`)
- [x] `diff-cover` against `origin/main`: 100% diff coverage on the 32 changed/added lines in `CssLayoutEngineTable.cs`
- [x] `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings
- [x] Two rounds of code review against the diff (C# conventions, CSS 2.1 spec compliance, edge cases)

Fixes #1157

https://claude.ai/code/session_01GQx88gAsAkdLJnMY43xKzB

---
_Generated by [Claude Code](https://claude.ai/code/session_01GQx88gAsAkdLJnMY43xKzB)_


dchupp:
Confirming this from the reporting side, since the case #1157 was reduced from is a document I can run but cannot share.

**The reduced fixture from #1157**, exactly as it appears in the issue — A4, 10pt margins, Chrome 141 `--headless --print-to-pdf` as the reference, both engines pinned to the same font set. Column start positions, in points:

| column | Chrome | `main` | this PR |
| --- | --- | --- | --- |
| C | 15.0 | 14.9 (−0.1) | 14.9 (−0.1) |
| Si | 57.7 | 63.4 (+5.7) | 57.6 (−0.1) |
| Mn | 99.1 | 111.9 (+12.8) | 99.0 (−0.1) |
| Cr | 157.7 | 167.5 (+9.8) | 157.6 (−0.1) |
| Ni | 212.3 | 219.7 (+7.4) | 212.2 (−0.1) |
| Mo | 251.3 | 266.1 (+14.8) | 251.2 (−0.1) |
| S | 315.3 | 327.1 (+11.8) | 315.2 (−0.1) |
| P | 360.0 | 377.3 (+17.3) | 359.9 (−0.1) |
| Cu | 424.6 | 440.4 (+15.8) | 424.6 (0.0) |
| N | 468.4 | 490.5 (+22.1) | 468.4 (0.0) |
| PREN | 515.7 | 540.7 (+25.0) | 515.7 (0.0) |
| Ferrite | 546.3 | 571.3 (+25.0) | 546.3 (0.0) |

Rightmost ink: Chrome 574.60pt, `main` 597.10pt — 1.8pt past a 595.28pt page, with the last header clipped — and 574.56pt with this PR. Every one of the twelve headers now wraps where Chrome wraps it, and the whole table lands within 0.1pt of Chrome.

**The 26-document corpus the defect surfaced in.** Real documents rendered through the production path and compared against a frozen Chrome baseline:

| engine | pass | review | fail |
| --- | --- | --- | --- |
| `a46a17e` (our previous pin) | 26 | 0 | 0 |
| `0336f3db` / `main` | 24 | 1 | 1 |
| `main` + this PR | 25 | 1 | 0 |

The document that failed is the one #1157 was reduced from, and with this PR it comes out content-identical to Chrome: same 6 pages, same 9,436 characters, text scale 1.000x, nothing clipped — so the downstream shrink-to-fit rescale that made the whole document 5% small no longer fires at all. The remaining `review` is an unrelated text overprint that the document also has in Chrome; it is present identically with and without this change, so nothing here moved it either way.

Our own service suite is 108/108 green against this branch.

Caveat on those numbers: measured against a vendored mirror of `0336f3db` built on net9.0 (this box has SDK 9, so a local target tweak), with only `CssLayoutEngineTable.cs` taken from this branch — so it exercises the engine change, not the new tests, and the totals will not line up with CI's.

One corroboration that may be worth something: I reached the same `min + t × (max − min)` interpolation, with the same `+Infinity` guard for vertical tables, independently and before finding this PR. Same fix from two directions.

PR #1161, opened 2026-09-17
Measure an absolute box's top/left from the padding edge

### LINKED ISSUE #1160: An absolute box's top/left are measured from the containing block's content edge, while right/bottom use its padding edge
## Summary

`top` and `left` on an absolutely positioned box are resolved against the nearest positioned ancestor's **content** box, while `right` and `bottom` are resolved against its **padding** box. When that ancestor has padding, the same box therefore lands in two different places depending on which pair of offsets you use, and only the `right`/`bottom` pair agrees with a browser.

`docs/html-css-support.md` already states the correct rule — "An absolute box's **containing block** is the padding box of the nearest positioned ancestor ([CSS 2.1 §10.1](https://www.w3.org/TR/CSS21/visudet.html#containing-block-details))" — so this is the implementation disagreeing with the documented behaviour, not an undocumented gap.

## Repro

```html
<!DOCTYPE html><html><head><meta charset="utf-8"><style>
@page { size: A4; margin: 20pt }
body { font-family: Arial, Helvetica, sans-serif; font-size: 10pt; margin: 0 }
.rel   { position: relative; width: 300pt; height: 60pt; padding: 30pt 40pt; border: 10pt solid #000; margin-bottom: 20pt; }
.nopad { position: relative; width: 300pt; height: 60pt; padding: 0;        border: 10pt solid #000; }
.tl { position: absolute; top: 0; left: 0; }
.br { position: absolute; bottom: 0; right: 0; }
</style></head><body>
<div class="rel"><span class="tl">TL</span><span class="br">BR</span></div>
<div class="nopad"><span class="tl">CTL</span><span class="br">CBR</span></div>
</body></html>
```

Text origins in points, Chrome 141 `--headless --print-to-pdf` against PeachPDF `0336f3db`, both on the same font set:

| run | offsets | Chrome | PeachPDF | delta |
| --- | --- | --- | --- | --- |
| `TL` | `top: 0; left: 0`, ancestor padded | (30.0, 29.95) | (70.0, 60.0) | **(+40.0, +30.05)** |
| `BR` | `bottom: 0; right: 0`, ancestor padded | (396.11, 138.7) | (396.11, 138.75) | (0.0, +0.05) |
| `CTL` | `top: 0; left: 0`, ancestor unpadded | (30.0, 189.7) | (30.0, 190.0) | (0.0, +0.3) |
| `CBR` | `bottom: 0; right: 0`, ancestor unpadded | (308.89, 238.45) | (308.89, 238.75) | (0.0, +0.3) |

The `TL` delta is exactly the ancestor's `padding-left` (40pt) and `padding-top` (30pt). Remove the padding and it goes away, which is the `CTL` row.

The containing block here is the padding box (10,10)–(390,130): a 300×60pt content box plus 30/40pt padding and a 10pt border, at the page origin. Chrome puts `top: 0; left: 0` at (30, 30) — the padding-box corner offset by the 20pt page margin — and PeachPDF puts it at the content-box corner instead.

## Cause

`CssBox.cs`'s absolute branch reads `ClientLeft`/`ClientTop`, and its comment says those are the padding edge:

```csharp
// ... are measured
// from the containing block's PADDING edge (ClientLeft/ClientTop - inside the
// border), not its border-box edge (Location.X/Y) ...
var left = nearestPositionedAncestor.ClientLeft + child.ActualMarginLeft + ...
```

But `CssBox.StyleProperties.cs` defines them as the **content** edge:

```csharp
public double ClientLeft => Location.X + ActualBorderLeftWidth + ActualPaddingLeft;
```

so the padding is added twice over. The padding edge is `Location.X + ActualBorderLeftWidth`. `right`/`bottom` do not go through this path, which is why they were already correct. (The name is worth noting on its own: the DOM's `element.clientLeft` is the border width and the client rect is the padding box, so the name reads as the padding edge and is the content edge.)

## Where it shows up

A chart overlay of the common shape — a `position: relative` plot area with padding, holding an absolutely positioned goal-line label pinned with `top`/`right` — draws the label one padding-top too low. In a real report that put a 6pt-tall tag directly on top of a neighbouring data label, printing two values over each other. It is not a regression: `a46a17e` (2026-09-10) and current `main` place the fixture's `TL` identically at (70.0, 60.0).

## Also differs, and not covered by the above

In the same padded ancestor, `width: 50%` on an absolute child resolves to 150pt (50% of the 300pt content box) where Chrome gives 189.75pt (50% of the 380pt padding box), and `top: 50%; left: 50%` puts the child at (270.0, 60.0) where Chrome gives (220.0, 89.95). The percentage base is a separate code path (`ActualWidth`/`ActualHeight` rather than the origin), so I have kept it out of the fix and am reporting it here rather than guessing at it.

dchupp:
## Summary

`top` and `left` on an absolutely positioned box were resolved against its containing block's **content** edge, while `right` and `bottom` were resolved against its **padding** edge. A positioned ancestor with padding therefore placed the same box in two different places depending on which pair of offsets you used, and only the `right`/`bottom` pair agreed with a browser. CSS 2.1 [§10.1](https://www.w3.org/TR/CSS21/visudet.html#containing-block-details) makes the containing block the padding box, and `docs/html-css-support.md` already documents it that way — this is the code disagreeing with its own docs.

The branch already intended the right thing; its comment reads *"measured from the containing block's PADDING edge (ClientLeft/ClientTop — inside the border)"*. `ClientLeft` is not the padding edge:

```csharp
public double ClientLeft => Location.X + ActualBorderLeftWidth + ActualPaddingLeft;
```

That is the **content** edge, so the ancestor's padding was added twice over. The name invites exactly this mistake — the DOM's own `element.clientLeft` is the border width and its client rect is the padding box — so the invariant note added here says so in as many words for the next reader.

## Measurement

Chrome 141 `--headless --print-to-pdf`, same font set, `position: relative` box with `padding: 30pt 40pt; border: 10pt`. Absolute child's text origin, in points:

| offsets | Chrome | before | after |
| --- | --- | --- | --- |
| `top: 0; left: 0` | (30.0, 29.95) | (70.0, 60.0) | (30.0, 30.0) |
| `bottom: 0; right: 0` | (396.11, 138.7) | (396.11, 138.75) | (396.11, 138.75) |

The `before` delta is exactly `padding-left` (40pt) and `padding-top` (30pt). With the ancestor's padding set to zero, every row agrees before and after.

## Why no existing test caught it

**Every** fixture in `AbsolutePositioningIntegrationTests` sets `padding: 0` on the positioned ancestor, which makes the padding edge and the content edge the same point. The full suite passes identically with and without this fix — 12,211 tests either way — so a green run was never going to find it. The three new tests are written as a bracket: a padded fixture that fails against unmodified code, its unpadded twin that must not move, and the `bottom`/`right` case that was already correct and must stay so.

## Test plan

- [x] Mutation check: reverting `CssBox.cs` alone, with the new tests in place, fails `AbsoluteZeroInsets_AnchorAtTheAncestorsPaddingEdge_NotItsContentEdge` with `Actual: 50, Expected: 10` — the 40pt padding-left. The other two pass both ways, as designed.
- [x] `PeachPDF.Tests` on net8.0: **12,214 passed, 0 failed, 9 skipped** (12,211 before the three new tests).
- [x] `dotnet build PeachPDF.Tests/PeachPDF.Tests.csproj -t:Rebuild`: 6 warnings, all pre-existing CS1998s on `main`, none added.
- [x] `diff-cover` against `origin/main`: **100%** on the 4 coverable added lines in `CssBox.cs`.
- [x] Verified against a 26-document corpus of real print documents rendered through a downstream service and compared to a frozen Chrome baseline: the one document affected by this defect goes from `review` to `pass`.

Caveat on the totals: this box has SDK 9, so the suite was run single-TFM on net8.0 via a local retarget, never committed. CI's numbers will differ. Nothing in the change is TFM-conditional.

## Deliberately out of scope

Two neighbouring behaviours differ from Chrome in the same fixture and are **not** touched, because they resolve through `ActualWidth`/`ActualHeight` rather than through the origin computed here:

- `width: 50%` on the absolute child resolves against the ancestor's 300pt content box (150pt) where Chrome uses its 380pt padding box (189.75pt);
- `top: 50%; left: 50%` places it at (270.0, 60.0) where Chrome gives (220.0, 89.95).

Both are recorded on #1160 with their numbers rather than guessed at. `docs/html-css-support.md` already claims the padding box for percentages, so the docs are ahead of the code there and no doc change belongs in this PR.

Fixes #1160


jhaygood86 (approved):
Verified independently rather than trusted at face value, and every number checked out.

**Spec accuracy**: fetched CSS 2.1 §10.1 directly — confirms the containing block for an absolutely positioned box with a positioned ancestor is that ancestor's padding edge. Both offset pairs must agree.

**Chrome ground truth, independently gathered**: rather than reuse the PR's own numbers, drove real Chrome via CDP `Page.printToPDF` on my own fixture — got (15.0, 15.0)pt for `top:0;left:0` and (390.0, 120.0)pt for `bottom:0;right:0` against a padded, bordered ancestor, exactly the padding edge in both cases. (Traced the PR's own reported ~0.05pt residual to a Chrome border-width device-pixel-snapping artifact, not a CSS semantics question — negligible.)

**PeachPDF output matches**: built the CLI, rendered the identical fixture, measured the PDF's own vector geometry directly — byte-for-byte identical to the independent Chrome measurement. Rasterized with both MuPDF and PDFium, visually confirmed.

**The `ClientLeft`/`ClientTop` naming-trap claim**: read the actual property definitions — confirmed they really do add both border and padding (the content edge), matching the PR's characterization exactly, and confirmed the fix's replacement (`Location.X + ActualBorderLeftWidth`) is the correct padding edge. Also confirmed the pre-existing `right`/`bottom` branch already algebraically reduces to the same correct padding edge, so the asymmetry claim is real.

**Mutation test**: reverting just `CssBox.cs` reproduces the exact claimed failure (`Expected: 10, Actual: 50`); the other two new tests are unaffected either way, as designed.

**No regressions**: full `AbsolutePositioningIntegrationTests` plus a broader positioning/Acid2/fixed/sticky sweep (175 tests) all pass. Full suite: 12,214 passed, 0 failed, 9 skipped — reran myself, exact match.

**Percentage-offset out-of-scope claim confirmed genuine**: traced the percentage width/height resolution path — it goes through `PercentageBase`/`ClientRight - ClientLeft` (the content box), a completely separate code path from what this PR touches, confirming zero overlap and that folding it in here would've been out of scope.

Build: 0 warnings, 0 errors (reran myself). Diff coverage 100%.

One non-blocking suggestion: the out-of-scope percentage-width/height gap (already measured with exact numbers in this PR's own description) currently lives only in a `recent-fixes/` note that expires in 30 days — worth a permanent `.claude/accepted-gaps/` file referencing #1160 so those numbers aren't lost. Not a blocker.

Clean, well-verified fix. Merging.

PR #1162, opened 2026-09-17
Share border and outline edge painting

### LINKED ISSUE #1155: Share border and outline edge painting and support rounded outlines
CSS borders and outlines currently use separate painting paths, which makes their rendering diverge. Solid and double outlines can show corner seams from separately painted sides, and outlines do not consistently follow border-radius. Rounded groove, ridge, inset, and outset outlines can expose a square inner contour even though the equivalent rounded border is curved. Expected behavior: outlines should use the same neutral box-edge painter as borders; solid and double rings should paint without corner seams; all outline styles should follow border-radius, including their inner contour; and border-specific CSS resolution should remain separate from shared edge geometry. Add showcase coverage for every rounded outline style and for elements that have both a border and an outline, plus regression tests for the rounded bevel paths.

ChrisVanDijk:
## Summary
- share one neutral edge painter between CSS borders and outlines
- render seamless rounded outline styles, including asymmetric corners and sliced inline fragments
- clamp large negative outline offsets so the outline remains paintable
- consolidate duplicated uniform/general edge geometry and reduce rounded-layer bookkeeping allocations
- extend the outline showcase and rendering regression coverage

## Known limitation (non-blocking)
Wrapped inline outlines retain open edges at internal line breaks. This is a pre-existing limitation, not a regression: `main` already leaves those edges open and does not follow `border-radius` for outlines at all. This PR adds the rounded physical start/end corners and correct clipping, but deliberately does not introduce an outline-specific fragment-joining algorithm.

CSS recommends connected outline parts but leaves the exact fragmented shape undefined; Chromium, Firefox, and Safari also produce different connected shapes. A future outline-specific joining strategy is tracked in #1163.

## Validation
- `dotnet test PeachPDF.Tests/PeachPDF.Tests.csproj --framework net8.0 --collect:"XPlat Code Coverage" --settings PeachPDF.Tests/coverlet.runsettings --results-directory coverage` (12,233 passed, 9 skipped)
- 95% diff coverage against `main`
- `dotnet build PeachPDF.slnx -t:Rebuild` (0 warnings, 0 errors)
- outline showcase rasterized successfully with PDFium and MuPDF

Closes #1155


ChrisVanDijk:
Chrome, Firefox, Safari and PeachPDF all renders the wrapped inline outlines different:

render-outline-border

jhaygood86 (changes requested):
Given this consolidates two painters that got extensively, independently fixed just yesterday (#1126, #1148, #1152), I checked for regressions against all of them first before looking at anything new.

**No regressions against yesterday's work.** Rendered from this PR's branch and from `origin/main` and compared rasterized output (both PDFium and MuPDF):
- #1126's dotted/dashed border corner fitting and mitred-corner seam fix: byte-identical to main.
- #1148's outline dotted/dashed fitted to the ring's outer edge: byte-identical to main — no regression back to centerline-fitting.
- #1152's `outline-style:auto` fixed-width centered ring: visually identical (the tiny pixel diff traces to this PR *extending* the existing seamless-ring treatment to `auto`'s ring too — a genuine improvement, not a regression).
- `double`/`groove`/`ridge`/`inset`/`outset` bevel shading: visually identical; the small diff at corner mitre diagonals is this PR extending the seamless-bevel path to those styles too (previously only `solid`/`double` got it) — confirmed intentional, not accidental.
- Plain `border: 4px solid` (no outline at all): byte-identical to main, confirming the shared-painter refactor didn't disturb ordinary border rendering.
- Targeted test files (border/outline/pen-cache/rounded-border): 163/163 pass. Full suite: 12,233 passed, 0 failed, 9 skipped — matches the PR's claim.

**Blocking: the flagship new capability — sliced/wrapped-inline rounded outlines — is broken, and worse than the disclosed limitation.** Rendering the PR's own test HTML (`border-radius:6pt; outline:2pt solid` on a `<span>` wrapping across three lines) and rasterizing with both PDFium and MuPDF shows this backwards from both the spec and the PR's own description: the outline's true physical edges (top of the first line, left/right/bottom of the outer boundary) render as a barely-visible ~1px hairline instead of the declared 2pt width, while the internal break edges — which should be open/undrawn per the disclosed limitation — instead render as a solid, fully-opaque line, more visible than the real edges.

Root cause: `FragmentPainter.PaintBoxContent` passes `geometry.DecorationRect` (the whole multi-line unbroken strip) as the outline's base rectangle while passing `geometry.ClipRect` (just this line's rect) as the slice clip. That combination is correct for borders, which never spill past their own edge — but an outline spills *outward* by `offset + width`, and `OutlineDrawHandler.DrawOutline`'s resulting clip ends up equal to the unexpanded fragment rect on every side, clipping away nearly all of the outward spill. Confirmed directly in the PDF content stream: the clip and fill geometry disagree by exactly the outline's reach.

The PR's own new test for this (`RoundedOutlineOnAWrappingInlineElement_ClipsOnlyAtBreakEdges`) passes despite the bug, because it only asserts clip-rectangle coordinates and `DrawPathCall` counts — it never rasterizes. This is exactly the trap CLAUDE.md's testing conventions call out for paint changes: a structural assertion isn't proof of correct rendering. Needs a rasterization-based regression test alongside the fix.

**Everything else checks out:**
- Asymmetric rounded corners (`border-radius: 8px 20px 4px 16px` + outline): correct corner-tightness ordering, clean in both solid and dotted variants.
- Negative `outline-offset` clamping: correct — a clamped, visible ring rather than a degenerate/vanished one, mutation-tested (disabling the clamp fails both new theory cases with "no outline rendered").
- Mutation-tested the seamless-bevel extension too: disabling it fails 5 tests across border/outline inset/outset/groove/ridge suites with specific diagnoses.
- Build: 0 warnings, 0 errors. Diff coverage: 95%, matches the claim.

Solid consolidation work overall — the sliced-inline-outline gap just needs to actually match the disclosed "open at breaks, full width at physical edges" behavior rather than the inverse, with a real rasterized test to catch it next time.

ChrisVanDijk:
image

jhaygood86 (approved):
Verified the fix, and found something useful about the bug's real shape along the way.

**The sliced-outline bug is genuinely fixed** — but the fix works differently than the clip-bounding approach I expected, and turned out to reveal the actual dominant cause. Root cause was twofold: a minor `OutlineDrawHandler`-level clip issue (mostly cosmetic on its own), and a much bigger one — the page-level content clip was cutting the outline's outward spill whenever a box's true physical edge sat near the page's content-area edge, which is exactly what my first reproduction attempt with generous margins failed to trigger. The fix removes the per-line clip from `OutlineDrawHandler` entirely (each line now builds its outline from its own rectangle directly, so no clip is needed there) and instead widens the page's content clip by the whole page's `MaximumOutlineReach` before painting, at `FragmentPainter.Paint()`.

Re-tested with the box flush against a tight page margin (reproducing the original bug precisely on the reverted code) and confirmed the fix on both PDFium and MuPDF: physical edges went from a ~3px hairline to the full declared stroke width; at a normal margin, physical edges paint at full width and interior break edges stay genuinely open — matching the disclosed design. Mutation-tested by reverting just this commit's two files: the hairline defect reproduces exactly, on both renderers.

**One real, non-blocking gap worth recording**: the updated test (`RoundedOutlineOnAWrappingInlineElement_UsesEachFragmentAsItsBase`) is a genuine improvement — it now asserts against actual recorded path geometry, not just clip-rectangle coordinates. But it uses `FragmentPaintHarness.PaintBox`, which is documented as painting without the surrounding page clip — so it structurally cannot exercise `FragmentPainter.Paint()`'s `MaximumOutlineReach` widening, which is the dominant real-world fix here. Confirmed `MaximumOutlineReach`/`OutwardReach` are never referenced anywhere in the test project. The 95% diff-coverage number reflects line execution via other full-`PdfGenerator` outline tests, not behavioral verification of this specific mechanism — worth a lightweight follow-up: a `PdfGenerator`-based integration test with a box flush against a tight page margin, asserting on the raw content stream that the outline's drawn extent reaches past the nominal content-area boundary. Not blocking since I independently verified the fix works via dual-renderer rasterization myself, but a future "simplification" of `Paint()` that drops the widened clip would sail through the current suite untouched.

**Regression check against everything confirmed correct in the last two rounds** — all still intact: #1126's dotted/dashed corner fitting and mitred-corner seam fix, #1126's bevel shading (including the black-color lightening case), #1148's outline outer-edge fitting, #1152's outline-auto fixed-width ring, asymmetric rounded corners, negative-offset clamping.

Build: 0 warnings, 0 errors (reran myself). Tests: 12,233 passed, 0 failed, 9 skipped (reran myself, exact match). CI green on all three platforms.

Good, simpler fix than I expected going in — merging.

Issue #1203, opened 2026-09-18
A multi-column container's floated children are never laid out (silently dropped)


jhaygood86:
## What happens

\`CssLayoutEngineColumns.Layout\`'s own item list is built with:

\`\`\`csharp
var children = columnsBox.Boxes
    .Where(b => b.DerivedStyle.ActualDisplay != Keywords.None && !b.IsExcludedFromFlow
                && (b.HtmlTag != null || !b.IsSpaceOrEmpty))
    .ToList();
\`\`\`

\`IsExcludedFromFlow\` is \`IsOutOfFlow || IsRunningPositioned\`, and \`IsOutOfFlow\` is
\`IsFloated || Position is Absolute/Fixed\`. So a floated child of a \`columns\`/\`column-count\`/
\`column-width\` container is filtered out of \`children\` entirely — unlike an absolutely/fixed
positioned child, nothing compensates by laying it out through a separate out-of-flow pass. If a
multicol container's children are *all* floats, \`children.Count == 0\` and the container returns
immediately, leaving every floated child at its default, never-assigned geometry (\`Location\`,
\`ActualBottom\`, \`ActualRight\` all zero).

## Why it's a bug, not a deliberate exclusion

\`IsOutOfFlow\` correctly excludes a float from *this engine's own column-distribution item list*
(a float isn't one of the boxes columnized independently — CSS Multi-column Layout Module Level 1
still expects a float to establish its own float context and float within the column it's placed
in), but nothing lays the float out afterward the way absolutely/fixed-positioned descendants are
handled elsewhere via a dedicated out-of-flow layout pass. The float is dropped rather than
floated.

## Repro

\`\`\`html
<div style="columns:2; column-gap:10px; width:200px">
    <div style="float:left; width:200px; height:400px;"></div>
    <div style="float:left; width:50px; height:20px;"></div>
    <div style="float:left; width:50px; height:20px;"></div>
</div>
\`\`\`

All three floated divs come out at \`Location=(0,0)\`, \`ActualRight=0\` — never positioned at all.
Verified present identically on \`main\` before issue #1038's dispatch changes, so it predates and
is unrelated to that work.

## Scope note

Found and confirmed pre-existing while adding multi-column regression coverage for #1038 (which
only had to confirm dispatch still reaches this engine at all for a floated-only-children
container, not that the engine positions those children correctly). Left out of scope there;
tracked here for a real fix — likely giving floats their own layout pass per column, mirroring how
[...]

Issue #1201, opened 2026-09-18
A float's own content taller than one page overflows rather than continuing across pages


jhaygood86:
## Summary

Since the #1038 fix (a float that follows inline content now shares one line with the text before and after it, laid out through `CssLayoutEngine.FlowFloatChild`), a float's own content is positioned via `CssBox.LayoutContentAtItsAssignedPosition` - the same entry point `CssLayoutEngine.FlowAtomicBlockContentChild` already uses for an inline-block holding real block-level content.

Neither call site threads a nested `PendingBreakToken` any further: if the float's (or the inline-block's) own content is taller than the remaining fragmentainer, the content that does not fit is never laid out at all - it is silently dropped rather than continuing onto a later page. This is CSS Fragmentation Level 3 §2 "monolithic content" territory in spirit, but unlike an explicitly-monolithic box (a replaced element, a scroll container), this is a byproduct of the current implementation not threading the resumption record through this call path, not a deliberate application of §2 to a genuinely unbreakable box.

## Where it happens

- `src/PeachPDF/Html/Core/Dom/CssLayoutEngine.cs`, `FlowFloatChild` (places a float encountered mid-inline-flow) and `FlowAtomicBlockContentChild` (places an inline-block with block-level content) both call `CssBox.LayoutContentAtItsAssignedPosition`, and neither checks the child's own `PendingBreakToken` afterward or threads it into the surrounding `InlineBreakToken`/`CssLineBoxCoordinates.Break` mechanism the rest of `FlowBox` uses for pagination.

## What would be needed to fix it

Extending `InlineBreakToken` (or an equivalent mechanism) to carry a nested per-child continuation record - comparable to `BlockBreakToken.ChildToken`'s existing "parallel flow" shape - so that a later fragmentainer pass can resume laying out exactly the child (float or inline-block) that stopped, without re-placing or duplicating content already committed. This is a non-trivial addition since `FlowBox`'s own per-word ordinal-based resumption model does not currently have a way to represent "resume this specific non-word child's own content, without re-deriving its position."

## Scope

Left out of #1038's own scope: that fix is specifically about a float's placement (sharing a line with surrounding inline content), not about extending this engine's pagination model to cover floats/inline-blocks whose own content spans multiple pages. Tracked separately here per this repo's convention for accepted gaps.

PR #1205, opened 2026-09-18
Fix float after inline content placed on the block edge instead of sharing its line

### LINKED ISSUE #1038: A float that follows inline content is placed on the next line, not beside it

jhaygood86:
## Summary

- A floated element that follows inline content in source order (e.g. `XY <span style="float:left">Z</span> more`) was placed at the block's edge on top of the preceding text, and the content that followed it was pushed onto a new line entirely — instead of sharing one inline formatting context with the surrounding content, per CSS 2.1 §9.5/§9.5.1 rule 6. A float that precedes inline content already worked correctly.
- `DomParser`/`DomUtils` now let a float join the same inline run as surrounding content (`JoinsTheInlineRun`, `ContainsInlinesOnly`, `ContainsVariantBoxes`, and the block/inline-parent correction walks), and `CssLayoutEngine.FlowBox` gained a real dispatch branch (`FlowFloatChild`) that positions a float encountered mid-flow from the live inline cursor and lays its content out through the same entry point an inline-block with block-level content already uses.
- Along the way, this also fixes: an infinite-regress stack overflow in `ContainsVariantBoxes` for a lone floated child, a real infinite recursion via `FlowBox`'s self-iteration convention, a trailing-space undercount in `GetMinMaxSumWords`'s float branch, a horizontal point-collision gap in the new inline-float intersection check, a multi-column dispatch regression from widening `ContainsInlinesOnly`, and (found by an independent review pass) a float's own left-side box spacing leaking into the next sibling's cursor position.
- Two narrower gaps are left out of scope and recorded under `.claude/accepted-gaps` with tracking issues: a float's own content taller than one page overflows rather than continuing onto a later page (same limitation an inline-block with block content already has — issue #1201), and a multi-column container whose children are all floats still renders no visible content for them, a separate, pre-existing `CssLayoutEngineColumns` gap confirmed unrelated to this change (issue #1203).

## Test plan

- [x] Full `net8.0` suite green (12,357 passed / 0 failed / 9 skipped)
- [x] 100% diff coverage on changed production lines (`diff-cover` against `origin/main`)
- [x] `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings across net8.0/net10.0/net11.0
- [x] New regression coverage: the accepted-gap's own repro, two adjacent floats with nothing between them, text wrapping on both sides of a float, the `float:right` counterpart, a float taller than one page, a page-boundary resume with a float present, a multi-column dispatch-routing guard, a float immediately followed by a forced break, and a float with left-side box-model insets amid inline content (the review-pass finding)
- [x] `docs/html-css-support.md` and a migration note updated for the behavior change

Fixes #1038


jhaygood86:
Reopening to retrigger CI - the Tests workflow did not fire on the initial push.

Issue #1207, opened 2026-09-18
Wrapped-inline outline union omits atomic-inline descendants' border boxes


ChrisVanDijk:
PeachPDF now unions a wrapped inline's line rectangles into one connected outline shape (`OutlineRegionPainter` / `RectilinearRegion`), matching Chromium's `ComplexOutlinePainter`.

Chromium feeds one more kind of rectangle into that same union that PeachPDF does not: the border box of each **atomic-inline descendant** - an `<img>`, an `inline-block`, an inline-level flex/grid container, a replaced element. In Blink these are collected by `OutlinePainter`'s descendant walk and unioned alongside the line rectangles, so an atomic inline that is taller than its line box pushes the outline outward around itself.

PeachPDF unions only the outlined inline's own per-line rectangles. Where an atomic-inline descendant overflows its line box, PeachPDF's outline follows the line box and Chromium's bulges around the descendant.

**Spec rule.** CSS Basic User Interface 4 section 5.2 says the outline is drawn around the border edge of the element's fragments. An atomic inline's own border box contributes to the union in Chromium's interpretation of what that element covers; PeachPDF's narrower reading leaves a visible difference only when the descendant is taller than the line that holds it.

**Why it was out of scope.** The paint-side outline entries are collected per line fragment in `FragmentPainter.PaintBoxContent`. Including descendant border boxes means a fragment-tree descent from the outlined box, gathering each atomic inline's `BoxFragment` rectangle, at the same collection site - a wider change than the geometry work this was scoped to, and one that interacts with the stacking-order hoist rules for those descendants.

**Scope if picked up.** A descendant walk at the outline collection site, plus the decision of which descendant kinds qualify (Chromium's rule is atomic inlines only, not every inline descendant). `RectilinearRegion.Union` already accepts arbitrary rectangles, so the geometry side needs no change - it is purely about what is fed in.

Documented in `docs/html-css-support.md`.

Issue #1206, opened 2026-09-18
Patterned and bevelled outline styles do not union across a wrapped inline's lines


ChrisVanDijk:
When an inline element's `outline` wraps across several lines, PeachPDF now unions the lines' rectangles and traces one connected shape around the region they cover, matching Chromium (`OutlineRegionPainter` / `RectilinearRegion`).

That union is currently limited to `solid`, `double` and `auto`. The patterned and bevelled styles - `dotted`, `dashed`, `groove`, `ridge`, `inset`, `outset` - fall back to the previous geometry: a closed ring per line.

**Spec rule.** CSS Basic User Interface 4 section 5.2 (`outline`) says the outline of a fragmented box must be drawn as if the fragments were joined, with no exemption by `outline-style`. The fallback is therefore a real deviation for those six styles, not an area the spec leaves open.

**Why it was out of scope.** Each of the six needs something the merged region does not yet define:

- `dotted`/`dashed` need a dash pattern fitted to the whole merged boundary. PeachPDF fits dashes per straight run against the ring's outer edge (`BoxEdgesDrawHandler`); a merged contour has concave corners and runs of arbitrary length, so the fit has to be re-derived over a closed polyline rather than per side.
- `groove`/`ridge`/`inset`/`outset` colour each side light or dark. A merged region has no four sides - a concave corner has two edges facing the same way - so ""which side is this"" has no answer without a new per-edge normal-to-shade rule. Chromium picks these bevel colours per rectangle side too.

Falling back keeps those six rendering exactly as they did before rather than rendering them wrongly.

**Scope if picked up.** Dash fitting over a closed polyline, and a per-edge bevel-shade rule defined on edge normal rather than on box side. Both are additive to `OutlineRegionPainter`; no change to `RectilinearRegion`.

Documented in `docs/html-css-support.md`.

ChrisVanDijk:
Agent is working on this

PR #1208, opened 2026-09-18
Union a wrapped inline's line rectangles into one outline shape

ChrisVanDijk:
An inline element whose `outline` wrapped across several lines drew a separate closed ring per line. Chromium instead unions the line rectangles and traces one connected shape around the region they cover. This implements that.

The difference shows up most obviously when the inline also has a `border`:

| before | after |
|---|---|
| a closed ring around each line | one connected shape around all of them |

## The border is not the trigger

It looks like the border causes the merge, but it doesn't. Chromium unions whenever there is more than one outline rectangle:

```cpp
if (*united_outline_rect == pixel_snapped_outline_rects[0]) { /* simple per-rect path */ }
else ComplexOutlinePainter(...).Paint();
```

A border simply makes each line's border box a few px taller, so consecutive line rectangles start **overlapping vertically** and the union collapses into one contour. Remove the border and the same union yields three *disjoint* contours, which coincides visually with the old per-line rings. Same algorithm, different topology - so this consults the border nowhere. `line-height` and `outline-offset` each cause the merge on their own, with no border present; the showcase demonstrates exactly that.

This also corrects a wrong claim in `docs/html-css-support.md`, which stated that a closed ring per line *was* Chromium's shape.

## Implementation

**`RectilinearRegion`** (new, `Html/Core/Utils/`) builds a coordinate grid from the rectangles' own distinct edge positions (epsilon-merged), marks covered cells, and emits boundary edges directed so **the covered side is always on the right of travel**. That invariant is load-bearing: `Shrink`, `DescribeCorner` and `FillBand` all derive from it, and reversing a contour's winding makes `Shrink` silently *expand* it. `Shrink` insets a contour by moving each corner along the sum of its two adjacent edges' inward normals, which is correct for convex, reflex and hole corners alike.

**`OutlineRegionPainter`** (new, `Html/Core/Handlers/`) turns contours into filled bands - one band for `solid`, two concentric for `double` - and applies `border-radius` with css-backgrounds-3 section 5.1 proportional scaling.

Two deliberate choices worth flagging for review:

- **Nonzero winding, not even-odd.** `BoxEdgesDrawHandler.DrawRectangularRing` uses even-odd for simple rings, but an inset contour can self-cross where a neck is narrower than `2 x width`. Under nonzero a collapsed neck fills solid (correct); even-odd would punch a hole through it.
- **Chromium's integer pixel-snapping is skipped.** There is no device-pixel grid here, and using the rectangles' own edge positions as grid lines is exact in point space.

The paint-side seam already existed - `FragmentPainter` was already collecting outline paint entries into a deferred list and draining them in one pass - so this is a branch at the drain site rather than a paint-walk rewrite.

## Scope

Limited to `solid`, `double` and `auto`. Two gaps are documented and tracked rather than silently left:

- #1206 - the patterned and bevelled styles (`dotted`, `dashed`, `groove`, `ridge`, `inset`, `outset`) keep a ring per line: a dash fit over a closed polyline and a per-edge bevel-shade rule are not defined for a merged region's concave corners.
- #1207 - the union covers the inline's own line boxes, not atomic-inline descendants' border boxes, which Chromium also feeds in.

Neither is closed by this PR. Chromium's cosmetic style degradations (`double` at width <= 2, `groove`/`ridge` at 1px) are not ported, and the existing vertical-writing-mode exclusion is preserved.

Page and column breaks are unaffected: the region is always built from one fragmentainer's own rectangles, so a box broken across pages gets its own shape on each.

## Verification

- 237 outline + geometry tests pass; full suite 12521 passed (one pre-existing unrelated table-layout failure, confirmed by stashing onto clean `main`)
- Diff coverage: `OutlineRegionPainter.cs` 100%, `RectilinearRegion.cs` 97.7%, `OutlineDrawHandler.cs` 97.5%
- `dotnet build PeachPDF.slnx -t:Rebuild` - zero warnings
- **Rasterized through both PDFium and MuPDF** across 13 scenarios plus the regenerated showcase; the two renderers agree. A translucent-outline case is the sharpest check - stacked per-line rings would double-darken where lines overlap, and the union shows an even band.
- No measurable perf cost on a 4000-word stress fixture (1.99s union vs 1.99s no-outline baseline). The first coverage fill was O(rects^3) and was replaced with direct block marking.

### A note on the tests

`TestGraphicsPath` was flattening every subpath into one point list, so a `DrawPathCall` could not distinguish one connected shape from three disjoint ones - precisely the property under test. That produced three failures that read as algorithm defects and were pure assertion blindness. It now records subpath starts, with `Subpath(i)`/`SubpathBounds(i)` accessors that any future topology assertion should use.


jhaygood86 (approved):
Independently re-verified end-to-end. Approving.

**Build/test (own run, matches PR/agent numbers exactly):**
- `dotnet build PeachPDF.slnx -t:Rebuild` — 0 warnings, 0 errors.
- `dotnet test --framework net8.0` — 12523 passed, 0 failed, 9 skipped.
- Diff coverage on the four changed production files: 100% (`OutlineDrawHandler.cs`, `OutlineRegionPainter.cs`, `FragmentPainter.cs`, `RectilinearRegion.cs`).

**Claims independently verified:**
1. **No-border union claim** — confirmed by rasterization. A no-border/tight-line-height wrapped inline unions into one connected contour; the same case with generous line-height (no overlap) stays disjoint. Border only ever changes topology by changing whether line rectangles overlap — never a separate code path.
2. **Nonzero winding for the collapsed-neck case** — swept outline width from 6pt to 20pt on a wrapped three-line case; the neck collapses at ~11–12pt (matches the analytical prediction from the middle line's height), and above that width the corner fills solid with no hole, no seam. PDFium and MuPDF agree pixel-for-pixel.
3. **Issues #1206/#1207** — both exist, open, and match what the accepted-gap files describe.
4. **Migration note replacement** — confirmed via `git merge-base --is-ancestor` that the old per-line-ring behavior (commit `abd83c53`) landed *after* `v0.9.18` and never shipped, so replacing rather than appending the migration note is correct; the new note is written against the last *released* behavior, not the unreleased intermediate state.
5. **Regression check** — rendered a simple non-wrapped outline+border+radius case against this PR and its parent commit; rasterized output is pixel-identical, consistent with `FragmentPainter.cs`'s new branch only engaging when there's more than one outline rect to union.
6. **Translucent-outline / double-darkening check** — rendered; the merged region shows one uniform band with no darker seam at the former ring boundary, in both renderers.

**Two non-blocking issues found — spec section numbers, not behavior:**
- `.claude/accepted-gaps/patterned-and-bevelled-outline-styles-do-not-union-across-lines.md` and `.claude/accepted-gaps/wrapped-inline-outline-union-omits-atomic-inline-descendants.md` both cite css-ui-4 **§5.2** for the "outline drawn as if fragments were joined" rule. I fetched the current spec: that text is in **§3.1** ("Outlines Shorthand"); §5.2 is "Insertion caret" (unrelated).
- `OutlineRegionPainter.cs`'s doc comments cite css-backgrounds-3 **§5.1** for the corner-radius-overlap proportional-reduction rule (should be **§4.5**, "Overlapping Curves" — §5.1 is border-image-source) and **§4.3** for the `double`-style two-lines-plus-gap rule (should be **§3.2**, "Line Patterns" — §4.3 is corner clipping).

Neither affects rendered output — comment/doc-only — but this repo's convention treats spec citations as durable reference material, so worth a quick follow-up fix at your convenience; not blocking merge given CI is green and behavior is verified correct.

Also confirmed: no regression to the adjacent #1126/#1148/#1152/#1161/#1162 border/outline work (single-line case is byte-identical pre/post), and the `TestGraphicsPath` subpath-tracking fix is real and necessary — the old test double genuinely couldn't distinguish one connected shape from several disjoint ones.

PR #1216, opened 2026-09-18
Union patterned and bevelled outline styles across a wrapped inline's lines

### LINKED ISSUE #1206: Patterned and bevelled outline styles do not union across a wrapped inline's lines

ChrisVanDijk:
Closes the gap deliberately left by #1208: an outline on a wrapped inline unioned its line boxes into one connected shape, but only for `solid` and `double`. The other six styles - `dotted`, `dashed`, `groove`, `ridge`, `inset`, `outset` - still painted one closed ring per line, on the assumption that a dash fit and a per-side bevel colour are undefined over a merged region's concave corners.

Blink does not degrade them. `ComplexOutlinePainter` handles every non-`auto` style over the unioned region, with `NOTREACHED()` as its default arm, and only takes a simple path when the union *is* the first rect. So this paints them on the merged shape instead: the dash pattern is fitted along each boundary edge of the region, and each edge's bevel face comes from the direction that edge runs in - the rule that still works once the lines have merged and a per-side one no longer does. `SupportsRegionOutline` becomes the single gate, now `is not (None or Hidden)`.

Nothing new was needed at the drawing layer: `StyledStrokeFitting.FitClosed` already does closed-path dash fitting and `BorderBevelColors.ForSide`/`ForSegment` already produce the contrast-aware faces. The feature is geometry plus dispatch.

### The bevel geometry

The clips are derived from the region's own right-angled corners, not from the rounded path the band is drawn along - an arc has no single direction of travel and so no shade of its own. Each edge gets the territory between the two 45-degree mitres it shares with its neighbours, accumulated one clip per shade so that same-shade neighbours don't abut two antialiased fills along a mitre (the seam `BoxEdgesDrawHandler.TryDrawSeamlessBevel` exists to avoid on the per-side path).

Three properties of those mitres turned out to be load-bearing, and each was found by rasterizing rather than by reading. Each drew as an unpainted slash or wedge across the band, and each survived a green suite:

- **A mitre leans the way its corner turns.** A convex corner is the outside of the turn so its mitre opens outwards; a reflex corner's leans the other way. Leaning them all the convex way leaves the inner half of every reflex corner - the notch between two lines - in neither edge's clip.
- **It has to reach past the straight run by the corner's radius.** Near a rounded corner the band deviates from the straight centreline by up to the radius, so a clip only `halfWidth` deep leaves a wedge of every arc uncovered.
- **It has to stop where the two mitres cross.** Both are cut from the edge's own line, so on any edge shorter than twice the reach they cross and the quad folds into a bowtie whose waist cancels under nonzero winding. A wrapped inline's staircase steps are routinely that short.

Deliberately not copied from Blink: its width-based degradations (`double` at width <= 2, `groove`/`ridge` at width == 1). There's no width-based style degradation on this repo's per-side border path either, so adding it only here would make the two disagree.

### Verification

The decisive test was mask-diffing each shape against the identical shape drawn `solid` - `solid` paints the true band, so any pixel it paints and the bevel leaves white is an uncovered wedge. Over a sweep of width x radius x style, in **both PDFium and MuPDF**, no unpainted pixels remain (the showcase swatch went 1121px -> 0). Cross-checked against real headless Chrome, including tight crops of the notch.

Three new tests, each confirmed to fail without its fix: rounded-bevel arc coverage, point-level band coverage, and a clip-convexity test. The last one is separate because a coverage test is structurally incapable of catching a bowtie - a ray cast from inside the cancelled waist still crosses the outline an odd number of times.

Full suite green, solution rebuilds with 0 warnings, 97% diff coverage. Docs, showcase and the accepted-gap note are updated in the same change.

Fixes #1206


jhaygood86 (changes requested):
Thorough review done; most of this checks out, but there is one blocking rendering regression and some stale/incorrect documentation to fix.

## What checks out
- `dotnet build PeachPDF.slnx -t:Rebuild`: 0 warnings, 0 errors. net8.0 suite: 12563 passed, 0 failed, 9 skipped. Diff coverage 97% (the 4 missing lines are the documented unreachable pen-reset fallback).
- Regression vs. main: single-line outlines in all nine styles, wrapped solid/double/auto, vertical writing-mode and column-broken blocks are byte-identical to main (0 diff pixels). Blink's `ComplexOutlinePainter` claims (every non-auto style, `NOTREACHED` default, width degradations not ported) are accurate.
- Coverage claim holds: my own sweep (288 cells x 4 bevel styles x PDFium+MuPDF, mask-diffed against `solid`) finds 0 unpainted pixels. All three mitre-geometry fixes are individually mutation-tested: removing each fails its named tests and reproduces the claimed symptom.

## Blocking: wrong bevel shade where edges sit within `reach` of each other (regression vs. main)
`OutlineRegionPainter.Styled.cs:245-246` sets `reach = halfWidth + max(corner radius)` and `AddMiterClip` extends each edge's clip `+-reach` perpendicular to its centreline. The lit and shaded fills both go through the *same* band path, shaded second, so any band pixel that falls inside another edge's over-wide clip is repainted in that edge's shade. It shows whenever a rounded outline has another boundary edge within about `outline-width/2 + border-radius` of an edge's centreline: separate line pieces close together, short staircase steps, narrow necks. It does not occur at `border-radius: 0` (reach is just `halfWidth`). The mask-diff against `solid` and the three new tests only check *coverage*, so a wrong-shade repaint passes them; the clip-convexity test checks polygon simplicity, not overlap between lit and shaded clips.

Repros (each is correct on main or in Chrome, wrong on this branch, in both PDFium and MuPDF):

1. Separate pieces (a regression vs. main, whose per-line rings are right here). A light sliver appears along the outer top edge of "Beta" and "Gamma":
```html
<div style="width:260px;line-height:2;font:16px Arial;padding:14px"><span style="outline:6px groove #4a90d9;border-radius:12px">Alpha<br>Beta<br>Gamma</span></div>
```
2. Staircase. A light bar overwrites the dark run of the third-line notch (also `groove` at 6px, radius 16, offset 0):
```html
<div style="width:260px;line-height:1.2;padding:14px"><span style="outline:10px inset #4a90d9;outline-offset:-3px;border-radius:16px">wwwwwwwwwwwwwwww<br>wwwwwwwwwwwww<br>wwwwwwwwww<br>wwwwwwwwwwwwwwwwwwww<br>ww</span></div>
```
3. Collapsed neck plus radius: lines `wwwwwwwwwwww` / `i` / `wwwwwwwwwwww` with `outline:8px inset`, `border-radius:12px`, `outline-offset:2px` has the left edge half light and the neck's lower edge light; Chrome has both dark. Same with `outline:10px` in `inset`/`outset`.

Suggested fix (not verified): clamp the clip's perpendicular extent to `halfWidth` along the straight run and grow it toward the radius only within about one radius of each edge end (a hexagonal clip), so an edge's clip never reaches a different edge's band. Building lit/shaded clips per contour would additionally stop cross-contour leaks. Please add a test that lit and shaded clips do not overlap inside the band, or a raster-level test using the repros above.

## Also please fix (documentation accuracy)
The PR does not add new section citations but leaves the wrong ones from #1208 in files it touches. I fetched css-ui-4 and css-backgrounds-3: the "outline or minimum set of outlines ... fully connected" text is css-ui-4 **§3.1**; outline-style / "width ignored when auto" is **§3.3**; the layout-neutral / outline-offset text is **§3.5**; css-backgrounds-3 **§3.2** is Line Patterns and **§4.5** is Overlapping Curves. Current text is wrong at:
- `OutlineRegionPainter.cs:17` ("CSS UI 4 §4" -> §3.1), `:80` ("Backgrounds 3 §4.3" for `double` -> §3.2), `:233` ("§5.1" for radius scaling -> §4.5)
- `OutlineDrawHandler.cs:13` (-> §3.5), `:49` (-> §3.1), `:199` (auto ignoring width -> §3.3)
- `.claude/migration-notes/2026-09-18-wrapped-inline-outline-unions-its-lines-into-one-shape.md:11` ("Interface 4 §5.2" -> §3.1; §5.2 is the insertion caret)
- `Paint/FragmentPainter.cs:731,739` ("CSS UI 4 §4" -> §3.1)

Stale statements now made false by this PR:
- `Paint/FragmentPainter.cs:738-742` says only styles "fully determined by which area is filled" take the union path and "the rest keep a ring per fragment".
- `.claude/recent-fixes/2026-09-18-wrapped-inline-outline-unions-its-lines-into-one-shape.md:100` still says patterned/bevelled are "accepted gaps with their own files and issues #1206/#1207"; the #1206 file no longer exists.
- The migration note says "a dash lands on every corner it turns, including the concave ones"; that is only true for square corners (rounded outlines use a closed-path fit, as the `Styled.cs` comment says).

## Non-blocking
- Translucent `dotted`/`dashed` with square corners double-darken the corner squares (Chrome wraps non-solid translucent outlines in an alpha layer). This matches main's existing single-box behaviour, so it is not a regression; a follow-up, if wanted.
- Bevel light shade is brighter than Chrome's; this is pre-existing `BorderBevelColors` behaviour on main's single-box path too.

Happy to re-review as soon as the shading fix lands.

jhaygood86 (changes requested):
Re-reviewed head `43035daa`. The blocking bug from my first review is fixed. One of the new commits, though, makes PeachPDF wrong against current Chromium, so I'm requesting changes again. The rest of the PR is in good shape.

## Fixed / verified
- **Wrong-shade bug (01b282fe):** fixed in both PDFium and MuPDF. Against Edge 153 as ground truth I ran about 3,000 sweep cells (separate pieces, wrap, staircase, neck, tight line-height, 1/2/4px steps, pinch necks, radius up to 60, offsets -8..+6, width 2-14). The buggy commit had 147 of 1728 cells with wrong shade (20,996 px); the head has 72 cells and 759 px, all tiny 45-degree mitre triangles at pinched notches. Coverage is still 0 unpainted / 0 overspill. Repro 2 went from 219 wrong-shade px to 0. Translucent bevels look like Chrome with no overlap blotches.
- **Build/tests/coverage:** 0 warnings; net8.0 12605 passed, 0 failed, 9 skipped; diff coverage 98% (the missing lines are the unreachable `FitClosed` fallback). A 4000-word paragraph is within noise of main.
- **Regression corpus (162 files, PR vs main):** single-line, wrapped solid/double/auto, vertical, column-broken and multi-page outputs are identical to main, except the intended new feature and the lit-colour change below.
- **Spec citations:** most of the list I gave is now correct against the fetched specs (`OutlineRegionPainter.cs`, `OutlineDrawHandler.cs`, the wrapped-inline migration note, the gap files).
- **New accepted gap:** the translucent dotted/dashed corner double-darkening gap references a real, open issue (#1222) whose body matches, and I reproduced the behavior. The docs note is issue-free.
- **New invariant** (`outline-bevel-territory-mapping-inverts-exactly.md`): the claim holds.

## Blocking: commit 5be8fe47 (lit face keeps the declared colour) regresses against current Chromium
Its premise, that Chrome leaves the lit face at the declared colour, is true of Chrome 148 and earlier only. Blink main's `CalculateInsetOutsetColor` (`box_border_painter.cc`) has the `TableDefaultBorderColorCurrentColor` feature marked `stable` ("shipped in M149, updated in M151"). Its rule: dark face = `Dark()`; lit face = the raw colour if relative luminance is above 0.83077, otherwise `Light()`; very dark colours lighten both faces. I measured real pixels in Edge 153 (inset/outset/groove/ridge, borders and outlines):

| colour | Chrome lit face | before 5be8fe47 | at head |
|---|---|---|---|
| `#4a90d9` | `#57a9ff` | matches | `#4a90d9` (wrong) |
| `#808080` | `#d4d4d4` | matches | `#808080` (wrong) |
| `rgba(74,144,217,.5)` over white | `#aad3ff` | matches | `#a4c7eb` (wrong) |
| `#f0f0f0` | `#f0f0f0` | `#ffffff` (wrong) | matches |
| black / white / red / `#101010` | | match | match |

Across 48 random colours plus 12 greys, the exact lit-face match against Chrome is **58/60 before 5be8fe47 and 7/60 at head**; only near-white colours improve. The dark face is identical everywhere. It also alters every bevelled border, outline and collapsed-table segment library-wide, inside a PR about outline unions. The lightening in `BorderBevelColors` shipped in v0.9.19 (`git merge-base --is-ancestor 8d664052 v0.9.19` is true), so this would also be a user-visible change against a released tag, in the wrong direction.

Please drop 5be8fe47 with its migration note and recent-fix. If you want the near-white behavior, add Blink's actual rule in its own change: raw colour when relative luminance > 0.83077, otherwise `Light()`, which should give 60/60. Also note: the test comment in `BorderStylePaintIntegrationTests.cs` ("Sampled from Chrome ... genuinely lightened") was correct for current Chrome and the replacement claims the opposite, and the recent-fix line saying there was no byte-exact measurement is not accurate. `docs/html-css-support.md:310` still says inset "lightens the bottom and right", so the docs now contradict the code.

## Also fix
1. `.claude/migration-notes/2026-09-18-bevel-shade-overreach.md` documents a bug that only existed in unmerged commit `66b3a4a6`, never on main or in a release. It is not a migration note and should be deleted (its content can stay in the recent-fix).
2. `.claude/invariants/outline-bevel-clips-come-from-the-rectilinear-region-not-the-rounded-path.md` still describes the old design (one accumulated clip per shade, reach-wide quads). The code now uses per-contour band fills with one-sided caps.
3. Citations still wrong: `Paint/FragmentPainter.cs:731` ("CSS UI 4 §4" -> §3.1), `OutlineStylePaintIntegrationTests.cs:422` and `:728` ("§4" -> §3.1), `:984` ("css-backgrounds-3 §4.3" -> §3.2).
4. `docs/html-css-support.md:388` still says "a dash lands on every corner it turns, including the concave ones"; this is only true for square corners (rounded outlines use a closed-path fit, as the migration note now says).

## Non-blocking suggestions
- The new shade test fails on the buggy code only for the staircase case; the per-contour rule that fixed the separate-pieces repro is not guarded. Assert the clip-pair count equals the contour count and that no clip of one contour overlaps another contour's band.
- Mutation "all mitres lean convex" fails only inset/outset tests; a groove/ridge convex-lean guard would close that.
- `EmitPolygon` in `Styled.cs` has body indentation two levels too deep.

jhaygood86 (approved):
Re-verified head `ca68e657`. Every item from my last review is addressed; approving.

- **Lit-face commit dropped.** `BorderBevelColors.cs` and `BorderStylePaintIntegrationTests.cs` are untouched relative to main again, so single-box bevel output is unchanged. The near-white gap it was reaching for is now tracked separately as #1224 (issue exists, open, and describes Blink's actual luminance rule), and the recent-fix records why the raw-colour sample was misleading. The tests in `OutlineStylePaintIntegrationTests.cs` were reverted to expect a genuinely lightened lit face.
- **Bevel shade fix stands.** `OutlineRegionPainter.Styled.cs` is whitespace-only since the head I verified at ~3,000 sweep cells against Edge 153 in both PDFium and MuPDF (wrong-shade cells 147 -> 72, all residual differences tiny mitre triangles; 0 unpainted / 0 overspill). With the lit-colour commit gone, the output is the 01b282fe behaviour that I measured.
- **Tests tightened.** The coverage test now checks each pass's band against its own clips, and there is a structural guard that every band fill holds exactly one contour (two subpaths), which closes the gap I flagged for the separate-pieces case. The new 0.35pt tolerance is justified in the comment (chord deviation of 0.26pt where two mitre territories meet across an arc, versus the wedge-sized failures it guards against) and is consistent with the 0-unpainted-pixel rasterization results.
- **Docs and notes.** The bogus `bevel-shade-overreach` migration note (it described a bug that never reached main) is deleted; the rectilinear-region invariant is rewritten for the current design; `docs/html-css-support.md` now limits the "dash lands on every corner" claim to square corners and still correctly says inset/outset lighten the lit face.
- **Citations.** `FragmentPainter.cs:731` and the test comments are now §3.1 / §3.2; the only remaining §-cite in the outline files is css-backgrounds-3 §4.5, which is right.
- **Accepted gap** for translucent dotted/dashed corners references #1222 (real, open, matches).

Own run: `dotnet build PeachPDF.slnx -t:Rebuild` 0 warnings / 0 errors; net8.0 12605 passed, 0 failed, 9 skipped; CI green on all three platforms.

Issue #1226, opened 2026-09-20
<hr> UA border colors are hardcoded per-side greys; Chrome uses color: gray + currentColor


ChrisVanDijk:
`<hr>`'s UA border colors are hardcoded per-side greys where Chrome uses `color: gray` + `currentColor`.

`CssDefaults` gives `<hr>` a fixed four-sided grey scheme:

```css
hr { border: 1px inset; }                                     /* line 92  */
hr { border-top-color: #9A9A9A; border-left-color: #9A9A9A;
     border-bottom-color: #EEEEEE; border-right-color: #EEEEEE; }  /* line 156 */
```

Blink's UA sheet (`third_party/blink/renderer/core/html/resources/html.css`) sets no border color at all — it sets `color`, and lets `border-color`'s initial `currentColor` pick it up:

```css
hr {
    display: block;
    overflow: hidden;
    margin-block-start: 0.5em;
    margin-block-end: 0.5em;
    margin-inline-start: auto;
    margin-inline-end: auto;
    border-style: inset;
    border-width: 1px;
    color: gray;
}
```

So in Chrome all four sides are declared `#808080` and the `inset` shading derives the two faces from it: `#2c2c2c` on the top/left, `#d4d4d4` on the bottom/right. PeachPDF declares two different greys up front and derives nothing.

## Impact

| | Chrome | PeachPDF |
| --- | --- | --- |
| declared, all sides | `#808080` (`currentColor`) | `#9A9A9A` top/left, `#EEEEEE` bottom/right |
| painted top/left | `#2c2c2c` | `#9a9a9a` |
| painted bottom/right | `#d4d4d4` | `#eeeeee` |

Two consequences beyond the colors themselves:

- **`color` has no effect on an `<hr>`.** In Chrome, `hr { color: red }` (or inheriting a non-default `color` where the author overrides the UA rule) recolors the rule, because the border is `currentColor`. In PeachPDF the hardcoded per-side declarations always win over `currentColor`, so `color` is inert on a rule.
- **The rule is lighter and lower-contrast than a browser's**, since `#9A9A9A`/`#EEEEEE` never darken to anything like `#2c2c2c`.

## Repro

```html
<hr>
<hr style="color: red">
```

Chrome: a grey engraved rule, then a red one. PeachPDF: two identical `#9a9a9a`/`#eeeeee` rules.

## Relationship to the paint-path issue

[...]

Issue #1225, opened 2026-09-20
<hr> ignores border-style: the legacy hr painter discards every bevelled, patterned and double rule


ChrisVanDijk:
`HrFragmentPainter` paints an `<hr>` through a legacy per-side path that ignores `border-style`, so every bevelled, patterned and double rule renders solid.

`HrFragmentPainter.Paint` (`src/PeachPDF/Html/Core/Paint/Content/HrFragmentPainter.cs:27-40`) draws each side itself, handing `BordersDrawHandler.DrawBorder` a solid brush built from the raw declared color:

```csharp
var b1 = g.GetSolidBrush(box.ActualBorderTopColor);
BordersDrawHandler.DrawBorder(Border.Top, g, box, b1, rect);
```

`DrawBorder` is explicitly a legacy shim — its own doc comment says so:

> Draws one plain side polygon for the legacy horizontal-rule painter, which supplies its brush separately. Normal CSS box borders use `DrawBoxBorders`.

It emits one flat polygon per side and never reads `BorderTopStyle`/`BorderLeftStyle`/etc. Every `border-style` implementation — the bevel shading in `BorderBevelColors`, the dot/dash fitting, `double`'s thirds — lives behind `BordersDrawHandler.DrawBoxBorders` → `BoxEdgesDrawHandler.DrawBoxEdges`, which an `<hr>` never reaches.

So `border-style` is silently discarded on `<hr>` and only on `<hr>`.

## Impact

`<hr style="border:2px inset #808080">` renders as a flat `#808080` slab. Chrome renders `#2c2c2c` on the top/left and `#d4d4d4` on the bottom/right. Same for `outset`, `groove`, `ridge`, `dotted`, `dashed` and `double`.

This also makes the UA default wrong, since Blink's own `<hr>` is `border-style: inset` (see the companion issue on the UA colors).

## Repro

```html
<div style="height:0; border:2px inset #808080"></div>   <!-- #2c2c2c / #d4d4d4 -->
<hr  style="border:2px inset #808080">                    <!-- #808080 flat    -->
<hr  style="border:2px groove #808080">                   <!-- #808080 flat    -->
<div style="height:0; border:2px groove #808080"></div>  <!-- 4 correct bands -->
```

Rasterized at 8× through PDFium, sampling a vertical line through each element:

| element | painted |
| --- | --- |
| `div{height:0}` `2px inset` | `#2c2c2c` / `#d4d4d4` |
[...]

PR #1227, opened 2026-09-20
Match Blink's bevel shading rule at the near-white and near-black ends

### LINKED ISSUE #1224: Bevel lit face lightens near-white colors where Chrome keeps the declared color
`BorderBevelColors.Shade` always lightens a bevel's lit face (`Light(color)`, `src/PeachPDF/Html/Core/Utils/BorderBevelColors.cs:78`). Current Chromium does not: since M149 (updated in M151, feature `TableDefaultBorderColorCurrentColor`, now stable) Blink's `CalculateInsetOutsetColor` in `box_border_painter.cc` keeps the **declared color** on the lit face when its relative luminance is above **0.83077**, and lightens only below that. Very dark colors still lighten both faces, which PeachPDF already matches.

The dark face is identical under either rule; only the lit one diverges, and only at the near-white end.

## Impact

| color | Chrome's lit face | PeachPDF |
| --- | --- | --- |
| `#4a90d9` | `#57a9ff` | `#57a9ff` ✔ |
| `#808080` | `#d4d4d4` | `#d4d4d4` ✔ |
| `rgba(74,144,217,.5)` over white | `#aad3ff` | `#aad3ff` ✔ |
| `#f0f0f0` | `#f0f0f0` | `#ffffff` ✘ |

Measured against Edge 153 over 48 random colors plus 12 greys, on `inset`/`outset`/`groove`/`ridge`, borders and outlines alike: PeachPDF's lit face matches exactly in 58 of 60 cases, the two misses being near-white. Implementing Blink's threshold should take that to 60/60.

## Repro

```html
<div style="width:200px;height:60px;border:12px outset #f0f0f0;background:#888"></div>
```

Chrome draws the top/left faces at `#f0f0f0`; PeachPDF clips them to `#ffffff`, so the bevel loses its lit/declared distinction against a white-ish backdrop.

## Why it is filed rather than fixed in place

CSS 2.1 §8.5.3 leaves the exact shading UA-defined, so this is not a spec violation — but this engine deliberately reproduces Chrome's rendering here, down to Blink's integer truncation and its 255.99998 scale factor, so the divergence is a defect by the standard the rest of the file is held to.

It is not a fix to fold into an unrelated change: `BorderBevelColors` is shared by every bevelled border, every bevelled outline and every collapsed-table segment, so changing `Shade` alters rendered output library-wide, and the current lightening shipped in v0.9.19.

## Suggested fix

In `Shade`, return the declared color instead of `Light(color)` when its relative luminance is above 0.83077, keeping `Light(color)` below it and the existing near-black both-faces-lighten fallback untouched. Note that `docs/html-css-support.md` documents the current behavior ("`inset` darkens the top and left and lightens the bottom and right") and has to change with it, along with the pinned expectations in `BorderStylePaintIntegrationTests.cs` and `OutlineStylePaintIntegrationTests.cs`.

ChrisVanDijk:
Fixes #1224.

`BorderBevelColors.Shade` now ports Blink's `CalculateInsetOutsetColor` directly instead of approximating it:

```csharp
var luminance = RelativeLuminance(color);

if (luminance <= NearBlackLuminance)          // 0.014443844 == lum(rgb(32,32,32))
{
    var light = Light(color);
    return darken ? light : Light(light);
}

if (darken)
    return Dark(color);

return luminance > NearWhiteLuminance ? color : Light(color);   // 0.83077 == lum(rgb(235,235,235))
```

## Scope is wider than the issue asked for

#1224 scoped this to the near-white lit face and explicitly said to leave the near-black fallback alone. Reading Blink's actual source first showed the two ends are one rule, and that the near-black end was also wrong.

PeachPDF's fallback was `ContrastRatio(color, Dark(color)) < 1.3`, derived by sampling Chrome at the gray boundary. That reproduces the boundary *on grays* exactly, which is why it survived, but it is not the rule Blink uses. The 1.75 contrast ratio that does appear in `box_border_painter.cc` belongs to the **pre-M149** branch and decides the *lit* face, not the dark one.

An exhaustive sweep of all 16,777,216 sRGB colors puts the difference at 223,626 of them:

- **200,298** at the near-white end — the miss #1224 filed. Lit face only.
- **23,328** very dark chromatic colors where the contrast-ratio test lightened both faces and Chrome darkens. One-directional: there is no color where Blink lightens and the old rule darkened. `#001e4c` is representative — luminance 0.014503, just above the threshold, but contrast ratio 1.290 against its own darkened form, just below 1.3.

That second bucket is invisible to grays and to uniformly-random sampling, which is why #1224's 60-sample measurement found only the near-white misses.

`BorderBevelColors` is shared by every bevelled border, outline and collapsed-table segment, and the previous behaviour shipped in v0.9.19, so this is a user-visible change — recorded in `.claude/migration-notes/` with a before/after table for the next release notes pass.

## Threshold boundaries

Both constants are knife-edges, pinned by tests:

| gray | luminance | vs constant | face |
| --- | --- | --- | --- |
| 32 | 0.014443843596 | `<= 0.014443844` (by 4e-10) | both lighten |
| 33 | 0.015208514423 | above | darkens |
| 235 | 0.830769876775 | **not** `> 0.83077` (by 1.2e-7) | still lightens |
| 236 | 0.838799011741 | `> 0.83077` | declared color |

Gray 235 clears the constant by 1.4e-7, so the `double` literals here and the `float` constants Blink uses agree on it only because the comparison is strict and the constant *is* lum(235). Recorded as an invariant so a later tidy-up does not shorten either literal or relax `>` to `>=`.

## Verification

- Full suite on net8.0: 12,608 passed, 9 skipped. CLI tests 96, source-generator tests 119.
- Diff coverage on `BorderBevelColors.cs`: every changed coverable line hit, none missed.
- `border_style` showcase grew a "colors too light to lighten" row, rasterized through **both** PDFium and MuPDF per the repo's paint-verification convention. Both agree, and sampling the PDFium raster gives lit `rgb(240,240,240)` / dark `rgb(156,156,156)` — Chrome's exact bytes for `border: 16px outset #f0f0f0`.
- `docs/html-css-support.md`'s bevel prose now covers both ends of the range.

Most of the existing bevel tests assert a painted face against `BorderBevelColors.Shade` on both sides of the comparison, so they cannot fail when the rule changes — the whole suite passed on the first run of the new code. The tests that actually guard this are the ones pinning literal `RColor`s, and those are what the new coverage adds.

## Note on CI

`CssLayoutEngineTableTests.TableLayout_AsymmetricWrappableHeaders_InterpolateBetweenColumnMinAndMax_Issue1157` fails on `main` at the same commit this branches from, verified by stashing and re-running it alone. It is unrelated to this change and not addressed here.

## Out of scope, filed separately

Review of this change flagged the UA default `hr { border: 1px inset }` + `border-bottom-color: #EEEEEE` as the most commonly hit instance of the new near-white rule. It is not hit at all — `<hr>` has its own content painter that draws each side with a solid brush of the raw declared color and never reads `border-style`, so a rule never reaches `Shade`. Rendering the same fixture before and after this change gives byte-identical `<hr>` output. Filed as #1225 (the paint path) and #1226 (the UA colors), and recorded as a trap in the fix note.


jhaygood86 (approved):
Independently verified end to end. Approving.

**The port is Blink's rule.** I read `CalculateInsetOutsetColor` in current Blink (`TableDefaultBorderColorCurrentColor`, stable, shipped M149, updated M151): constants `0.014443844f` / `0.83077f`, near-black uses `<=`, near-white uses `>`, luminance is taken on the declared colour's RGB (alpha ignored), near-black lightens the dark face once and the lit face twice. The 1.75 contrast ratio is only in the pre-M149 branch, as the PR says.

**Exhaustive Chrome ground truth (Chrome 153).** Rendering `border: 6px outset` for all 16,777,216 sRGB colours and comparing exact dark/lit pairs: old rule 98.55% match, this PR 99.72%. In the two changed buckets (200,298 near-white, 23,328 near-black chromatic at the rule level) the PR mismatches Chrome on 0 colours, versus 173,376 and 23,328 for the old rule. Nothing is right on main and wrong here. Gray rows 31/32/33 and 234-237 match Chrome pixel for pixel, and there are 0 colours where double-vs-float luminance falls on different sides of either threshold.

**Real PeachPDF output.** About 309 colours across outset/inset/groove/ridge, borders and outlines, rendered through PeachPDF and rasterized: PDFium 304/304 exact for the PR (old rule 187/304; the 5 translucent cases within +-1); MuPDF within +-1 (it truncates where PDFium rounds, a renderer artefact). Showcase row samples lit `rgb(240,240,240)` / dark `rgb(156,156,156)` for `outset #f0f0f0`, identical to Chrome.

**Regression check against the merge base.** 43 colours x border, outline, rounded, per-side, collapsed-table, wrapped-inline-outline and wrapped-inline-border: content streams byte-identical except for exactly the colours the rule predicts change (#f0f0f0, #eee, #ececec, #ededed, #f8f0e0, #f7fafc, #001e4c, #001e4d). `<hr>` is byte-identical, and the claim about its own painter never reading `border-style` is accurate (#1225 and #1226 exist).

**Own run:** `dotnet build PeachPDF.slnx -t:Rebuild` 0 warnings / 0 errors; net8.0 12609 passed, 0 failed, 9 skipped; diff coverage on `BorderBevelColors.cs` 100%; CI green on all three platforms. The change is genuine against v0.9.19 (the contrast-ratio fit is in that tag) and the migration-note before/after values match real renders.

**Non-blocking follow-ups (none block merge):**
1. `Light()` rounds +-1 differently from Chrome on ~47,138 colours (0.28%), lit face, one channel; identical on main, so not introduced here (e.g. `rgb(0,9,192)` outset lit: Chrome `(0,12,255)`, PeachPDF `(0,11,255)`). A float32 port of `Dark`/`Light` reproduces Chrome on all 16.7M colours. Worth an issue; it also means "reproduce it exactly" / "matches Chrome exactly" is slightly overclaimed.
2. The PR body says `TableLayout_AsymmetricWrappableHeaders_..._Issue1157` fails on main; it passes for me in full runs on both main and this branch, so it looks flaky/order-dependent rather than consistently failing.
3. Mutation testing: the two boundary tests catch the constant changes I tried within ~1e-4 of each threshold but not a near-black constant in (0.014443844, 0.0145] or near-white in (0.8307699, 0.8388]. Pinning a boundary chromatic (Chrome: `rgb(189,255,90)` -> `(127,171,60 | 189,255,90)`, `rgb(255,235,156)` -> `(171,158,104 | 255,235,156)`) and a groove/ridge or collapsed-table near-white case would close that. (`<=`/`<` and `>`/`>=` mutants are equivalent for 8-bit colours, so the invariant's emphasis on them is slightly overstated.)
4. Small note wording: "223,626 colours differ" is a rule-level count (196,704 actually change output, since 26,922 near-white colours have a channel at 255); the gray-235 margin is 1.2e-7 in the table but 1.4e-7 in prose; the migration note says "stable since M151" where the source says shipped M149, updated M151.

PR #1228, opened 2026-09-20
Format a computed CSS fixture width with InvariantCulture

ChrisVanDijk:
`TableLayout_AsymmetricWrappableHeaders_InterpolateBetweenColumnMinAndMax_Issue1157` fails on a developer machine under any comma-decimal locale, while staying green on CI.

## Cause

The test computes `desiredWidth` at runtime and interpolates it straight into its fixture:

```csharp
table {{ ...; width: {desiredWidth}pt }}
```

That formats with `CurrentCulture`. Under nl-NL / de-DE / fr-FR the declaration reaches the parser as `width: 213,98193359375pt`, which is not valid CSS. Nothing throws — the declaration is dropped, the table falls back to `width: auto`, and it lays out at exactly its max-content width:

```
minSum=176.80, maxSum=251.15, desiredWidth=213.98, tableWidth=251.15
Table width 251.1572265625 overflowed its specified 213.98193359375pt width
```

`tableWidth == maxSum` to the bit. The assertion reports an overflow, but the table was never given a width at all — the failure reads like the layout bug the test guards, which is what makes it expensive to diagnose.

CI agents run invariant culture, so the declaration parses there and the test passes.

## Fix

Format with `InvariantCulture` before interpolating. Rejecting a comma decimal is spec-correct — CSS numbers are culture-invariant — so this belongs in the test, not the parser. No library change.

`OutlineStylePaintIntegrationTests` already carries an inline comment about this exact trap, where the same silent fallback left `outline-width`/`outline-offset` at `medium`/`0` and would have compared `outline-style: auto` against the wrong ring and passed **green** for the wrong reason. Second sighting, so it is recorded under `.claude/invariants/` — the point being that the "green on CI, red locally" signature reads like a font or platform difference and sends you into the layout engine instead of the fixture.

## Scope

Only *computed* values are at risk; literals and integral `[InlineData]` doubles format identically in every culture, which is why the rest of the suite's interpolated fixtures are fine and why a grep for interpolations doesn't isolate this.

## Verification

Full suite run under nl-NL (`net8.0`): **12609 passed, 0 failed, 9 skipped** — confirming this was the only remaining culture-dependent fixture, rather than just the one that happened to be reported.


jhaygood86 (approved):
Verified the claim directly. I forced nl-NL via a throwaway module initializer (not committed) and ran the test:

- origin/main under nl-NL: `TableLayout_AsymmetricWrappableHeaders_InterpolateBetweenColumnMinAndMax_Issue1157` fails with "Table width 251,1572265625 overflowed its specified 213,98193359375pt width", exactly as described.
- This branch under nl-NL: the same test passes.
- Full suite under nl-NL on this branch: 12609 passed, 0 failed, 9 skipped, which supports the claim that this was the only remaining culture-dependent fixture.

The fix is test-only (no library change), which is right: CSS numbers are culture-invariant, so rejecting a comma decimal in the parser is correct. The invariant note is accurate.

CI note: the first ubuntu/net10.0 run failed on `XGraphicsTests.AppendPdfNumber_UsesAllocationFreeFastPathAndPreservesFallbackFormatting` ("allocated 1,424 bytes", cap 1,024). That test is unrelated to this change and passed on re-run of the same commit, so it looks like a flaky allocation-count assertion under CI load/coverage (worth a separate look, since a `GC.GetAllocatedBytesForCurrentThread` cap that tight can be tripped by tiered-JIT or instrumentation allocations).

Issue #1230, opened 2026-09-20
<hr> resolves a percentage width against a basis reduced by its own border widths


ChrisVanDijk:
`<hr>` resolves a percentage `width` against a basis reduced by its own border widths.

[CSS 2.1 §10.2](https://www.w3.org/TR/CSS21/visudet.html#the-width-property) resolves a percentage
`width` against the containing block's **content** width, with the element's own padding and border
added outside it under the default `box-sizing: content-box`. `CssBoxHr.PerformLayoutImp`
(`src/PeachPDF/Html/Core/Dom/CssBoxHr.cs:93-103`) instead builds the percentage basis as the
containing block's content width *minus the rule's own left and right border widths*, and passes that
to `CssValueParser.ParseLength`:

```csharp
double width = ContainingBlock.Size.Width
               - ContainingBlock.ActualPaddingLeft - ContainingBlock.ActualPaddingRight
               - ContainingBlock.ActualBorderLeftWidth - ContainingBlock.ActualBorderRightWidth
               - ActualMarginLeft - ActualMarginRight - ActualBorderLeftWidth - ActualBorderRightWidth;

if (Width != Keywords.Auto && !string.IsNullOrEmpty(Width))
{
    width = CssValueParser.ParseLength(Width, width, this);
}
```

The last two subtrahends belong to the `auto` branch (where the rule's content width genuinely is the
available space less its own borders) and not to the percentage basis.

## Repro

```html
<div style="width: 200pt">
  <hr  style="margin:0; width:50%; border:4px solid #000">
  <div style="margin:0; height:0; width:50%; border:4px solid #000"></div>
</div>
```

Measured after layout (1px = 0.75pt, so a 4px border is 3pt):

| element | content width | border-box width |
| --- | --- | --- |
| `div` `width:50%` | 100 ✔ | 106 ✔ |
| `hr` `width:50%` | **97** | **103** |
| `div` `width:100%` | 200 ✔ | 206 ✔ (overflows, as a browser does) |
| `hr` `width:100%` | **194** | **200** |

A browser agrees with the `div` row in every case.

## Impact

Any `<hr>` with a percentage `width` and a border wider than 0 is narrower than it should be, by twice
its border width times the percentage. Most visible at `width: 100%`, where the rule is exactly its
own border widths short of the box a `<div>` with the identical declaration produces — which is how
this surfaced: pairing an `<hr>` against its equivalent zero-height `<div>` in a showcase.

[...]

Issue #1229, opened 2026-09-20
<hr> advances the block flow by a constant 2 units, so the next block overlaps any rule taller than that


ChrisVanDijk:
`<hr>` advances the block flow by a constant 2 units regardless of its own height, so the next block overlaps it.

A block-level box's following in-flow sibling starts at that box's border-box bottom
([CSS 2.1 §9.4.1](https://www.w3.org/TR/CSS21/visuren.html#normal-flow), normal flow: "boxes are laid
out one after the other vertically, beginning at the top of a containing block"). An `<hr>` does not
participate: whatever its resolved height, the next sibling is placed exactly 2 units below the
rule's *top*, so any rule taller than 2 units is overlapped by whatever follows it.

## Repro

```html
<hr  id='h' style='margin:0; border:6px solid #808080'>
<div id='d' style='margin:0; height:0; border:1px solid #000'></div>
```

Measured after layout (`LayoutHarness.LayoutAsync`, 1px = 0.75pt, so a 6px border is 4.5pt):

| first element | its `Location.Y` | its `ActualBottom` | following div's `Location.Y` |
| --- | --- | --- | --- |
| `hr` `border:1px` | 26 | 28 | 28 |
| `hr` `border:2px` | 26 | 29 | **28** |
| `hr` `border:3px` | 26 | 30.5 | **28** |
| `hr` `border:6px` | 26 | 35 | **28** |
| `hr` `border:10px` | 26 | 41 | **28** |
| `hr` `border:6px; height:10px` | 26 | 42.5 | **28** |
| `div{height:0}` `border:6px` | 26 | 35 | 35 ✔ |

The last row is the control: the equivalent zero-height `<div>` advances the flow correctly, so this
is specific to `CssBoxHr`. The rule's own `ActualBottom`/`StaticBottom` are correct, and its fragment's
`WholeBoxRect` is the right height too (`{X=26,Y=26,Width=543,Height=9}` for the 6px case) — it is only
the *following* sibling's placement that ignores them.

The constant 2 is the `height = 2` fallback in `CssBoxHr.PerformLayoutImp`
(`src/PeachPDF/Html/Core/Dom/CssBoxHr.cs`), so the following sibling appears to be placed against an
`ActualBottom` written before the rule's real height was resolved, and never re-placed once it was.

## Impact

Any `<hr>` with a border thicker than about 1px is painted over by the content after it. It is most
[...]

Issue #333, opened 2026-07-25
css-break-3: OffsetFlowedWords relocates atomic-inline words outside the break-token model


jhaygood86:
### Spec

CSS Fragmentation Module Level 3 §4.1 (a line box is a monolithic break unit): https://www.w3.org/TR/css-break-3/

### Gap

#321 moved inline fragmentation onto the break-token model: `CssLayoutEngine.FlowBox` now asks, before placing a word, whether the line would straddle a fragmentainer boundary, and if so stops and records where to resume — so the whole line moves rather than individual words relocating behind layout's back.

One call site was left on the old per-word mechanism. `CssLayoutEngine.OffsetFlowedWords` (called from `ApplyAtomicInlineVerticalInsets`) shifts already-flowed words down by an atomic inline's border+padding-top inset, and then re-runs `CssRect.BreakPage()` on each shifted word — because the shift can push a line that legitimately fitted back across a boundary. That relocation happens *after* the line's break decision has been made, and it moves words individually, so it can still split a line across a boundary in exactly the way the rest of the inline path no longer does.

It was left alone deliberately: it operates within an already-decided line, and folding it in means the inset has to be known before the line's break decision rather than applied to the line afterwards.

`CssRect.BreakPage` exists only for this call site plus monolithic content, where breaking is suppressed anyway. Closing this gap is what would let it be deleted.

### Work

Apply the atomic-inline vertical inset before the line's break decision, so the line's real height is known when the decision is taken, and drop the post-hoc `BreakPage` re-check. Then remove `CssRect.BreakPage` and `CssBox.BreakPage` (the latter already has no production call site).

### Key files

`Html/Core/Dom/CssLayoutEngine.cs` (`OffsetFlowedWords`, `ApplyAtomicInlineVerticalInsets`, `FlowBox`), `Html/Core/Dom/CssRect.cs` (`BreakPage`, `WouldStraddleFragmentainer`).

### Dependencies

Builds on #321.

jhaygood86:
**Adjacent finding from #331 (PR #364), worth knowing before this is picked up.**

[...]

PR #1231, opened 2026-09-20
Paint an <hr> through the ordinary border path

### LINKED ISSUE #1225: <hr> ignores border-style: the legacy hr painter discards every bevelled, patterned and double rule

ChrisVanDijk:
Fixes #1225

`HrFragmentPainter` drew each of the rule's four sides itself, as one flat polygon filled with the raw declared color, through a `BordersDrawHandler.DrawBorder` shim whose own doc comment already called itself legacy. It read `BorderTopStyle`/`BorderLeftStyle`/etc. nowhere, so every bevelled, patterned and `double` rule painted as a solid slab — and only on `<hr>`, since nothing else reached that shim.

An `<hr>` is an ordinary block box whose rule *is* its border, so the generic `FragmentPainter.PaintBoxContent` → `BordersDrawHandler.DrawBoxBorders` path was already the right one. The fix is a deletion: drop the painter, its arm in `FragmentContentPainters.For`, and `DrawBorder`/`GetBandPoints`.

| `<hr style="border: 2px inset #808080">` | before | after | Chrome |
| --- | --- | --- | --- |
| top/left face | `#808080` | `#2c2c2c` | `#2c2c2c` |
| bottom/right face | `#808080` | `#d4d4d4` | `#d4d4d4` |

### Layout was also rewriting computed style

`CssBoxHr.PerformLayoutImp` overwrote `BorderTopStyle`/`BorderBottomStyle` with `solid` and both widths with `1px` whenever the resolved height was two units or less and both horizontal borders were under one unit. The UA default's `1px` is 0.75pt, so this fired on *every* default rule: after layout a plain `<hr>` reported `styleT=Solid styleR=Inset styleB=Solid styleL=Inset`. Fixing the painter alone would have left the default rule still painting `solid` on two of its four sides.

It also silently discarded an author's `border-style` on any thin rule, and turned `border: none` into two visible 1px lines. Both are gone; the height fallback it was tangled with stays.

The old painter's `rect.Height > 1` / `> 2` guards decided whether the left/right/bottom sides and the background painted at all. They are not load-bearing: a browser paints all four sides of a 2px-tall rule, mitred at the ends, which is what `DrawBoxEdges` does with the rule's real `WholeBoxRect`.

### Two consequences the fix makes load-bearing, so they come with it

**`hr[color], hr[noshade] { border-style: solid }`.** The HTML Standard pairs that rule with the `hr { border-style: inset }` the UA sheet already had. While the rule painted flat regardless, `noshade` was right by accident; the moment `border-style` started working, `<hr noshade>` began rendering engraved — the exact thing the attribute exists to turn off.

**Valueless attributes were stored as null, so `[attr]` never matched them.** `HtmlParser.ParseHtmlTag` did `x.First().Value!`; the tokenizer reports a valueless attribute with a null value, and the HTML Standard gives it the empty string. So `hr[noshade]` matched nothing while `hr[noshade=""]` matched — and the same held for `[disabled]`, `[hidden]`, `[required]` and every other boolean attribute in the language. One `?? string.Empty`. Worth a reviewer's attention for blast radius: this changes what `TryGetAttribute(name, default)` returns for *any* valueless attribute, so every caller that branches on "absent" sees one more attribute as present. The whole suite was run against it specifically; no test moved.

### Testing

The regression here is silent — the old path painted something plausible for every style, nothing threw, and no content-stream token went missing, so a substring or polygon-count check would have passed throughout. `HrBorderStylePaintIntegrationTests` therefore compares a rule against the zero-height `<div>` carrying the identical border, over the whole ordered `TestRecordingGraphics.Log` normalized to each box's own border-box origin, for all seven non-`solid` styles. Plus one test pinning Chrome's literal `#2c2c2c`/`#d4d4d4`, so a future change to `BorderBevelColors` cannot quietly drift the rule with it.

- Full suite green on net8.0 (12629 passed). The one failure, `CssLayoutEngineTableTests.TableLayout_AsymmetricWrappableHeaders_InterpolateBetweenColumnMinAndMax_Issue1157`, fails identically on a clean `main` and is unrelated.
- Whole solution rebuilds with zero warnings.
- The library gains one executable line and one UA stylesheet rule; everything else added is a comment and everything else changed is a deletion (108 lines out).
- The `border_style` showcase gained an `<hr>`-vs-`<div>` section, verified by diffing each rule's raster strip against its div's in **both** PDFium and MuPDF: mean channel difference 1.07–2.74 across both rows in both engines, i.e. antialiasing.

### Found while verifying, filed not fixed

Two pre-existing `CssBoxHr` layout defects, both measured, both unchanged by this diff (before and after place content identically):

- #1229 — an `<hr>` advances the block flow by a constant 2 units regardless of its resolved height, so anything after a rule with a border thicker than about 1px lands on top of it. An equivalent zero-height `<div>` advances correctly.
- #1230 — a percentage `width` on a rule resolves against a basis reduced by the rule's own left and right border widths, so `width: 50%` in a 200pt block gives 97pt of content where CSS 2.1 §10.2 and the equivalent `<div>` say 100pt.

Both are described reader-facing in the `hr` row of `docs/html-css-support.md`, with the issue numbers recorded in `.claude/accepted-gaps/`.

#1226 (the hardcoded per-side UA greys) is deliberately not touched — that issue notes the paint path can land first, and this is it.


jhaygood86 (changes requested):
Thorough review done. The approach is right and most of it verifies well, but one regression blocks merge: **a default `<hr>` (no author CSS) now paints a darker top/left edge than both `main` and Chrome.** That is the most common `<hr>` usage on the web.

## Blocking
**1. Default `<hr>` top/left colour is doubly darkened.**
Chrome 153 paints a plain 1px default rule as `#9a9a9a` (top/left) and `#eeeeee` (bottom/right). `main` painted exactly that. This PR paints `#464646` / `#eeeeee`. Cause: `CssDefaults.cs:162` hard-codes `hr { border-top-color: #9A9A9A; border-left-color: #9A9A9A; border-bottom-color: #EEEEEE; border-right-color: #EEEEEE }`. Those values were already Chrome's *post-shading* output, painted flat under the old painter. Now that `hr { border: 1px inset }` (line 92) actually bevels, `#9A9A9A` is shaded a second time. Your migration note records the change (`top #9a9a9a -> #464646`) but frames it as "higher contrast", not as a divergence from Chrome. Same doubled darkening shows up for `border-top:3px double #333` (left edge `#464646` vs Chrome `#9a9a9a`) and `border-bottom-color`, and everywhere the default rule appears (`size=`, `height:0`, in a table cell, flex container, multicol, blockquote, list item).

The deferral to #1226 rests on a premise that doesn't hold: it says a default `<hr>` should land on `#2c2c2c` / `#d4d4d4`. In Chrome 153, an unstyled `<hr>`, `border:1px inset gray`, `border:2px inset currentColor` and `<div style="border:2px inset;color:red">` all paint `#9a9a9a` / `#eeeeee`; only an *explicit* `#808080` gives `#2c2c2c` / `#d4d4d4`. Chrome treats an inset/outset border whose colour is `currentColor` as if its base were about `#eee`. Since this PR is what makes the bevel apply to `<hr>`, the UA colours have to change here, not in a follow-up. The review agent measured that `hr { border-color: #eee }` reproduces Chrome's `#9a9a9a` / `#eeeeee` exactly (one trade-off: `border-style:dashed` on an unstyled hr would then paint `#eee` where Chrome paints gray, which is already wrong today), so that is one option. Please also add a test that pins the default rule's painted colours (`#9a9a9a` / `#eeeeee`); none of the current tests would catch this, which is why the suite stayed green.

**2. Wrong HTML Standard section number** (a recurring issue in this area, so I checked the source). The `hr` rendering section is **15.3.11 "The hr element"**; 15.3.6 is "Sections and headings". "HTML Standard 15.3.6" appears in `CssDefaults.cs:93`, in `HrBorderStylePaintIntegrationTests.cs:114` and in the recent-fix note.

## What checks out (verified independently)
- Every explicit style on `<hr>` (none/hidden/solid/dotted/dashed/double/groove/ridge/inset/outset) now agrees with Chrome and with the zero-height `<div>` oracle across ~300 documents, PDFium and MuPDF agreeing. `none`/`hidden` paint nothing, as in Chrome; `border-radius` and `background` on an `<hr>` are now honoured.
- The very common idioms `border:0; border-top:1px solid #ccc` and `border:0; height:1px; background:#ccc` are fixed, not regressed (on `main` both drew phantom black lines from the old style rewrite; `height:10px` rules were 1.5pt too tall and now match Chrome).
- Layout: the `CssBoxHr` style/width overwrite is removed as described and the height fallback stays. Location, flow advance (#1229), width (#1230), pagination and tagged-PDF wrapping are unchanged vs `main`; the old `> 1` / `> 2` guards were not hiding any overlap. #1229 and #1230 exist and match their descriptions.
- Valueless-attribute change: single attribute parser (SVG/MathML/MHTML all go through it). Across 74 element/attribute combinations x (valueless / `=""` / `="name"` / absent), plain and with interactive forms + tagged PDF, no previously-working boolean attribute flipped behaviour. It also fixes crashes on `main` for valueless `<a href>`, `<td bgcolor|width|align|valign|background>`, `<font color>`, `<table cellspacing>`, ``, `<input type>`, `<p style>` and several MathML attributes.
- `hr[color], hr[noshade] { border-style: solid }` is verbatim in the HTML Standard and matches Chrome.
- Mutation tests all bite: restoring the flat painter fails 11 tests; re-adding the layout overwrite fails 3; reverting `?? string.Empty` fails 6; removing the `hr[color], hr[noshade]` rule fails 3. The `<hr>`-vs-`<div>` oracle is not tautological.
- Own numbers: `dotnet build PeachPDF.slnx -t:Rebuild` 0 warnings/0 errors; net8.0 12630 passed, 0 failed, 9 skipped; diff coverage 100%; CI green on all three platforms. No AI attribution in the commit or PR body.

## Non-blocking (please fold in if convenient)
- The recent-fix says a valueless attribute previously returned "the caller's own default". It returned `null` (key present, null value, so `TryGetValue` succeeded), which is precisely what made `.ToLower()`/`int.Parse` throw.
- The migration note omits: the phantom-line fix for `border:0` rules, `height:Npx` rules getting 1.5pt shorter, the crash fixes above, and two spec-correct semantic changes (`<option value>` valueless now exports `""` rather than the label in interactive forms; `<mfenced open>` now yields an empty open fence).
- `docs/architecture.md` is stale: line 562 still lists `<hr>` among boxes drawn by an `IFragmentContentPainter`, and line 364 still says the rule "falls back to a 1px top and bottom border".
- #1229 also collapses the *parent container's* height when an `<hr>` is the last child (Chrome gives 7.5/12/15 for 1/4/10px rules), not only the following block.
- #1229's cited basis (CSS 2.1 §9.4.1) only says boxes lay out "one after the other, vertically"; §10.6.3 or §8.3.1 with §10.5 would be a tighter citation.
- 1px `groove`/`ridge` and 1-2px `double` paint two-tone on an `<hr>` where Chrome paints solid; that's the shared border code, not specific to hr.
- Pre-existing and unchanged (identical on `main`), so not for this PR: `<hr color=red>` still paints gray, `<hr noshade>` paints two-tone where Chrome paints flat `#808080`, `<hr size=5>` is 7px tall vs Chrome's 5px, the top margin differs (8.25pt vs 3.75pt), an hr as a flex item wraps, and `page-break-before/after` on an hr are ignored.

Happy to re-review as soon as the default-colour fix and the citation land.

ChrisVanDijk:
Thanks — the blocking finding was right, and it was the one thing here that only looking could catch. Fixed in 9b8ed32c.

## 1. Default `<hr>` double darkening — fixed

The four per-side UA declarations were Chrome's *already-shaded output*, baked in as literals, so a rule that bevels for real shaded them a second time. The single base whose two faces **are** those greys is `#eee`:

```
Shade(#eee, darken) == #9a9a9a        // lit face is above the near-white
Shade(#eee, lit)    == #eeeeee        // threshold, so it keeps the declared colour
```

So `CssDefaults` now carries one declaration, `hr { border: 1px inset #eee }`, and the per-side rule is gone. Rendered against a **`main` worktree** at 8×, sampling each rule:

| | main | this PR | Chrome 153 |
| --- | --- | --- | --- |
| unstyled `<hr>` | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` |
| `<hr size=3>` | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` | — |
| `<hr>` in a table cell | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` | — |
| `border-top:3px double #333`, left edge | `#9a9a9a` | `#9a9a9a` | `#9a9a9a` |
| `<hr noshade>` | `#9a9a9a` / `#eeeeee` | `#808080` flat | `#808080` flat |

Byte-identical to `main` for the default rule everywhere it appears, and `<hr noshade>` now matches Chrome as well (the `hr[color], hr[noshade]` rule additionally carries `border-color: gray` — a flat rule derives nothing from a bevel, so the spec's own `color: gray` stands).

A note on method, since it nearly cost me the diagnosis: `git stash -- src/PeachPDF` does **not** restore a file whose deletion is already committed, so stashing to get a "main" baseline silently reproduces the PR instead. My first comparison came out self-inconsistent for exactly that reason. The numbers above are from a real worktree at `ceb4db99`.

**Two regression guards added**, and I checked both fail against the old sheet (the second one reproduces your `#464646` exactly):
- `DefaultHr_PaintsTheTwoGreys_ABrowserPaints` — the painted colours of a plain `<hr>`, which nothing else in the suite covers because every other test declares its own border.
- `DefaultHrBaseColor_ShadesToExactlyThoseGreys` — pins the `#eee → #9a9a9a/#eeeeee` relationship the sheet depends on, so a future `BorderBevelColors` change fails naming its cause rather than only in the paint test.

### Your point about #1226's premise

You're right, and it goes further than the deferral: the issue's *suggested fix* would have made things worse. I've [commented there](https://github.com/jhaygood86/PeachPDF/issues/1226#issuecomment-5750431229) correcting it. The real content is that Blink doesn't bevel a `currentColor` border from `currentColor` — it shades a fixed light base — and PeachPDF has no cascade signal for "this colour came from currentColor". That's a general defect (`<div style="border:2px inset">` is the plain case), not an `hr` one, and it's the only thing that fixes your `hr { border-style: dashed }` trade-off, which I've recorded rather than papered over. The UA rule carries a "do not restore this to `currentColor` or `gray` without implementing that first" comment next to it.

## 2. Section number — fixed

§15.3.11 "The hr element"; 15.3.6 is "Sections and headings". Corrected in `CssDefaults.cs` and the test. It was not in the recent-fix note (only those two places, plus my PR text above).

## Non-blocking

Folded in: the "caller's own default" wording (it returned **null** — key present, `TryGetValue` succeeded, `defaultValue` never consulted); `<option value>` and `<mfenced open>` in the migration note, both verified against the actual call sites; `border: 0` alongside `border: none` in both the note and the test; `docs/architecture.md` lines 364 and 562.

Two I could not reproduce, so I left them out rather than write them down:

- **Crashes from the null.** 23 valueless attributes (`style`, `class`, `id`, `href`, `src`, `colspan`, `rowspan`, `border`, `size`, `align`, `nowrap`, `cellpadding`, `start`, `bgcolor`, …) all parse and lay out on `main` without throwing. `DomParser.cs:1041`'s `TryGetAttribute("style")!` is the closest thing to the shape you describe, but `<div style>` doesn't throw either. If you have the fixture that does, I'll add it.
- **`height:Npx` rules 1.5pt shorter.** 13 height/border combinations measure identically on both sides (`height:1px`→2.25, `2px`→3, `4px`→4.5, `10px`→9, `20pt`→21.5, and the same with `border:none` / `border:2px solid`).

On #1229 I've [added your container-height case](https://github.com/jhaygood86/PeachPDF/issues/1229#issuecomment-5750429462) with my own measurements — the parent's used height is the same constant 2 for a 1/2/4/10px rule where a `<div>` child gives 1.5/3/6/15 — and replaced the §9.4.1 citation with §10.6.3 and §8.3.1+§10.5. I couldn't reproduce your 7.5/12/15, so the comment uses numbers I measured.

The 1px groove/ridge and 1–2px double two-tone point is real and is in the shared border code rather than anything `hr`-specific; I've left it alone here.

## Re-verification

Full suite 12632 passed (`Issue1157` still fails identically on clean `main`). Zero warnings on a full rebuild. Showcase regenerated and the `<hr>`-vs-`<div>` pairs re-diffed through both engines — mean channel difference 1.07–2.74, unchanged.

jhaygood86 (changes requested):
Re-reviewed head `9b8ed32c`. The blocking regression is fixed. One factual error in the notes remains, so I'm asking for a small correction rather than approving.

## Verified fixed
- **Default `<hr>` colours.** I generated PDFs from `main` and from this head and compared the vector fills. A default `<hr>` and `<hr size=3>` now paint `#9a9a9a` + `#eeeeee`, the same colours as `main` and Chrome 153 (the shapes are two bevel polygons instead of two strips, but the total height is identical). `<hr style="border:2px inset #808080">` still gives `#2c2c2c` / `#d4d4d4`. `<hr noshade>` is now flat `#808080`, matching Chrome. `border:0; border-top:1px solid #ccc` and `border:0; height:1px; background:#ccc` lose `main`'s phantom black lines.
- **`DefaultHr_PaintsTheTwoGreysABrowserPaints`** pins the default colours to Chrome's literal bytes, and `DefaultHrBaseColor_ShadesToExactlyThoseGreys` states the `#eee` relationship, so this cannot silently regress again.
- **Spec section** is now 15.3.11 in `CssDefaults.cs` and the test comment, `docs/architecture.md` is updated, and #1229's basis now cites §10.6.3 / §8.3.1 / §10.5.
- Own run: `dotnet build PeachPDF.slnx -t:Rebuild` 0 warnings / 0 errors; net8.0 12633 passed, 0 failed, 9 skipped.

## Needs correcting: the "did not crash anything" claim is false
`.claude/recent-fixes/2026-09-20-hr-paints-through-the-ordinary-border-path.md` now says the old null valueless-attribute value "did not crash anything: 23 valueless attributes (`style`, `class`, `id`, `href`, `src`, `colspan`, `rowspan`, `border`, `size`, `align`, `nowrap`, `cellpadding`, `start`, `bgcolor`, ...) all parse and lay out on `main`", and says this was "checked rather than assumed".

I checked it on `origin/main` with a throwaway layout test (24 snippets, each wrapped in a full document) and it does not hold. **`main` throws on 11 of 24; this branch throws on none:**

| snippet | `main` |
| --- | --- |
| `<p style>` | `ArgumentNullException` |
| `<hr size>` | `HtmlRenderException: Exception in box layout` |
| `<td bgcolor>`, `<td align>`, `<td valign>` | `NullReferenceException` |
| `<td width>` | `HtmlRenderException` |
| `<table cellspacing>` | `HtmlRenderException` |
| `<font color>`, `<font size>`, `` | `NullReferenceException` |
| `<input type>` | `HtmlRenderException` |

Those attributes only fail on an element that reads them (`<td bgcolor>`, `<font size>`, `<hr size>`); on an element that ignores them (for instance a `<div>`) they parse fine, which is presumably how the probe missed it. `main` at `v0.9.19` has the same `First().Value!`, so this is user-visible relative to the last release: a document that threw an exception now renders.

Please (1) replace the "did not crash anything" paragraph with what is true (valueless attributes on elements that consume them threw on `main`, and now render), and (2) add that to `.claude/migration-notes/2026-09-20-hr-honors-border-style.md` alongside the `<option value>` / `<mfenced open>` items already there. The change is an improvement, and the note should say so rather than deny it.

## Non-blocking (already disclosed, no action needed)
`hr { border-style: solid }` (or `dashed`) with no colour now paints `#eee` where Chrome paints gray, since the base colour is declared directly rather than derived from `currentColor`. The recent-fix and accepted-gap file record this under #1226 and warn against "simplifying" it, which is the right call for now. It is the one place this PR is visibly worse than `main` for an author-set style-only override, so it is worth keeping #1226 near the top of the queue.

Everything else is ready. I'll approve as soon as the two note edits land.

ChrisVanDijk:
Follow-up: `<hr color="red">` now paints red (5733f8c7). My earlier explanation for it was wrong.

The `color` attribute was never unwired. `DomParser.TranslateAttributes` maps it to the `color` property (`case HtmlConstants.Color: box.Color = value.ToLower()`), applied between the UA sheet and author rules — exactly where HTML puts a presentational hint. It reached `color` fine.

What blocked it was the UA sheet pinning `border-color` over the top: first the four per-side greys, then — in my own first pass at the flat-rule arm — `border-color: gray`. Which is the same mistake as the baked-in greys, one size down: **a colour named in the sheet that the engine could have derived.** A flat rule derives nothing from a bevel, so there was never a reason to resolve it by hand.

```css
hr              { color: gray; border: 1px inset #eee; }
hr[color],
hr[noshade]     { border-style: solid; border-color: currentcolor; }
```

Rendered against the `main` worktree, 8×, sampling each rule:

| | main | now | Chrome |
| --- | --- | --- | --- |
| `<hr>` | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` |
| `<hr size=3>` | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` | — |
| `<hr>` in a table cell | `#9a9a9a` / `#eeeeee` | `#9a9a9a` / `#eeeeee` | — |
| `<hr noshade>` | `#9a9a9a` / `#eeeeee` | `#808080` flat | `#808080` flat |
| `<hr color="red">` | `#9a9a9a` / `#eeeeee` | red flat | red flat |

The default rule is still byte-identical to `main` everywhere; the two attribute cases both now match Chrome. The spec's `hr { color: gray }` is what a bare `<hr noshade>` resolves to, and is also what stops an inherited `color` reaching a rule — `<div style="color:green"><hr></div>` is unchanged, pinned by its own test.

Tests: `HrWithAPresentationalColorAttribute_PaintsOneFlatColor` now asserts the painted colour per attribute (`noshade` → gray, `color=red` → red, `color=#0000ff` → blue) rather than just the style, plus `DefaultHr_KeepsItsOwnGrayAgainstAnInheritedColor`. Suite 12634 passed, zero warnings.

This retires point 3 of my [#1226 comment](https://github.com/jhaygood86/PeachPDF/issues/1226#issuecomment-5750431229); that issue is now only about the beveled case, which genuinely still needs a declared base until Blink's currentColor light-base rule exists in the cascade.

ChrisVanDijk:
You're right, and it's worse than 11 of 24 — corrected in 56316fe0.

Measured against a `main` **worktree** this time: **22 of 39 probed snippets throw on `main`, none here.**

| markup | main | this PR |
| --- | --- | --- |
| `<p style>` | `ArgumentNullException` | renders |
| `<div align>`, `<font color>`, `<font size>`, `<font face>` | `NullReferenceException` | renders |
| `<td bgcolor>`, `<td align>`, `<td valign>` | `NullReferenceException` | renders |
| `<p bgcolor>`, `<p background>`, `<body bgcolor>`, `<table bordercolor>` | `NullReferenceException` | renders |
| `<hr align>`, `<hr color>`, ``, `` | `NullReferenceException` | renders |
| `<hr size>`, `<td width>`, `<td height>`, `` | `HtmlRenderException` | renders |
| `<table cellspacing>`, `<input type>` | `HtmlRenderException` | renders |

Your diagnosis of why the probe missed it is exactly right: only an element that *consumes* the attribute dereferences the null, so `<div colspan>` and `<p nowrap>` parse fine and a probe that sprays attributes across arbitrary elements comes back clean.

## Why my probe was worthless, which is the part worth recording

The baseline came from `git stash -- src/PeachPDF`. **`git stash` cannot restore a file whose deletion is already committed** — and by then `HrFragmentPainter.cs`'s deletion was in `edb85ddb`. So the stash was a partial no-op and both columns were the branch.

Both columns came back byte-identical, which I read as confirmation. That is the tell, not the confirmation: a comparison you expect to differ somewhere, and which differs nowhere, means the baseline is wrong. I'd already hit the same trap on the UA-colour comparison earlier in this PR and switched to a worktree for *that* measurement without going back to re-run this one.

It's now written up as an invariant, since two false conclusions in one change is enough: `.claude/invariants/tooling-a-main-baseline-needs-a-worktree-not-a-stash.md`.

## Both requested changes

1. The "did not crash anything" paragraph in the recent-fix note is replaced with what's true, including the "only reachable through a consuming element" point and the methodology failure.
2. The migration note gains the table above under **"It also fixes 22 ways a document could fail to render"** — framed as the larger half of the change in practice, since a document containing `<p style>` or `<td width>` simply could not be converted before. The `<option value>` / `<mfenced open>` items now sit under it as the two *semantic* changes.

All 22 are pinned as `[InlineData]` cases on `ValuelessAttribute_OnAnElementThatConsumesIt_DoesNotThrow`.

## One more thing that fell out of redoing it properly

Your `height:Npx` claim still doesn't reproduce — 15 height/border combinations are identical against the real baseline. But re-running it *did* surface a flow difference I'd previously missed for the same reason: content after `<hr style="border: 0">` now starts at the rule's top rather than 2 units below it. That matches what `border: none` already did on `main`, so the two spellings agree now, and both are governed by #1229. It's in the migration note.

Suite 12656 passed, zero warnings.

jhaygood86 (changes requested):
Re-reviewed head `3c7a0443`. The code is good and I found no defects in it; the one thing I'm asking for is a missing migration-note paragraph, then this is ready.

## Verified (Chrome 153.0.8010.48, PDFium + MuPDF, real `v0.9.19` / `main` worktrees)
- **`currentcolor` change.** `<hr color=red>`, `color="#0000ff"`, `color=red noshade` and `color=red width=50%` now paint that colour flat like Chrome (main painted two-tone gray); `<hr noshade>` and `<div style="color:green"><hr noshade></div>` paint flat `#808080` like Chrome. A default `<hr>` (alone, or in a green/red-text block) is still `#9a9a9a` / `#eeeeee`, byte-identical to `main` and equal to Chrome. Spec check: HTML Standard 15.3.11 has `hr[color], hr[noshade] { border-style: solid }` and the `color` attribute as a presentational hint for `color`, so the design is spec-faithful, and 15.3.6 is correctly cited as "Sections and headings".
- **The 22-crash table is accurate.** All 22 markups throw on `v0.9.19` and on `main` with the claimed exception types, and none throw on this branch. A wider probe of 248 valueless attributes on their consuming elements found 83 that throw on `v0.9.19`/`main` and **0 on this branch** (it also covers `<a href>`, `<svg style|width|height>`, `<canvas>`, `<object>`, `<embed>`, many MathML attributes, `<table height|bgcolor|background|align>`, `<html style>`, `<body style|background>` and more).
- **#1232 is real and pre-existing.** Root cause confirmed at `CssBoxHr.cs:118-126` by instrumenting the fallback (`Size.Height = -1.5` with 0.75pt borders, so the sum is exactly 0 and `height = 2` always wins). The 0.5pt white gap on a default rule is byte-identical on `main`, and the gap-file's follow-on numbers (+2 to +9 after a `border:6px` rule, matching Chrome; declared-height rules unaffected) hold.
- **Mutation tests bite:** reverting `border-color: currentcolor` to `gray` fails 3 of the new colour cases; dropping `color: gray` fails `DefaultHr_KeepsItsOwnGrayAgainstAnInheritedColor`. `Hr|Valueless` filter: 857 passed, 0 failed.
- **Own run:** `dotnet build PeachPDF.slnx -t:Rebuild` 0 warnings / 0 errors; net8.0 12657 passed, 0 failed, 9 skipped; CI green on all three platforms. No convention violations (no `#NNN` in `docs/**`, one-file-per-entry, no AI attribution).

## Required: document the change to author `border-top: 1px solid` rules
Giving `hr` its own `color: gray` (correct, and it matches Chrome's UA sheet) changes a very common author idiom that the migration note does not mention. A rule whose border colour is `currentColor` used to take the *inherited text colour*; it now takes the hr's own gray:

```html
<style>hr { border: 0; border-top: 1px solid }</style>
<div style="color: green"><hr></div>
```
- `v0.9.19` / `main`: black (green inside the div). This PR: `#808080`. Chrome: `#808080`.
- Same for `border: 1px solid`, `border: 4px double`, `border-top: 2px dashed`, `border-bottom: 1px solid`, and an explicit `border-color: currentColor` on a bare rule.
- An author `color` on the rule (`hr { color: red; ... }`) or `color: inherit` still wins, as before.

That is an improvement and it matches Chrome, but it is a visible change against the last release, so please add a short paragraph to `.claude/migration-notes/2026-09-20-hr-honors-border-style.md` (the existing text only says an *inherited* colour no longer reaches a default rule). A test pinning the `border-top: 1px solid` (no colour) idiom would also protect it.

## Non-blocking
- **Legacy colour parsing for `<hr color>`** (`DomParser.cs:2044`, shared with `<font color>` and pre-dating this PR): `color=""`, valueless `color` and `color=transparent` now paint black flat or nothing where Chrome paints gray (no colour hint), and `color=ff0000` (no `#`) paints black where Chrome paints red. Main ignored the attribute on `<hr>`, so these rare inputs differ from `main`. Running the HTML legacy-colour algorithm there and skipping the hint on failure would fix it; fine as a follow-up.
- Reword "22 ways" to "at least 22" in the note and recent-fix, since the real set is larger (`<a href>` alone throwing is notable).
- The migration note's "Nothing that rendered before renders differently" sits a few lines above the `<option value>` / `<mfenced open>` changes, which are exactly that; "Nothing that threw before ..." would read cleaner.
- The PR body is stale (still says "no test moved" and omits `color: gray`/`currentcolor`, the 22 crash fixes and #1232).
- Documented residue, unchanged and not asked for here: a rule that sets only `border-style` (no colour), e.g. `hr { border-style: solid }`, paints `#eee` where Chrome paints gray, so it is close to invisible on white. That cannot be fixed without Blink's currentColor-bevel rule (#1226).
- Pre-existing and identical on `main`: Chrome fills a coloured/`noshade` rule's interior (visible with `height:20px`), 2pt total height vs Chrome's 1.5pt (#1232), `size` attribute heights, 272 vs 274 percentage width (#1230).

I'll approve as soon as the migration-note paragraph lands.

ChrisVanDijk:
Right, and it's broader than the default rule — fixed in cac1319b.

Measured against a `main` worktree:

| | v0.9.19 / main | this PR | Chrome |
| --- | --- | --- | --- |
| `hr { border: 0; border-top: 1px solid }` inside `color: green` | `#008000` | `#808080` | `#808080` |
| the same with no surrounding colour | `#000000` | `#808080` | `#808080` |
| `hr { border: 1px solid }` | `#008000` | `#808080` | `#808080` |
| `hr { border: 4px double }` | `#008000` | `#808080` | `#808080` |
| `hr { border: 0; border-top: 2px dashed }` | `#008000` | `#808080` | `#808080` |
| `hr { border: 0; border-bottom: 1px solid }` | `#008000` | `#808080` | `#808080` |
| `hr { border-color: currentcolor }` | `#008000` | `#2c2c2c` / `#d4d4d4` | — |
| `hr { color: red; ... }` | `#ff0000` | `#ff0000` | unchanged |
| `hr { color: inherit; ... }` | `#008000` | `#008000` | unchanged |

Two notes on that table beyond your list. The **bare** case (no surrounding colour) changes too — black to gray — which is the one an author is most likely to have been relying on without a wrapper anywhere. And `border-color: currentcolor` changes for *two* reasons at once: the colour base, and the UA's `inset` now actually applying, so it goes from one flat green band to two beveled gray faces.

## Both requested changes

**Migration note**: new section, *"An author border with no colour of its own is now gray, not the inherited text colour"*, with the table, the list of affected shorthands, and the two escape hatches (`color: red` wins; `color: inherit` restores the old behaviour). The existing sentence that only mentioned the default rule now hands off to it rather than implying that was the whole story.

**Tests**: `AuthorBorderWithNoColor_ResolvesToTheRulesOwnGray` over five shorthand idioms, and `AuthorColorOnARule_StillWinsOverTheUaGray` over both author-colour spellings.

One thing the tests turned up while writing them: `hr { border-top: 3px solid; border-bottom: 0 }` — which I'd included as a sixth idiom — correctly paints **three** colours, because it leaves the UA's left and right borders at `1px inset #eee`. That's right, not a bug, but it made a bad test case, so it's out. The dashed idiom is in and needed the assertion to read stroke colours as well as fills, since a patterned border strokes rather than fills.

Suite 12663 passed, zero warnings. #1233 is rebased on top of this and re-verified (12688 passed).

jhaygood86 (approved):
Head `cac1319b` has the migration-note paragraph and tests I asked for. Approving.

- **Migration note.** The new section on an author border with no colour of its own is accurate: `hr { border: 0; border-top: 1px solid }` was black (or the surrounding text colour) on v0.9.19 and is `#808080` now, matching Chrome 153, and `hr { color: inherit }` restores the old behaviour. The rows in its table agree with what I measured against `main`/`v0.9.19` and Chrome.
- **Tests.** `AuthorBorderWithNoColor_ResolvesToTheRulesOwnGray` pins the five common spellings (`border-top:1px solid`, `border:1px solid`, `border:4px double`, `border-bottom:1px solid`, `border-top:2px dashed`) inside a green block, and `AuthorColorOnARule_StillWinsOverTheUaGray` pins `color: red` and `color: inherit`.
- Everything else from the earlier rounds stands: default `<hr>` `#9a9a9a`/`#eeeeee` byte-identical to `main` and Chrome, `<hr color=red>` and `<hr noshade>` matching Chrome, all 22 crash-table rows accurate (248 valueless-attribute probes, 83 throwing on `main`, 0 here), #1232 root cause confirmed.
- Own run on this head: `dotnet build PeachPDF.slnx -t:Rebuild` 0 warnings / 0 errors; net8.0 12664 passed, 0 failed, 9 skipped; CI green on all three platforms.

Still open and non-blocking (unchanged from my last review): legacy colour parsing for `<hr color="">` / `color=ff0000` / `color=transparent`, wording "22 ways" -> "at least 22", the stale PR body, and the documented `hr { border-style: solid }` residue tracked as #1226.

PR #1233, opened 2026-09-20
Resolve an <hr>'s own used height instead of a constant 2 units

### LINKED ISSUE #1229: <hr> advances the block flow by a constant 2 units, so the next block overlaps any rule taller than that

### LINKED ISSUE #1232: <hr> paints a white gap between its top and bottom borders: the auto-height fallback double-counts the borders
`<hr>` paints a white gap between its top and bottom borders, because the auto-height fallback double-counts the borders.

A default `<hr>` renders as three bands, not two: the top border, **a strip of white content**, then the bottom border. Chrome draws the two borders touching. It is visible at any zoom past ~400% and shows up in print.

Scanned down the middle of a default rule at 16 raster px per point:

| band | painted | size |
| --- | --- | --- |
| top border | `#9a9a9a` | 0.75pt |
| **content** | **white** | **0.5pt** |
| bottom border | `#eeeeee` | 0.75pt |
| | | **2pt total** |

Chrome's default rule is 1px + 1px with no content between them. `<hr>`'s used content height should be 0 here: [CSS 2.1 §10.6.3](https://www.w3.org/TR/CSS21/visudet.html#normal-block) gives a block-level non-replaced element in normal flow with `height: auto` and no in-flow children a used content height of 0, and nothing in the [HTML Standard's rendering rules for `hr`](https://html.spec.whatwg.org/multipage/rendering.html#the-hr-element) sets a height.

## Cause

`CssBoxHr.PerformLayoutImp` (`src/PeachPDF/Html/Core/Dom/CssBoxHr.cs:117-125`):

```csharp
double height = ActualHeight;
if (height < 1)
{
    height = Size.Height + ActualBorderTopWidth + ActualBorderBottomWidth;
}
if (height < 1)
{
    height = 2;
}
```

By the time this runs, `PlaceAsBlockChild` has written `ActualBottom = Location.Y`, and that setter stores `Size.Height = value - ActualBoxSizeIncludedHeight - Location.Y`. So for an auto-height rule `Size.Height` is exactly **`-(borderTop + borderBottom)`** — instrumented: `-1.5` with the default 0.75pt borders, `-3` with 2px ones, `ActualHeight` `0` in both cases.

Adding the borders back to that is therefore **exactly 0**, always. The second check can never fail to fire, so every auto-height rule takes the hard-coded `height = 2`, and a 2pt border box with less than 2pt of border has the remainder left over as content.

## Measured

| declaration | border box | borders | content |
| --- | --- | --- | --- |
| *(default)* | 2 | 1.5 | **0.5** |
| `border: 1px solid` | 2 | 1.5 | **0.5** |
| `border: 0.5pt solid` | 2 | 1 | **1** |
| `border: 1pt solid` | 2 | 2 | 0 |
| `border: 1.5pt solid` | 3 | 3 | 0 |
| `border: 2px solid` | 3 | 3 | 0 |
| `border: 3px solid` | 4.5 | 4.5 | 0 |
| `border: 6px groove` | 9 | 9 | 0 |

Points, at 1px = 0.75pt. The gap is `max(0, 2 - (borderTop + borderBottom))`, so it appears exactly when the two borders total under 2pt — which includes the UA default. A rule with a thin author border is worse than the default: `border: 0.5pt` gives a 1pt gap, wider than either border.

## Not a regression

Identical on `main` and on #1231 — the same three bands at the same sizes. #1231 changed what colour the bands are, not how many there are.

## Same root cause as #1229

`height = 2` is also why a rule advances the block flow by a constant 2 units and why its parent's used height is a constant 2. #1229 recorded the symptom and guessed at "an `ActualBottom` written before the rule's real height was resolved"; this is the actual mechanism, and the two should be fixed together.

## Fix

Dropping the poisoned term is a one-liner:

```csharp
height = ActualBorderTopWidth + ActualBorderBottomWidth;
```

Measured with that applied (full suite passes):

| | now | with the fix | Chrome |
| --- | --- | --- | --- |
| default rule's border box | 2pt, 0.5pt gap | **1.5pt, no gap** | 2px, no gap |
| next block after `border: 6px` | +2 | **+9** | +9 |
| next block after `height: 10px` | +2 | +1.5 | +10.5 |

So it closes this issue and the auto-height half of #1229. It does **not** fix a rule with a declared `height`, whose following sibling still advances by the border total — that sibling is placed on an earlier pass and never re-placed, which is the remaining half of #1229.

It is a user-visible change to every default rule's height (2pt → 1.5pt), so it wants its own change and a migration note rather than riding along with an unrelated one.

ChrisVanDijk:
Fixes #1229
Fixes #1232

> **Stacked on #1231 — merge that first.** This branches off `fix/1225-hr-border-style`, so until #1231 merges the diff below also contains its commits. This change's own commit is the last one, `66e96827`; reviewing that commit alone gives the clean diff. Without #1231 underneath, the `<hr>` this fixes the *geometry* of would still be discarding `border-style`.

A rule's used height, the next in-flow sibling's position, and the height of a container holding only the rule were all the same constant — **2 units** — whatever the rule's borders or declared `height`. One expression caused all three:

```csharp
double height = ActualHeight;
if (height < 1)
    height = Size.Height + ActualBorderTopWidth + ActualBorderBottomWidth;
if (height < 1)
    height = 2;
```

`PlaceAsBlockChild`, a few lines above, writes `ActualBottom = Location.Y`, and that setter stores `Size.Height = value - ActualBoxSizeIncludedHeight - Location.Y`. So by the time this runs `Size.Height` is exactly `-(borderTop + borderBottom)` — instrumented: `-1.5` with the default 0.75pt borders, `-3` with 2px ones, `ActualHeight` `0` in both. Adding the borders back cancels to zero every time. **The middle branch was unreachable**, and every rule took the hard-coded `2`.

### #1232 — the painted gap

A 2pt box with less than 2pt of border leaves the remainder as real *content*, which paints white between the two border edges. The default rule drew three bands at high zoom where a browser draws two. The gap is `max(0, 2pt − borders)`, so a *thinner* author border was worse: `border: 0.5pt` got a 1pt gap, wider than either border it separated.

### #1229 — the constant flow advance and container height

The same `2` is what the frame commits the next sibling's offset against, and what a content-sized container takes. `<div><hr style="border:10px solid"></div><p>after</p>`: the rule is 15pt tall, `<p>` started 2pt below its *top*, and the `<div>` was 2pt tall.

## The declared-height half needed the call order

Dropping the poisoned term alone fixes auto-height rules but not `height: 10px`. Tracing showed why:

```
COMMIT child=hr#el top=26
HR-LAYOUT ActualHeight=0 sizeH=-1.5 declaredHeight='10px'
  HR-DONE height=1.5 bottom=27.5          <- the declared 10px never reached it
COMMIT child=div#nx top=27.5              <- the sibling is committed HERE
  APPLYHEIGHT before bottom=27.5 boxHeight=9
  APPLYHEIGHT before bottom=35            <- corrected, one step too late
```

A declared height normally arrives via `CssLayoutEngine.ApplyHeight` in `PerformLayoutEpilogue`, which for this box runs *after* the frame has placed the next sibling — and `ActualHeight` could never see it anyway, since it reads the poisoned `Size.Height`. So the rule now resolves through `GetBoxHeight`, `ApplyHeight`'s own resolver and the only one that doesn't read `Size.Height` back:

```csharp
double height = CssLayoutEngine.GetBoxHeight(this) ?? 0;
if (height <= 0)
    height = ActualBorderTopWidth + ActualBorderBottomWidth;
if (height <= 0)
    height = 2;
```

`<= 0` rather than `< 1` deliberately: while `ActualHeight` was always 0 the old threshold was dead code, but once real values flow through it `< 1` would reject a genuinely small declared height (`height: 0.5pt`) or a thin border pair.

## What still reaches the constant

A rule with neither a height nor a border. Removing it is tempting — Chrome gives that rule height 0 — but a zero-size box **emits no fragment at all**, so it would be absent from the fragment tree rather than merely invisible, and `BorderlessHr_PaintsNothing` fails on the missing fragment. Tried, measured, reverted, and pinned by a test so it isn't "cleaned up" later. That leaves one pre-existing inconsistency, unchanged and invisible: a borderless rule's box is 2 while its flow advance is 0, because it collapses through.

## Behaviour change

Only rules whose borders total under 2pt change height, and they lose the leftover content — which is what closes the gap:

| declaration | before | now |
| --- | --- | --- |
| *(default)* | 2 | **1.5** |
| `border: 0.5pt solid` | 2 | **1** |
| `border: 2px solid` | 3 | 3 |
| `border: 6px groove` | 9 | 9 |
| `height: 10px` | 9 | 9 |
| `border: none` | 2 | 2 |

Everything from 2pt of border upward was already the right height and is untouched; what changes for those is the *flow* around them. Migration note added.

## Verification

- Suite 12681 passed (`Issue1157` still fails identically on clean `main`). Zero warnings. All four changed lines covered, both branches 100%.
- `HrAutoHeightLayoutTests` asserts the used height, the absence of content between the borders, the next sibling's position and the container's height across the declaration matrix — plus a paint test asserting the two borders share **one** inner edge, on the distinct Y coordinates of everything painted rather than a band count (a bevel's two draw calls are mitred L-pairs whose bounding boxes each span the whole rule either way; what the gap actually added was a fourth Y).
- The `border_style` showcase's `<hr>`-vs-`<div>` section got *stronger*: per cell, the rule and its equivalent zero-height `<div>` are now **pixel-identical in MuPDF** (0.000 on seven of eight; `dotted` is 0.84 from sub-pixel dash phase) and 0.08–0.72 in PDFium, against 1.07–2.74 before — because the rule no longer overlaps the div it's compared with.

One measurement note for anyone re-checking that section: diff it **per cell**, not per row. Each pair's offset is now the rule's own height plus its margin, so it differs per cell (90, 90, 108, 108 raster px at 6× on the first row); a single whole-row offset finds nothing and reports a large difference, which is what mine did at first.


jhaygood86 (changes requested):
Reviewed commit `66e96827` only (the first five commits are #1231's, which I'm reviewing there, and this can't merge before #1231 does; it will also need a rebase once #1231 lands). The core fix works and is a clear improvement, but it introduces one geometry regression and one gap, and neither is caught by the tests.

## What checks out (Chrome 153.0.8010.48 headless print-to-PDF as ground truth; v0.9.19, the #1231 head and this head compared; ~138 `<hr>` declarations in two wrappers plus context and pagination batches)
- **#1232 is fixed.** The white slit between the two borders is gone in PDFium and MuPDF at 16x for the default rule, `0.5pt`, `1px`, `2px`, `size=1/3/5`, `3px inset`, `6px groove`, `noshade`, `color=red`, `background`, `opacity`, outline and `border-radius`. (One 0.06pt antialiasing seam remains at 16x where two faces meet; invisible at 1x.) A default rule is now 1.5pt, matching Chrome.
- **#1229 is fixed for the common idioms.** Flow advance is within 0.1pt of Chrome for the default rule, `border` 0.5pt/1px/2px/3px inset/6px groove/10px, `border:0; border-top:1px`, `border:0; height:1px; background`, `height` in `px`/`%`/`calc()`/`em`/`rem`/`vh`/`vw`, `min-height`, `box-sizing:border-box` with a border, and in td / flex / grid / li / blockquote / inline-block / float contexts. Multicol no longer overlaps the follower.
- **Pagination improves.** An exact-fit rule, a rule taller than a page (`height:2000px` gives 3 pages like Chrome instead of 1), a rule in a `break-inside:avoid` wrapper at the page bottom, and a 120-section document (tail at 546.0 vs Chrome 545.2, baseline 606.0) all move toward Chrome. No lost or empty fragments for `height:0.5pt`, `border:0`, `height:0`.
- **Hazards:** `-5px`, `NaNpx`, `calc(1px/0)`, `1e400px` neither crash nor hang.
- **Mutations:** the old poisoned expression fails 15 tests (including `DefaultHr_PaintsItsTwoBordersMeetingOnOneEdge`); dropping the `GetBoxHeight` term fails 3; changing the final `2` to `0` fails `BorderlessHr_KeepsItsNominalHeight` and both `BorderlessHr_PaintsNothing` cases, as the commit claims. The tests use Chrome-derived literals, not self-consistency.
- Migration-note before/after numbers hold against v0.9.19 (box heights 2/2/3/9/9, follower advance 2 everywhere); no AI attribution; no `#NNN` in `docs/**`.

## Required changes
**1. Padding is counted twice (regression vs the #1231 head and v0.9.19).** `GetBoxHeight` returns a border-box height (declared height plus `ActualBoxSizeIncludedHeight`, floored at borders plus padding), and `PerformLayoutImp` then adds padding again: `ActualBottom = Location.Y + ActualPaddingTop + ActualPaddingBottom + height` (`CssBoxHr.cs:153`).
- `<div style="height:10px;background:#f00"></div><hr style="padding:5px"><div style="height:10px;background:#00f"></div>`: the rule paints **16.5pt**; Chrome 9.0, #1231 head 9.5. The follower and container are 7.6pt too low (28.6 / 43.6 vs Chrome 21.0 / 36.0).
- `<hr style="padding:5px;height:10px">`: follower offset 36.1 vs Chrome 28.5 (#1231 head 21.6); with `border:2px` added, 37.6 vs 30.0; `box-sizing:border-box;padding:5px;height:20px`: 34.6 vs 27.0.
- `<hr style="border:0; height:1px; padding:2px 0; background:#ccc">`: 3.0pt gap below it vs Chrome 0.
The review agent confirmed the cause with a scratch edit (floor at padding plus borders and use `ActualBottom = Location.Y + height`): it matched Chrome to +0.1pt on all four padding cases with no other change across 70 other cases. Please make that change (or subtract padding once) and add tests for padding-only, padding plus declared height, and `border-box`; none of the current tests uses padding.

**2. `max-height` is never applied when placing the follower.** `GetBoxHeight` omits it (CSS 2.1 §10.7 applies `max-height` first, then `min-height`); `ApplyHeight` applies it, but only in the epilogue, after the next sibling is already committed. `<hr style="height:20px;max-height:5px">` paints 5.25pt but places the follower 11.35pt too low (28.6 vs Chrome 17.25); `height:200px;max-height:50%` in a 100px parent leaves it 112.5pt too low (163.6 vs 51.0). Either clamp with `max-height` in the rule's own pass, or record it as a tracked gap (with an issue and an accepted-gaps entry) and add a test that pins the current behaviour.

**3. The `<= 0` change is untested.** Reverting `<= 0` to `< 1` passes all 876 hr tests in all three variants, although the commit message explains exactly why `<= 0` matters. Please add `border:0; border-top:1px solid` (0.75pt, the final fallback) and `border:0; height:0.5pt` (the declared branch) to `Hr_UsedHeight_IsItsContentPlusItsOwnBorders`.

## Non-blocking
- Stale comment at `CssBoxHr.cs:66-72` ("makes the fallback below fire on a re-layout instead of reading the previous pass's height") no longer describes the code; two consecutive blank lines after the `height = 2` block.
- `.claude/accepted-gaps/hr-flow-advance-and-percentage-width.md` now only covers #1230 (percentage width) and #1226 (bevel base); the name no longer matches, and gap files are named for the gap.
- The docs sentence "that height is what the next block is placed against and what a container holding only the rule takes" isn't true for padding or `max-height` combinations until items 1-2 are addressed; the migration note also says nothing about padding.
- The `size` attribute maps to `Height = N` (`DomParser.TranslateSize`); the HTML Standard says height N-2 (or `border-bottom-width: 0` for `size=1`). Chrome paints `<hr size=3>` at 2.25pt, PeachPDF at 3.75pt, and this PR makes the flow follow that oversized paint (+1.6pt vs Chrome; the old constant only looked closer because it overlapped). Pre-existing; worth its own issue since `size=1` is common in legacy HTML.
- `height:calc(100% - 5px)` in an indefinite parent resolves to ~1.5pt in the rule's own pass but 26.35pt in the epilogue, so paint and flow disagree (the old code had the same class of error). Chrome treats it as auto.
- Any zero-margin `hr` after a sibling still gets the 12.1pt gap from `CssBox.cs:8204` (pre-existing since v0.9.19); `hr { margin: 0 }` is a common idiom, so that deserves its own issue.
- Pre-existing and unchanged: `page-break-before:always` / `break-before:page` on an `<hr>` are ignored; `<hr style="height:0;border:0">` still advances 2pt (Chrome 0).
- I didn't re-run the showcase, so the "pixel-identical in MuPDF" claim for the `<hr>`-vs-`<div>` section is unverified by me.
- §10.5 is a loose citation for follower placement; §9.4.1 or §8.3.1 fits better.

Happy to re-review once #1231 has merged and this has been rebased with the padding fix.

ChrisVanDijk:
Rebased onto merged `main` (#1231 is in), so this is now a clean two-commit branch. All three required changes are in `e7636aa5`.

## 1. Padding counted twice — fixed

`ActualBottom` is now `Location.Y + height`, since `GetBoxHeight` already includes the padding. Measured against a `main` worktree:

| | main | now | Chrome |
| --- | --- | --- | --- |
| `padding: 5px` — height | 9.5 | **9** | 9 |
| `padding: 5px; height: 10px` — next block | +9.5 | **+16.5** | +16.5 |
| `box-sizing: border-box; padding: 5px; height: 20px` — next block | +9 | **+15** | +15 |
| `border: 0; height: 1px; padding: 2px 0` — height | 5.25 | **3.75** | 3.75 |

**I did not take the floor**, and the reason is worth a look. Flooring the resolved height at `borders + padding` gives the same answer on every case bar one: `box-sizing: border-box; height: 5px; padding: 5px`, whose border box is smaller than its own edges. `GetBoxHeight` returns it as declared, and `ApplyHeight` re-runs `GetBoxHeight` in the epilogue — so a floor applied *only here* puts the flow at 9 and the paint at 3.75. Your scratch edit's 70 cases wouldn't have caught it. So it stays a fallback (`height <= 0`), which is your "or subtract padding once" alternative, and box == paint == flow on every case I measured except the known borderless collapse-through. Whether `GetBoxHeight` should clamp that case at all is a separate question; what this method must not do is answer it differently from the paint. That reasoning is in the code comment.

## 2. max-height — clamped in the rule's own pass

Rather than tracking it as a gap, `ApplyHeight`'s trailing clamp is extracted as `CssLayoutEngine.ClampToMaxHeight` and called from both places — the repo's one-resolver-two-callers rule. It only ever shrinks a bottom past the maximum, so the epilogue's own call is a no-op once the rule has clamped itself.

| | main | now | Chrome |
| --- | --- | --- | --- |
| `height: 20px; max-height: 5px` — next block | +2 | **+5.25** | +5.25 |
| `height: 20px; max-height: 5px; min-height: 12px` | — | **10.5** | §10.7: min wins |

A percentage `max-height` against an indefinite parent still doesn't apply, which is also what a browser does.

## 3. `<= 0` — now tested, and the first attempt didn't test it

Both cases added. Worth flagging how nearly this stayed untested: putting them on `Hr_UsedHeight_...` alone catches only **one** of the two, because the epilogue re-resolves the rule's own bottom and corrects the wrong value before anything reads it back. The sibling's offset is committed from the rule's own pass and never revisited, so it's the only observable that sees the difference. Both are now pinned there as well — reverting `<= 0` to `< 1` fails three assertions instead of one.

## Non-blocking — all addressed

- Stale `<remarks>` rewritten to describe what the code does now; no double blank line (that was already gone).
- Gap file renamed to `hr-percentage-width-and-currentcolor-bevel-base.md`.
- Docs sentence now holds — it covers padding and `max-height`, both of which this commit makes true.
- Migration note gained a padding section and a `max-height` section with the numbers above.
- §10.5 → §9.4.1 with §8.3.1, in the test comment and the note.
- **Showcase re-verified**, since you flagged it as unverified. Per cell, both renderers: PDFium worst **0.715**, MuPDF worst **0.837**, and **0.000 on seven of eight cells in MuPDF** — the outlier is `dotted`, sub-pixel dash phase. Three traps worth knowing if you re-run it, each of which gave me a confident wrong number first: measure per cell (offsets now differ per cell), find the swatches by shape not by row position (cell heights differ), and start the scan below the section heading — the `<h2>` text *and* its `border-bottom` both span the full page width, so "spans this cell" matches them too. Cells 3 and 4 looked correct throughout because the heading text doesn't reach that far right, which reads exactly like a real rendering difference.

## Filed rather than fixed here

- **#1234** — `size=N` maps to `height: N`, spec says `N−2` (and `size=1` → `border-bottom-width: 0`). Your numbers reproduce: `<hr size=3>` is 3.75pt against Chrome's 2.25pt.
- **#1235** — `hr { margin: 0 }` defeated by the hard-coded 1.1em at `CssBox.cs:8204`. The suggested fix is to add the spec's `margin-block: 0.5em; margin-inline: auto` to `CssDefaults` (absent today) and delete the special case, which also gets `<hr width="50%">` centring.
- **#1236** — the percentage/`calc()` height paint-vs-flow split. Recorded as a tracked gap with a reader-facing docs note, because the fix is in `GetBoxHeight`'s contract (return `null` for a percentage against an indefinite containing block) and reaches every caller.

Suite **12706 passed, 0 failed** — the `Issue1157` failure that was outstanding is fixed on `main` by #1228. Zero warnings. All changed lines covered, every branch 100%.

jhaygood86 (approved):
Head `e7636aa5` addresses all three points from my review. Approving.

- **Padding counted once.** I rendered each case into a real PDF on this head and on `main` and measured red-block-bottom to blue-block-top. This head is within +0.1pt of Chrome 153's numbers from my earlier review in every case (the +0.1 is the pre-existing 12.1pt zero-margin gap at `CssBox.cs:8204`): `padding:5px` 21.1 (Chrome 21.0, `main` 21.6), `padding:5px;height:10px` 28.6 (28.5; `main` 21.6), with `border:2px` 30.1 (30.0), `box-sizing:border-box;padding:5px;height:20px` 27.1 (27.0), default 13.6 (13.5), `border:0; border-top:1px` 12.85, `border:10px` 27.1.
- **`max-height` in the rule's own pass.** `height:20px;max-height:5px` gives 17.35 (Chrome 17.25; `main` 14.1) and `height:200px;max-height:50%` in a 100px parent gives 51.1 (Chrome 51.0; `main` 14.1). `ClampToMaxHeight` is a pure extraction of the existing clamp at the tail of `ApplyHeight` (the diff moves the body into a helper and calls it from the same place), so the epilogue path for every other block is unchanged.
- **`<= 0` is tested.** `border: 0; border-top: 1px solid` (0.75) and `border: 0; height: 0.5pt` (0.5) are now in `Hr_UsedHeight_IsItsContentPlusItsOwnBorders` and `NextInFlowSibling_StartsAtTheRulesBottom`.
- **Mutation tests, all caught** (restored afterwards): re-adding padding to `ActualBottom` fails 5 tests (padding-only, padding plus height, `border-box`, container); dropping the `ClampToMaxHeight` call fails 2; reverting `<= 0` to `< 1` fails 3.
- **Housekeeping:** the stale comment is replaced, the gap file is renamed to `hr-percentage-width-and-currentcolor-bevel-base.md`, `docs/**` carries no `#NNN`, and the migration note's "before" numbers agree with what I measured on `main` (9.5 / +9.5 / +9 for the padding rows).
- **Own run:** `dotnet build PeachPDF.slnx -t:Rebuild` 0 warnings / 0 errors; net8.0 12706 passed, 0 failed, 9 skipped; CI green on all three platforms.

Still open and non-blocking, unchanged from my earlier review: the `size` attribute mapping (Chrome paints `<hr size=3>` at 2.25pt, PeachPDF 3.75pt) and `hr { margin: 0 }` still getting the 12.1pt gap from `CssBox.cs:8204` are both worth their own issues; `page-break-before/after` on an `<hr>` is still ignored; `height:calc(100% - 5px)` in an indefinite parent still disagrees between the rule's own pass and the epilogue.

